PR #65, opened 2018-12-11
fix: throw "No AuthInfo found" if alias doesn't exist

jayree:
Suggested solution for #64


jayree:
@rbergman I've updated the unit test "missing config" and fixed "cache hit and miss". Please review the changes.

jayree:
That would be the best solution. Are you taking it from here?

rbergman:
Tim has a new PR open -- thanks for getting this started!

PR #69, opened 2018-12-13
fix: ensure AuthInfoConfig can throw when config file not found

tnoonan-salesforce:
Fixes #64

### DIFF
--- a/src/authInfo.ts
+++ b/src/authInfo.ts
@@ -444,7 +444,9 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
 
     this.logger.debug(dataToSave);
 
-    const config = await AuthInfoConfig.create(AuthInfoConfig.getOptions(username));
+    const options = AuthInfoConfig.getOptions(username);
+    options.throwOnNotFound = false;
+    const config = await AuthInfoConfig.create(options);
     config.setContentsFromObject(dataToSave);
     await config.write();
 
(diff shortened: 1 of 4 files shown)


src/config/authInfoConfig.ts
```diff
@@ -15,16 +15,38 @@ import { ConfigFile } from './configFile';
...
+
+export namespace AuthInfoConfig {
+  /**
+   * Options for the AuthInfoConfig.
+   */
+  export interface Options extends ConfigFile.Options {
+    /**
+     * Indicates if init should throw if the corresponding config file is not found.
+     */
+    throwOnNotFound?: boolean;
```
rbergman: Since `ConfigFile.read` accepts `throwOnNotFound`, couldn't we just add `throwOnNotFound?` to the base `ConfigFile.Options`?  Then we could drop it as an `AuthInfo.init` param and any `ConfigFile` subtype could take advantage of it.

src/authInfo.ts
```diff
@@ -444,7 +444,9 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
 
     this.logger.debug(dataToSave);
 
-    const config = await AuthInfoConfig.create(AuthInfoConfig.getOptions(username));
+    const options = AuthInfoConfig.getOptions(username);
+    options.throwOnNotFound = false;
+    const config = await AuthInfoConfig.create(options);
```
rbergman: I believe you could do this in one line still like so:
```
const config = await AuthInfoConfig.create({
  ...AuthInfoConfig.getOptions(username),
  throwOnNotFound: true
});
```

PR #101, opened 2019-02-08
fix: use default options off the constructor

amphro:
also update all the incorrect docs for AsyncCreatable

Fixes: @W-5676674@, @W-5676708@

### DIFF
--- a/src/config/configFile.ts
+++ b/src/config/configFile.ts
@@ -52,7 +52,7 @@ export class ConfigFile<T extends ConfigFile.Options> extends BaseConfigStore<T>
     return {
       isGlobal,
       isState: true,
-      filename: filename || ConfigFile.getFileName()
+      filename: filename || this.getFileName()
     };
   }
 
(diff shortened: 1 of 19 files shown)


src/config/configFile.ts
```diff
@@ -182,14 +182,16 @@ export class ConfigFile<T extends ConfigFile.Options> extends BaseConfigStore<T>
   /**
    * Used to initialize asynchronous components.
    *
-   * **Throws** *`Error`{ code: 'ENOENT' }* If the username.json file is not found when
+   * **Throws** *`Error`{ code: 'ENOENT' }* If the {@link ConfigFile.getFilename} file is not found when
    * options.throwOnNotFound is true.
    */
   protected async init(): Promise<void> {
```
amphro: This is how I was proposing we fix the issue @shetzel - lets chat about it in the morning.

shetzel: Yes, this is essentially how I fixed it locally.  It all works this way, and is backward compatible, but being static makes it a less intuitive ux imo.  For example, the link in the comment isn't exactly accurate since it's not `ConfigFile.getFilename` but `this.constructor.getFilename`.  Seems more intuitive to just be `this.getFilename`, as well as `this.getDefaultOptions`.

src/config/configFile.ts
```diff
@@ -182,14 +182,16 @@ export class ConfigFile<T extends ConfigFile.Options> extends BaseConfigStore<T>
...
+    const statics = this.constructor as typeof ConfigFile;
     let defaultOptions = {};
     try {
-      defaultOptions = ConfigFile.getDefaultOptions();
+      defaultOptions = statics.getDefaultOptions();
     } catch (e) {
+      console.log(e);
```
shetzel: `this.logger.error()`? or should this be removed entirely?

PR #104, opened 2019-03-08
fix: support error.code for backwards compatibility

amphro:


src/sfdxError.ts
```diff
@@ -231,6 +236,11 @@ export class SfdxError extends NamedError {
   // Additional data helpful for consumers of this error.  E.g., API call result
   public data: any; // tslint:disable-line:no-any
 
+  /**
+   * Some errors support `error.code` instead of `error.name`. This keeps backwards compatability.
+   */
+  public code: string;
```
tnoonan-salesforce: I guess the weird thing would be doing a set on name and code will still have the old value.

I wonder if making code a javascript style getter where it always returns the current value of name unless code is set itself.

amphro: Good point. Ya, we could do that.

src/sfdxError.ts
```diff
@@ -205,6 +205,11 @@ export class SfdxError extends NamedError {
       sfdxError.stack = sfdxError.stack.replace(`${err.name}: ${err.message}`, 'Outer stack:');
       sfdxError.stack = `${err.stack}\n${sfdxError.stack}`;
     }
+
+    // If the original error has a code, use that instead of name.
+    if (hasString(err, 'code')) {
+      sfdxError.code = ensureString(getString(err, 'code'));
```
rbergman: No need to change this here, but when you really only need one of the type-specific fns here when "ensuring".  e.g. `ensure(getString(err, 'code'))` or `ensureString(get(err, 'code'))`.  I tend to prefer using the untyped `ensure` style when the type is inferrable from the arg.

rbergman: Oh actually can make this much simpler since `hasString` will re-type `err`: it can be `sfdxError.code = err.code`.

src/sfdxError.ts
rbergman: The comment makes this sound like should be optional?  `public code?: string`

src/sfdxError.ts
amphro: That does work! I could have swore I tried that.

Issue #143, opened 2019-06-20
SecureBuffer cipherName not included in the BoringSSL library for Electron 4


brpowell:
### Summary

Electron 4 will be shipping with BoringSSL instead of OpenSSL, and the former does not include the short name for aes-256-cbc. Currently, SecureBuffer is using the short name (aes256). This particularly affects the salesforce extensions for VS Code when using the sfdx-core library since VS Code stable will soon be running on Electron 4.

### Steps To Reproduce:
1.  Open the latest VS Code insiders with the Salesforce Extensions
1.  Authorize against a regular org
1.  Authorize against a scratch org
1.  Open the command palette and enter for "SFDX: Push source to default scratch org"

### Expected result

Should see the command present

### Actual result

Command is not present.

### Additional information

This is happening because the extensions rely on sfdx-core to determine what type of org the user is working against in order to correctly set the workspace org context.

**SFDX CLI Version**(to find the version of the CLI engine run sfdx --version):

sfdx-cli/7.11.2-4a5bf6aa97

**SFDX plugin Version**(to find the version of the CLI plugin run sfdx plugins --core):
```
@oclif/plugin-commands 1.2.2 (core)
@oclif/plugin-help 2.2.0 (core)
@oclif/plugin-not-found 1.2.2 (core)
@oclif/plugin-plugins 1.7.8 (core)
@oclif/plugin-update 1.3.9 (core)
@oclif/plugin-warn-if-update-available 1.7.0 (core)
@oclif/plugin-which 1.0.3 (core)
@salesforce/sfdx-trust 3.0.5 (core)
analytics 1.1.2 (core)
generator 1.1.0 (core)
salesforcedx 46.0.22 (pre-release)
├─ force-language-services 46.0.8
└─ salesforce-alm 46.0.18

sfdx-cli 7.11.2 (core)
```

**OS and version**:

macOS Mojave 10.14.5

amphro:
Released.

PR #144, opened 2019-06-20
fix: Update SecureBuffer cipherName to use aes-256-cbc

brpowell:
aes256 is the short name for aes-256-cbc. See https://wiki.openssl.org/index.php/Enc under Cypher Algorithms.

Issue: #143


tnoonan-salesforce:
This has to be tested on windows, mac, and linux before merging to master.

PR #157, opened 2019-07-17
fix: yarn.lock

rvwatch:


rvwatch:
Just a quick update to the yarn.lock file to point to npm registry rather than yarn registry. Proxy whitelist issue.

PR #164, opened 2019-08-12
fix: deleting and org auth file never worked

tnoonan-salesforce:


test/unit/orgTest.ts
```diff
@@ -710,12 +729,40 @@ describe('Org Tests', () => {
 
   describe('sandbox org config', () => {
     it('set field', async () => {
+      // Stub exists so only the auth file and sandbox config file exist. No users config file.
+      stubMethod($$.SANDBOX, ConfigFile.prototype, 'exists').callsFake(async function() {
```
khogan18: The delete should a different test.  The original test just handled verifying the set sandbox org config worked.

src/org.ts
```diff
@@ -487,6 +449,25 @@ export class Org extends AsyncCreatable<Org.Options> {
     return this.connection;
   }
 
+  /**
+   * Returns a promise to delete an auth info file from the local file system and any related cache information for
+   * this Org.. You don't want to call this method directly. Instead consider calling Org.remove()
```
khogan18: Can the method be made private?

src/org.ts
```diff
@@ -540,6 +532,66 @@ export class Org extends AsyncCreatable<Org.Options> {
...
+        let orgForUser;
+        if (username === this.getUsername()) {
+          orgForUser = this;
+        } else {
+          const _info = await AuthInfo.create({ username });
+          const connection: Connection = await Connection.create({ authInfo: _info });
+          orgForUser = await Org.create({ connection });
+        }
+
+        const orgType = this.isDevHubOrg() ? Config.DEFAULT_DEV_HUB_USERNAME : Config.DEFAULT_USERNAME;
```
khogan18: It's been a while since I've been through this code.  What is done differently for devhub vs 'regular' org?  We now have another org type with sandbox.

tnoonan-salesforce: It's cleaning up configs. For example, if the org in question is configured as the targetusername or defaultdevhubusername.

khogan18: 👍

PR #168, opened 2019-08-29
fix: add debug logger that accepts a function

tnoonan-salesforce:
[W-6509405](https://gus.my.salesforce.com/one/one.app#/alohaRedirect/a07B00000070ku5IAA)

### DIFF
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -588,6 +588,22 @@ export class Logger {
   }
 
   /**
+   * Logs at `debug` level with filtering applied.
+   *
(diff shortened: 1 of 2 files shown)


src/logger.ts
```diff
@@ -587,6 +587,22 @@ export class Logger {
     return this;
   }
 
+  /**
+   * Logs at `debug` level with filtering applied.
+   *
+   * @param cb A callback that returns on array objects to be logged.
```
rvwatch: *an array of objects* ?

PR #183, opened 2019-10-16
adding code for auth using auth code exchange

nramyasri-sf:


src/authInfo.ts
```diff
@@ -645,8 +647,10 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
         } else if (!options.authCode && options.refreshToken) {
           // refresh token flow (from sfdxUrl or OAuth refreshFn)
           authConfig = await this.buildRefreshTokenConfig(options);
+        } else if (options.authCode && options.clientId) {
+          authConfig = await this.authorize(options);
```
tnoonan-salesforce: Maybe add a comment that this is to support authCode exchange for access token.

tnoonan-salesforce:
Good job Ramya.

src/authInfo.ts
```diff
@@ -24,6 +25,7 @@ import {
 import { createHash, randomBytes } from 'crypto';
 import * as dns from 'dns';
 import { OAuth2, OAuth2Options, TokenResponse } from 'jsforce';
+import { Connection as JSConnection } from 'jsforce';
```
rvwatch: I'm wondering if it's possible to merge these imports into one line? I think I've seen this before. I could be wrong. 

_example:_
import { OAuth2, OAuth2Options, TokenResponse, Connection as JSConnection  } from 'jsforce';

PR #195, opened 2019-11-15
fix: rogue import causing compile issue

tnoonan-salesforce:


tnoonan-salesforce:
Fixes this: https://github.com/forcedotcom/sfdx-plugin-generate/issues/62

PR #200, opened 2019-12-30
Wr/mac keychain sync

williamruemmele-sf:
related to 3192 in FCT @W-5880614@

### DIFF
--- a/messages/crypto.json
+++ b/messages/crypto.json
@@ -1,0 +1,3 @@
+{
+  "MacKeychainOutOfSync": "\nWe’ve encountered an error with the Mac keychain being out of sync with your `sfdx` credentials. To fix the problem, sync your credentials by authenticating into your org again using the auth commands."
+}
--- a/src/crypto.ts
+++ b/src/crypto.ts
@@ -5,7 +5,7 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
-import { AsyncOptionalCreatable } from '@salesforce/kit';
+import { AsyncOptionalCreatable, env } from '@salesforce/kit';
 import { ensure, Nullable, Optional } from '@salesforce/ts-types';
 import * as crypto from 'crypto';
 import * as os from 'os';
@@ -111,17 +111,15 @@ export class Crypto extends AsyncOptionalCreatable<CryptoOptions> {
 
     const iv = crypto.randomBytes(BYTE_COUNT_FOR_IV).toString('hex');
 
-    return this._key.value(
-      (buffer: Buffer): string => {
-        const cipher = crypto.createCipheriv(_algo, buffer.toString('utf8'), iv);
+    return this._key.value((buffer: Buffer): string => {
+      const cipher = crypto.createCipheriv(_algo, buffer.toString('utf8'), iv);
 
-        let encrypted = cipher.update(text, 'utf8', 'hex');
-        encrypted += cipher.final('hex');
+      let encrypted = cipher.update(text, 'utf8', 'hex');
+      encrypted += cipher.final('hex');
 
-        const tag = cipher.getAuthTag().toString('hex');
-        return `${iv}${encrypted}${TAG_DELIMITER}${tag}`;
-      }
-    );
+      const tag = cipher.getAuthTag().toString('hex');
+      return `${iv}${encrypted}${TAG_DELIMITER}${tag}`;
+    });
   }
 
   /**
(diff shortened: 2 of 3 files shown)


src/crypto.ts
```diff
@@ -154,6 +152,11 @@ export class Crypto extends AsyncOptionalCreatable<CryptoOptions> {
         dec = decipher.update(secret, 'hex', 'utf8');
         dec += decipher.final('utf8');
       } catch (e) {
+        const useGenericUnixKeychain =
+          env.getBoolean('SFDX_USE_GENERIC_UNIX_KEYCHAIN') || env.getBoolean('USE_GENERIC_UNIX_KEYCHAIN');
+        if (os.platform() === 'darwin' && !useGenericUnixKeychain) {
+          e.message += Messages.loadMessages('salesforce-alm', 'crypto').getMessage('MacKeychainOutOfSync');
```
shetzel: Should be: `Messages.loadMessages('@salesforce/core', 'crypto');`

test/unit/cryptoTest.ts
```diff
@@ -130,5 +133,14 @@ describe('CryptoTest', function() {
...
     });
+
+    it('Decrypt should fail, error handling', () => {
+      const sandbox = sinon.createSandbox();
+      const message = Messages.loadMessages('@salesforce/core', 'crypto').getMessage('MacKeychainOutOfSync');
+      const errorMessage = SfdxError.wrap(new Error(message));
+      sandbox.stub(Crypto.prototype, 'decrypt').throws(errorMessage);
+      expect(() => Crypto.prototype.decrypt.call(secret)).to.throw(errorMessage);
+      sandbox.restore();
+    });
```
shetzel: You'll have to do something like this instead:
`const errMsg = 'decrypt failed';
stubMethod($$.SANDBOX, os, 'platform').returns('darwin');
stubMethod($$.SANDBOX, crypto, 'createDecipheriv').callsFake(() =>
      ({
        setAuthTag: () => throw new Error(errMsg),
        update: () => {},
        final: () => {}
      })
    );
crypto = new Crypto();
expect(crypto.decrypt('testing')).to.throw() // check that the error message is there`

Also would be worth adding another test with the env var set to make sure the error message isn't added.

messages/crypto.json
```diff
@@ -0,0 +1,3 @@
+{
+  "MacKeychainOutOfSync": "\nWe’ve encountered an error with the Mac keychain being out of sync with your `sfdx` credentials. To fix the problem, sync your credentials by logging into your Dev Hub"
```
amphro: I'm assuming this has been reviewed already, but it might be worth it be very specific on how. I wouldn't put it past a customer to ask. `...logging into your Dev Hub using the auth commands.`

src/crypto.ts
```diff
@@ -154,6 +152,11 @@ export class Crypto extends AsyncOptionalCreatable<CryptoOptions> {
         dec = decipher.update(secret, 'hex', 'utf8');
         dec += decipher.final('utf8');
       } catch (e) {
+        const useGenericUnixKeychain =
+          env.getBoolean('SFDX_USE_GENERIC_UNIX_KEYCHAIN') || env.getBoolean('USE_GENERIC_UNIX_KEYCHAIN');
+        if (os.platform() === 'darwin' && !useGenericUnixKeychain) {
+          e.message += Messages.loadMessages('@salesforce/core', 'crypto').getMessage('MacKeychainOutOfSync');
```
amphro: What does the error look like with the `+=`? It seems a little weird to have "<random crypto error>. We have encountered...". It might be worth it to just overwrite it, put it on a new line, put it in an error action, or something else to make it a little more readable. Or change the verbiage. Like "<random crypto error>. This is usually caused by...".

williamruemmele-sf: IF the message is added : `Failed to decipher auth data. reason: Unsupported state or unable to authenticate data
We’ve encountered an error with the Mac keychain being out of sync with your 'sfdx' credentials. To fix the problem, sync your credentials by logging into your Dev Hub.` vs `Failed to decipher auth data. reason: Unsupported state or unable to authenticate data`

messages/crypto.json
williamruemmele-sf: I think it was a deliberate choice from doc to avoid telling them what commands to use

amphro: I figured but figured "auth commands" was still generic enough. i.e. logging into your Dev Hub via the browser (without running a command) won't do anything, which a customer can read it as that.

src/crypto.ts
amphro: Ya, that seems weird to have the We've right after data with no punctuation or anything. 

```
Failed to decipher auth data. reason: Unsupported state or unable to authenticate data 

Try this: The Mac keychain might be out of sync with your 'sfdx' credentials. To fix the problem, sync your credentials by logging into your Dev Hub using an auth command.
```

Note, you get the `Try this` by adding an action to the error. That to me looks way more clear. I would run it by Terresa one more time.

messages/crypto.json
```diff
@@ -0,0 +1,3 @@
+{
+  "MacKeychainOutOfSync": "\nWe’ve encountered an error with the Mac keychain being out of sync with your `sfdx` credentials. To fix the problem, sync your credentials by logging into your Dev Hub using the auth commands."
```
amphro: I was about to merge this @williamruemmele-sf but then I noticed it said `by logging into your Dev Hub using...`. This can fail with any org though right? Maybe we should say, `by authenticating into your org again using...` or something like that?

williamruemmele-sf: yeah, I think 'authenticating' and 'logging in' are both conveying the same action - I can change that

PR #211, opened 2020-04-01
Update to use @salesforce/bunyan

lcampos:
Update the project to use @salesforce/bunyan which uses dayjs. DayJS being a lighter library compared to moment. This is the first step into making @salesforce/core compatible with bundler tools like webpack, rollup, etc.

@W-6980157@

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,11 +38,11 @@
     }
   },
   "dependencies": {
+    "@salesforce/bunyan": "^2.0.0",
     "@salesforce/kit": "^1.0.0",
     "@salesforce/ts-sinon": "^1.0.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -38,11 +38,11 @@
     }
   },
   "dependencies": {
+    "@salesforce/bunyan": "^2.0.0",
```
lcampos, line 41: @amphro looks like we still run tests for NodeJS 8 but I made changes to bunyan that specify it should be used on NodeJS 10 and above. Which version should be the min version we support here and on bunyan ?

amphro: We can change it to node10 and node12

PR #217, opened 2020-04-17
fix(security): deps security updates to be able to pass `yarn snyk test`

dhagberg-sf:
Downstream projects (sf-fx-sdk-nodejs) are failing snyk security checks.  Appears these can be resolved with the following updates to sfdx-core:

- mkdirp, its transitive dependency on minimist, yargs*, handlebars, and dot-prop all had Medium or High issues on Snyk
- Use `resolutions` block in package.json to force transitive dependency upgrades to versions indicated on Snyk reports
- Ensure tests all still pass
- Add yarn tasks and snyk devDependencies entries to match evergreen-cli project

Note: had to fix a a seemingly unrelated tsc type error in authInfo.ts.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,8 @@
     "lint-fix": "yarn sfdx-lint --fix",
     "postcompile": "tsc -p test; tsc -p typedocExamples",
     "prepack": "yarn sfdx-build",
+    "prepare": "yarn run snyk-protect",
+    "snyk-protect": "snyk protect",
     "test": "yarn sfdx-test"
   },
   "keywords": [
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -16,6 +16,8 @@
     "lint-fix": "yarn sfdx-lint --fix",
     "postcompile": "tsc -p test; tsc -p typedocExamples",
     "prepack": "yarn sfdx-build",
+    "prepare": "yarn run snyk-protect",
+    "snyk-protect": "snyk protect",
```
amphro, line 20: `snyk protect` changes dependencies right? I want CI to run on any new deps.

dhagberg-sf: It does not appear that `snyk protect` alters the dependency tree - only that it will apply patches that are known to snyk

<https://support.snyk.io/hc/en-us/articles/360000920818-What-is-the-difference-between-snyk-test-protect-and-monitor->
```
snyk protect is used to apply patches to your vulnerable dependencies.
It's useful after opening a fix pull request from our website (GitHub only)
or after running snyk wizard on the CLI. snyk protect reads a .snyk policy
file to determine what patches to apply.
```

None of those dependencies (transitive or direct) in the `resolutions` section had patches on the snyk website, but did have updated versions available.

What I'm not sure of is whether a downstream project like the sf-fx-sdk-nodejs or its resulting Functions will respect the `resolutions` in this project or if we need to repeat them in downstream projects.  Not sure how to test that without finding a way to do a local release of this project to pull in as a dependency -- any ideas?

dhagberg-sf:
@amphro --
- Added the bits for the latest `lodash` and `faye` vulnerabilities. 
- Is there a better way to do these version pins?  I'm happy to go down a different path.

amphro:
We were hoping to explore fixing them in the actual repos where the problems are coming from. I'll take another look at this and we can always merge this in while we continue to work on the actual offenders. I just don't want to lose track of it.

PR #225, opened 2020-05-01
fix: don't ship with ts-sinon

amphro:


sfryffdc:
What is the rationale around this? The code requires this (https://github.com/forcedotcom/sfdx-core/blob/develop/src/testSetup.ts), and is now broken.

Does this need to be a peer dependency?

amphro:
@sfryffdc The rationale was the consumer was always supposed to have this as a dev dependency when using this file, not bundled with core. The right pattern is to have this in a separate package like `@salesforce/core-test` which is what we will be moving to. We put this in a patch because it was blowing out some install sizes. Just do an install on ts-sinon in your plugin. Sorry for the inconvenience.

PR #230, opened 2020-05-07
WIP: Sh/sfdx project package dirs

shetzel:
Adds some config file caching so that instances are created once per path and files are not read more than once since they are kept up to date internally.  If you really want to read the file more than once pass `force=true` to read().

Adds sfdx-project.json schema validation.  By default a warning message is displayed when validation fails, but errors can be thrown if you set the env var, `SFDX_SCHEMA_VALIDATE`

Adds a method to SfdxProjectJson to get the packageDirectories and transform any path separators in the package dir paths to match the OS.

@W-7536443@

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -1,6 +1,6 @@
 ## Pre-requisites
 
-1.  We are using Node 12. If you need to work with multiple versions of Node, you
+1.  We are using Node 10 LTS. If you need to work with multiple versions of Node, you
     might consider using [nvm](https://github.com/creationix/nvm).
 1.  This repository uses [yarn](https://yarnpkg.com/) to manage node dependencies. Please install yarn globally using `npm install --global yarn`.
 
@@ -28,7 +28,7 @@ This compiles the typescript to javascript.
 
 ### `yarn clean`
 
-This cleans all generated files and directories. Run `yarn cleal-all` will also clean up the node_module directories.
+This cleans all generated files and directories. Run `yarn clean-all` to also clean up the node_module directories.
 
 ### `yarn test`
 
@@ -37,9 +37,9 @@ This tests the typescript using ts-node.
 ### `yarn lint`
 
 This lints all the typescript. If there are no errors/warnings
-from tslint, then you get a clean output. But, if they are errors from tslint,
+from tslint, then you get clean output. But, if there are errors from tslint,
 you will see a long error that can be confusing – just focus on the tslint
-errors. The results of this is deeper than what the tslint extension in VS Code
+errors. The results of this are deeper than what the tslint extension in VS Code
 does because of [semantic lint
 rules](https://palantir.github.io/tslint/usage/type-checking/) which requires a
 tsconfig.json to be passed to tslint.
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -1,6 +1,6 @@
 ## Pre-requisites
 
-1.  We are using Node 8. If you need to work with multiple versions of Node, you
+1.  We are using Node 12. If you need to work with multiple versions of Node, you
     might consider using [nvm](https://github.com/creationix/nvm).
 1.  This repository uses [yarn](https://yarnpkg.com/) to manage node dependencies. Please install yarn globally using `npm install --global yarn`.
 
@@ -17,7 +17,7 @@ You would only do this once after you cloned the repository.
 
 When you are ready to commit
 
-1.  We enforces commit message format. We recommend using [commitizen](https://github.com/commitizen/cz-cli) by installing it with `yarn add --global commitizen` then commit using `git cz` which will prompt you questions to format the commit message.
+1.  We enforce commit message format. We recommend using [commitizen](https://github.com/commitizen/cz-cli) by installing it with `yarn add commitizen --global` then commit using `git cz` which will prompt you questions to format the commit message.
 1.  Before commit and push, husky will run several hooks to ensure the message and that everything lints and compiles properly.
 
 ## List of Useful commands
@@ -36,7 +36,7 @@ This tests the typescript using ts-node.
 
 ### `yarn lint`
 
-This lists all the typescript. If there are no errors/warnings
+This lints all the typescript. If there are no errors/warnings
 from tslint, then you get a clean output. But, if they are errors from tslint,
 you will see a long error that can be confusing – just focus on the tslint
 errors. The results of this is deeper than what the tslint extension in VS Code
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -17,7 +17,7 @@ You would only do this once after you cloned the repository.
 
 When you are ready to commit
 
-1.  We enforce commit message format. We recommend using [commitizen](https://github.com/commitizen/cz-cli) by installing it with `yarn add commitizen --global` then commit using `git cz` which will prompt you questions to format the commit message.
+1.  We enforce commit message format. We recommend using [commitizen](https://github.com/commitizen/cz-cli) by installing it with `yarn global add commitizen` then commit using `git cz` which will prompt you questions to format the commit message.
 1.  Before commit and push, husky will run several hooks to ensure the message and that everything lints and compiles properly.
 
 ## List of Useful commands
--- a/src/config/configFile.ts
+++ b/src/config/configFile.ts
@@ -75,10 +75,8 @@ export class ConfigFile<T extends ConfigFile.Options> extends BaseConfigStore<T>
   // whether file contents have been read
   protected hasRead = false;
 
-  // Assigned when read
-  protected logger!: Logger;
-
   // Initialized in init
+  protected logger!: Logger;
   protected messages!: Messages;
 
   // Initialized in create
@@ -119,7 +117,6 @@ export class ConfigFile<T extends ConfigFile.Options> extends BaseConfigStore<T>
    * @param [force = false] Optionally force the file to be read from disk even when already read within the process.
    */
   public async read(throwOnNotFound = false, force = false): Promise<ConfigContents> {
-    this.logger = await Logger.child(this.constructor.name);
     try {
       // Only need to read config files once.  They are kept up to date
       // internally and updated persistently via write().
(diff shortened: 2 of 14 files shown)


DEVELOPING.md
```diff
@@ -17,7 +17,7 @@ You would only do this once after you cloned the repository.
 
 When you are ready to commit
 
-1.  We enforces commit message format. We recommend using [commitizen](https://github.com/commitizen/cz-cli) by installing it with `yarn add --global commitizen` then commit using `git cz` which will prompt you questions to format the commit message.
+1.  We enforce commit message format. We recommend using [commitizen](https://github.com/commitizen/cz-cli) by installing it with `yarn add commitizen --global` then commit using `git cz` which will prompt you questions to format the commit message.
```
amphro, line 20: Isn't the right way to do this `yarn global add commitizen` now? Probably doesn't matter.

src/config/configFile.ts
```diff
@@ -103,9 +110,16 @@ export class ConfigFile<T extends ConfigFile.Options> extends BaseConfigStore<T>
    * @param [throwOnNotFound = false] Optionally indicate if a throw should occur on file read.
    */
   public async read(throwOnNotFound = false): Promise<ConfigContents> {
```
amphro, line 112: Can we add another param here to `updateCache` or something? There will be times we want to update from the filesystem I would imagine.

src/config/configStore.ts
```diff
@@ -69,7 +69,7 @@ export abstract class BaseConfigStore<T extends BaseConfigStore.Options> extends
   public constructor(options: T) {
     super(options);
     this.options = options;
-    this.setContents(this.options.contents || {});
```
amphro, line 72: How come this isn't needed anymore? Done by `setContents`?

src/sfdxProject.ts
```diff
@@ -57,6 +96,10 @@ export class SfdxProjectJson extends ConfigFile<ConfigFile.Options> {
     if (upperCaseKey) {
       throw SfdxError.create('@salesforce/core', 'core', 'InvalidJsonCasing', [upperCaseKey, this.getPath()]);
     }
+
+    if (env.getBoolean('SFDX_SCHEMA_VALIDATE', false)) {
```
amphro, line 100: We could also call `this.schemaValidate` no matter what, but only throw if this is set to true, otherwise, display a warning?

src/sfdxProject.ts
```diff
@@ -78,6 +124,42 @@ export class SfdxProjectJson extends ConfigFile<ConfigFile.Options> {
...
+
+  /**
+   * Validates sfdx-project.json against the schema.
+   *
+   * ***See*** [sfdx-project.schema.json] (https://raw.githubusercontent.com/forcedotcom/schemas/master/schemas/sfdx-project.schema.json)
+   */
+  public async schemaValidate(): Promise<void> {
+    if (!this.hasRead) {
+      // read calls back into this method after setting this.hadRead = true
+      await this.read();
```
shetzel, line 136: This is a bit dangerous since `read()` conditionally calls `schemaValidate()` but `this.hasRead` should be true in that case.  It would only cause a problem if `read` were changed to catch any errors from the super and continue.  However, maybe throwing an error instead of reading here is safer.

src/sfdxProject.ts
```diff
@@ -78,6 +124,42 @@ export class SfdxProjectJson extends ConfigFile<ConfigFile.Options> {
...
+
+  /**
+   * Returns the `packageDirectories` within sfdx-project.json, first reading
+   * and validating the file if necessary.
+   */
+  public async getPackageDirectories(): Promise<PackageDir[]> {
+    // Ensure sfdx-project.json has first been read and validated.
+    if (!this.hasRead) {
+      await this.read();
+    }
```
shetzel, line 152: I need to add more validation here.  Since this is a new method I should be able to necessarily validate the file against the schema.

src/sfdxProject.ts
```diff
@@ -98,7 +180,13 @@ export class SfdxProject {
    * **Throws** *{@link SfdxError}{ name: 'InvalidProjectWorkspace' }* If the current folder is not located in a workspace.
    */
   public static async resolve(path?: string): Promise<SfdxProject> {
-    return new SfdxProject(await this.resolveProjectPath(path));
+    const _path = path || process.cwd();
```
amphro, line 183: 👍

src/sfdxProject.ts
```diff
@@ -78,6 +124,42 @@ export class SfdxProjectJson extends ConfigFile<ConfigFile.Options> {
...
+   * Returns the `packageDirectories` within sfdx-project.json, first reading
+   * and validating the file if necessary.
+   */
+  public async getPackageDirectories(): Promise<PackageDir[]> {
+    // Ensure sfdx-project.json has first been read and validated.
+    if (!this.hasRead) {
+      await this.read();
+    }
+
+    const contents = (this.getContents() as unknown) as ProjectJson;
```
shetzel, line 154: Is there a better way to do this?

src/sfdxProject.ts
```diff
@@ -98,7 +180,13 @@ export class SfdxProject {
    * **Throws** *{@link SfdxError}{ name: 'InvalidProjectWorkspace' }* If the current folder is not located in a workspace.
    */
   public static async resolve(path?: string): Promise<SfdxProject> {
-    return new SfdxProject(await this.resolveProjectPath(path));
+    const _path = path || process.cwd();
+    if (!SfdxProject.instances.has(_path)) {
+      const project = new SfdxProject(await this.resolveProjectPath(_path));
+      SfdxProject.instances.set(_path, project);
+    }
+    // @ts-ignore Because of the pattern above this is guaranteed to return an instance
```
amphro, line 188: 👍

src/sfdxProject.ts
shetzel, line 183: This should be ok (defaulting to `process.cwd()`) since it's done in `resolveProjectPath` anyway.  We need the path for caching.

src/sfdxProject.ts
```diff
@@ -4,18 +4,57 @@
...
-import { SfdxError } from './sfdxError';
+import { SchemaValidator } from './schema/validator';
 import { resolveProjectPath, SFDX_PROJECT_JSON } from './util/internal';
+
+// @ts-ignore
+import projectJsonSchema = require('@salesforce/schema/sfdx-project-schema.json');
+import { SfdxError } from './sfdxError';
 import { sfdc } from './util/sfdc';
 
+export interface PackageDirDependency {
```
amphro, line 23: Hmm, so how did you generate these? Via a tool? It would be great to keep them up-to-date with the schemas. Maybe the schemes package should also ship types....

amphro: Obviously outside the scope of this, but something that's would be useful I would think.

src/sfdxProject.ts
```diff
@@ -98,7 +180,13 @@ export class SfdxProject {
...
    */
   public static async resolve(path?: string): Promise<SfdxProject> {
-    return new SfdxProject(await this.resolveProjectPath(path));
+    const _path = path || process.cwd();
+    if (!SfdxProject.instances.has(_path)) {
+      const project = new SfdxProject(await this.resolveProjectPath(_path));
+      SfdxProject.instances.set(_path, project);
+    }
+    // @ts-ignore Because of the pattern above this is guaranteed to return an instance
+    return SfdxProject.instances.get(_path);
```
shetzel, line 189: `ensureInstance` wouldn't work here due to the private constructor for SfdxProject.

DEVELOPING.md
shetzel: That worked for me, but it's a dev dependency anyway so happy to update to whatever.

src/config/configFile.ts
shetzel: I couldn't think of usecases where we'd want to.  We're always up to date while the process is running.

src/config/configStore.ts
shetzel: Yeah, `setContents` does this so it's not needed.

src/sfdxProject.ts
shetzel: I generated them and massaged them a bit.  Yeah it would be nice to have better TS support with the schema lib.

src/sfdxProject.ts
shetzel: Yeah definitely.  I'll make that change.

DEVELOPING.md
```diff
@@ -1,6 +1,6 @@
 ## Pre-requisites
 
-1.  We are using Node 8. If you need to work with multiple versions of Node, you
+1.  We are using Node 12. If you need to work with multiple versions of Node, you
```
amphro, line 3: CLI isn't using 12 yet. Does this have a dep on 12 or does 10 still work?

src/sfdxProject.ts
```diff
@@ -78,6 +120,57 @@ export class SfdxProjectJson extends ConfigFile<ConfigFile.Options> {
...
+      // read calls back into this method after necessarily setting this.hasRead=true
+      await this.read();
+    } else {
+      try {
+        const projectJsonSchemaPath = require.resolve('@salesforce/schemas/sfdx-project.schema.json');
+        const validator = new SchemaValidator(this.logger, projectJsonSchemaPath);
+        await validator.load();
+        await validator.validate(this.getContents());
+      } catch (err) {
+        if (env.getBoolean('SFDX_SCHEMA_VALIDATE', false)) {
```
amphro, line 143: What are your thoughts about putting this in the schemaValidator class? The envar seems generic and i would expect it on all validation.

src/sfdxProject.ts
amphro: You can override.

First, use types instead of an interface.

```
export type PackageDir = {
...

export type ProjectJson = ConfigContents & {`
...
```

Then override the method.

```
public getContents(): ProjectJson {
    return super.getContents() as ProjectJson;
}
```

DEVELOPING.md
shetzel: 10 still works.  I think I was just being optimistic that we'd bundle and recommend 12, but not yet.  I'll change to 10 LTS.

src/sfdxProject.ts
shetzel: We can't put it there though because that could break existing users of the schema validator.  I'll change the name of the env var to be more specific to sfdx-project.json validation.  Maybe `SFDX_PROJECT_JSON_VALIDATION`

src/sfdxProject.ts
shetzel: I'll give that a try.  I like it better.  Thanks!

PR #234, opened 2020-06-03
feat: add fs.fileExists

mdonnalley:


src/util/fs.ts
```diff
@@ -158,5 +158,14 @@ export const fs = {
       encoding: 'utf8',
       mode: fs.DEFAULT_USER_FILE_MODE
     });
+  },
+
+  fileExists: async (filePath: string): Promise<boolean> => {
```
shetzel, line 163: Please add jsdoc.

PR #238, opened 2020-06-10
fix: multiple uncaught exception log fix @W-7558552@

xyc:
Fix multiple log lines when an uncaught exception happens.

There're multiple loggers printing logs when an uncaught exception happens, for example in local dev server:
```js
["sfdx", "Start", "Org", "AuthInfo", "crypto", "keyChain", "connection", "Org", "AuthInfo", "crypto", "keyChain", "connection", "TelemetryReporter", "AppInsights"]
```
The fix is to only log uncaught exception in root logger.

### DIFF
--- a/test/unit/authInfoTest.ts
+++ b/test/unit/authInfoTest.ts
@@ -1452,6 +1452,10 @@ describe('AuthInfo', () => {
   describe('audienceUrl', () => {
     const sfdxAudienceUrlSetting = process.env.SFDX_AUDIENCE_URL;
 
+    beforeEach(() => {
+      process.env.SFDX_AUDIENCE_URL = '';
+    });
+
     afterEach(() => {
       process.env.SFDX_AUDIENCE_URL = sfdxAudienceUrlSetting || '';
     });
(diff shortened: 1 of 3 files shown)


test/unit/authInfoTest.ts
```diff
@@ -1452,6 +1452,10 @@ describe('AuthInfo', () => {
   describe('audienceUrl', () => {
     const sfdxAudienceUrlSetting = process.env.SFDX_AUDIENCE_URL;
 
+    beforeEach(() => {
+      process.env.SFDX_AUDIENCE_URL = '';
```
xyc, line 1456: I mock this out since I have `process.env.SFDX_AUDIENCE_URL` set for internal env. Tests should not depend on the actual `process.env.SFDX_AUDIENCE_URL `

test/unit/loggerTest.ts
```diff
@@ -157,6 +157,17 @@ describe('Logger', () => {
...
       expect(rootLogger.addLogFileStream['called']).to.be.false;
     });
+
+    it('should log uncaught exception in root logger', async () => {
+      process.env.SFDX_ENV = 'dev';
+
+      const rootLogger = await Logger.root();
+      $$.SANDBOX.stub(rootLogger, 'fatal');
+
+      // @ts-ignore
```
shetzel, line 167: could you please add the reason for turning off TS compilation.  something like, "to access private properties for testing" or however you'd like to word it.  In the child test as well.  Thanks!

xyc: Sure, will do. Thanks for the feedback

PR #244, opened 2020-06-22
Added LifecycleEvents to Core

maxcamp:
In line with the efforts of __@W-5796702@__ to add command specific hooks to the Salesforce CLI, the event emitter and listener class Lifecycle has been added along with tests to this repo. This class is generic and does not have to be used only for hooks in commands. 

The tests are basic but thorough in demonstrating that every method performs as intended.

### DIFF
--- a/src/lifecycleEvents.ts
+++ b/src/lifecycleEvents.ts
@@ -11,9 +11,23 @@ import * as Debug from 'debug';
 type callback = (data: AnyJson) => Promise<void>;
 
 /**
- * An asynchronous event listener and emitter that follows the singleton pattern.
+ * An asynchronous event listener and emitter that follows the singleton pattern. The singleton pattern allows lifecycle events to be emitted from deep within a library and still be consumed by any other library or tool. It allows other developers to react to certain situations or events in your library without them having to manually call the method themselves.
+ *
+ * An example might be transforming metadata before it is deployed to an environment. As long as an event was emitted from the deploy library and you were listening on that event in the same process, you could transform the metadata before the deploy regardless of where in the code that metadata was initiated.
+ *
+ * @example
+ * ```
+ * // Listen for an event in a plugin hook
+ * Lifecycle.getInstance().on('deploy-metadata', transformMetadata)
+ *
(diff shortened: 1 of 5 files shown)


src/lifecycleEvents.ts
```diff
@@ -0,0 +1,71 @@
...
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { AnyJson, Dictionary } from '@salesforce/ts-types';
+import * as Debug from 'debug';
+
+type callback = (data: AnyJson) => Promise<void>;
+
+/**
+ * An asynchronous event listener and emitter that follows the singleton pattern.
```
amphro, line 14: Put some more context here with `@examples` like some of the other classes in force. Something like:

An asynchronous event listener and emitter that follows the singleton pattern. The singleton pattern allows lifecycle events to be emitted from deep within a library and still be consumed by any other library or tool. It allows other developers to react to certain situations or events in your library without them having to manually call the method themselves.

An example might be transforming metadata before it is deployed to an environment. As long as an event was emitted from the deploy library and you were listening on that event in the same process, you could transform the metadata before regardless of where in the code that metadata was initiated. 

@example
// Listen for an event in a plugin hook
Lifecycle.getInstance().on('deploy-metadata', transformMetadata)

// Deep in the deploy code, fire the event for all libraries and plugins to hear.
Lifecycle.getInstance().emit('deploy-metadata', metadataToBeDeployed);


Or something like that....

src/lifecycleEvents.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  private static instance: Lifecycle;
+  private debug = Debug(`sfdx:${this.constructor.name}`);
+  private listeners: Dictionary<callback[]>;
+
+  private constructor() {
+    this.listeners = {};
+  }
+
+  public removeAllListeners(eventName: string) {
```
amphro, line 32: Add doc comments to all public methods. I know it may look weird saying `Remove all listeners for a given event` when the method clearly communicates that, but I think it looks weird in the generated documentation and autocompletion if there isn't anything.

PR #246, opened 2020-06-25
Feat: adds actOn / areFilesEqual / getContentHash from toolbelt

rvwatch:
**What does this PR do?**
Brings in a few functions from toolbelt. 

- **actOn** - Recursively act on all files or directories in a directory
- **areFilesEqual** - Checks if files are the same
- **getContentHash** - Creates a hash for any string that is passed in

**Acceptance Criteria**
Unit tests should pass in both core and npm/yarn linked toolbelt using these functions. 

**Testing Notes**
yarn link from core --> toolbelt. Update actOn / areFilesEqual references to core and run tests.

### DIFF
--- a/src/util/fs.ts
+++ b/src/util/fs.ts
@@ -7,11 +7,14 @@
 
 import { parseJson, parseJsonMap } from '@salesforce/kit';
 import { AnyJson, JsonMap, Optional } from '@salesforce/ts-types';
+import * as crypto from 'crypto';
 import * as fsLib from 'fs';
 import * as mkdirpLib from 'mkdirp';
 import * as path from 'path';
 import { promisify } from 'util';
 import { SfdxError } from '../sfdxError';
+
+type PerformFunction = (filePath: string, file?: string, dir?: string) => Promise<void>;
 
 export const fs = {
   /**
@@ -172,5 +175,68 @@ export const fs = {
     } catch (err) {
       return false;
     }
+  },
+
+  /**
+   * Recursively act on all files or directories in a directory
+   *
+   * @param dir path to directory
+   * @param perform function to be run on contents of dir
+   * @param onType optional parameter to specify type to actOn
+   * @returns void
+   */
+
+  actOn: async (dir: string, perform: PerformFunction, onType: 'file' | 'dir' = 'file'): Promise<void> => {
+    for (const file of await fs.readdir(dir)) {
+      const filePath = path.join(dir, file);
+      const stat = await fs.stat(filePath);
+
+      if (stat) {
+        if (stat.isDirectory()) {
+          await fs.actOn(filePath, perform, onType);
+          if (onType === 'dir') {
+            await perform(filePath);
+          }
+        } else if (stat.isFile() && onType === 'file') {
+          await perform(filePath, file, dir);
+        }
+      }
+    }
+  },
+
+  /**
+   * Checks if files are the same
+   *
+   * @param file1Path the first file path to check
+   * @param file2Path the second file path to check
+   * @returns boolean
+   */
+  areFilesEqual: async (file1Path: string, file2Path: string): Promise<boolean> => {
+    try {
+      const file1Size = (await fs.stat(file1Path)).size;
+      const file2Size = (await fs.stat(file2Path)).size;
+      if (file1Size !== file2Size) {
+        return false;
+      }
+
+      const contentA = await fs.readFile(file1Path);
+      const contentB = await fs.readFile(file2Path);
+
+      return fs.getContentHash(contentA) === fs.getContentHash(contentB);
+    } catch (err) {
+      throw new SfdxError(`The path: ${err.path} doesn't exist or access is denied.`, 'DirMissingOrNoAccess');
+    }
+  },
+
+  /**
+   * Creates a hash for the string that's passed in
+   * @param contents The string passed into the function
+   * @returns string
+   */
+  getContentHash(contents: string | Buffer) {
+    return crypto
+      .createHash('sha1')
+      .update(contents)
+      .digest('hex');
   }
 };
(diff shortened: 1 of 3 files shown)


src/util/fs.ts
```diff
@@ -172,5 +175,74 @@ export const fs = {
...
+    try {
+      contentA = await fs.readFile(file1Path, 'utf8');
+    } catch (error) {
+      throw new SfdxError(`The path: ${file1Path} doesn\'t exist.`, 'DirMissingOrNoAccess');
+    }
+    try {
+      contentB = await fs.readFile(file2Path, 'utf8');
+    } catch (error) {
+      throw new SfdxError(`The path: ${file1Path} doesn\'t exist.`, 'DirMissingOrNoAccess');
+    }
```
WillieRuemmele, line 230: Can we combine these two try/catch? I bet somewhere in the error you can find the filepath, also the second try/catch should have the message `The path: ${file2Path} doesn\'t exist.`, 'DirMissingOrNoAccess' - you have file1path

test/unit/util/fsTest.ts
```diff
@@ -226,4 +226,83 @@ describe('util/fs', () => {
...
+    afterEach(() => {
+      $$.SANDBOX.restore();
+    });
+
+    it('should run custom functions against contents of a directory', async () => {
+      const actedOnArray = [];
+      const pathToFolder = pathJoin(osTmpdir(), 'foo');
+
+      const file1 = pathJoin(pathToFolder, 'test1.json');
+      const file2 = pathJoin(pathToFolder, 'test2.json');
```
WillieRuemmele, line 294: Does this cover the recursive call of 'actOn'?

test/unit/util/fsTest.ts
```diff
@@ -226,4 +226,83 @@ describe('util/fs', () => {
...
+        )
+        .onCall(1)
+        .resolves(
+          `{
+          "key": 12345,
+          "value": true,
+        }`
+        );
+      const exists = await fs.areFilesEqual('foo/bar.json', 'foo/bar2.json');
+      expect(exists).to.be.false;
```
WillieRuemmele, line 252: we should create files like below and use them in the 'areFilesEqual' call

src/util/fs.ts
rvwatch: Yep, good call!

test/unit/util/fsTest.ts
rvwatch: Agreed.

src/util/fs.ts
```diff
@@ -172,5 +175,70 @@ export const fs = {
...
+   * @param file1Path the first file path to check
+   * @param file2Path the second file path to check
+   * @returns boolean
+   */
+  areFilesEqual: async (file1Path: string, file2Path: string): Promise<boolean> => {
+    let contentA: string;
+    let contentB: string;
+
+    try {
+      contentA = await fs.readFile(file1Path, 'utf8');
```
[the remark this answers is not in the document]
rvwatch: Whoa. That's awesome!

test/unit/util/fsTest.ts
```diff
@@ -226,4 +226,71 @@ describe('util/fs', () => {
       expect(exists).to.be.false;
     });
   });
+
+  describe('areFilesEqual', () => {
+    afterEach(() => {
+      $$.SANDBOX.restore();
+    });
+
+    it('should return true if the files are the same', async () => {
```
amphro, line 235: I don't love writing files during tests... can we just stub out the fs calls to return data?

rvwatch: Wanted to put our chat here just in case someone else is curious. @williamruemmele-sf 

> why do you prefer stubbing files instead of writing them?

>Because, IMO, unit test should be testing as small amount of code as possible. Instead of just testing your method, you are also testing that the fs  library reads/writes the files from the disk. That also makes test take longer. Really you are assuming that fs is going to return you something, and your method is going to react to that. That is what you should be testing.

src/util/fs.ts
```diff
@@ -172,5 +175,67 @@ export const fs = {
...
+  },
+
+  /**
+   * Checks if files are the same
+   *
+   * @param file1Path the first file path to check
+   * @param file2Path the second file path to check
+   * @returns boolean
+   */
+  areFilesEqual: async (file1Path: string, file2Path: string): Promise<boolean> => {
```
amphro, line 214: After talking with Steve, we should also first stat the file and check the size. If the size is different, it is never equal which will be the majority of the case. That is a much much faster operation than hashing.

src/util/fs.ts
```diff
@@ -172,5 +175,77 @@ export const fs = {
...
+    try {
+      const file1Size = (await fs.stat(file1Path)).size;
+      const file2Size = (await fs.stat(file2Path)).size;
+      if (file1Size !== file2Size) {
+        return false;
+      }
+    } catch (err) {
+      throw new SfdxError(`no such file or directory, stat ${err.path}`, 'DirMissingOrNoAccess');
```
amphro, line 222: Why throw a different error message? You can really have just one try-catch. You can also have the last return in the try as well so you don't have to have the contentA/contentB outside of the try.

test/unit/util/fsTest.ts
```diff
@@ -226,4 +226,130 @@ describe('util/fs', () => {
...
+        .onCall(0)
+        .resolves({})
+        .onCall(1)
+        .resolves({});
+      const results = await fs.areFilesEqual('foo/bar.json', 'foo/bsar2.json');
+      expect(results).to.be.false;
+    });
+
+    it('should return error when fs.readFile throws error', async () => {
+      $$.SANDBOX.stub(fs, 'stat')
```
amphro, line 308: Great job on stubbing these out.

PR #253, opened 2020-07-10
fix: throw NoUsername error when missing required username

mdonnalley:
Throw `NoUsername` error instead of `AuthInfoCreationError` when required username is missing

@W-5907549@

### DIFF
--- a/messages/core.json
+++ b/messages/core.json
@@ -16,5 +16,6 @@
   "NoAliasesFound": "Nothing to set",
   "InvalidFormat": "Setting aliases must be in the format <key>=<value> but found: [%s]",
   "NoAuthInfoFound": "No authorization information can be found.",
-  "InvalidJsonCasing": "All JSON input must have heads down camelcase keys.  E.g., { sfdcLoginUrl: \"https://login.salesforce.com\" }\nFound \"%s\" at %s"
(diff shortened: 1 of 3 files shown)


messages/core.json
```diff
@@ -16,5 +16,6 @@
   "NoAliasesFound": "Nothing to set",
   "InvalidFormat": "Setting aliases must be in the format <key>=<value> but found: [%s]",
   "NoAuthInfoFound": "No authorization information can be found.",
-  "InvalidJsonCasing": "All JSON input must have heads down camelcase keys.  E.g., { sfdcLoginUrl: \"https://login.salesforce.com\" }\nFound \"%s\" at %s"
+  "InvalidJsonCasing": "All JSON input must have heads down camelcase keys.  E.g., { sfdcLoginUrl: \"https://login.salesforce.com\" }\nFound \"%s\" at %s",
+  "NoUsername": "This command requires an org username set either with a flag or by default in the config."
```
amphro, line 20: Technically `@salesforce/core` can be used without the CLI, so I think a now auth info found is still valid. I think we should wrap that error and throw this one in `@salesforce/command`.

mdonnalley: Sounds good. I'll close this PR and open one in `@salesforce/command` instead: https://github.com/forcedotcom/cli-packages/pull/123

PR #258, opened 2020-07-23
Wr/max query limit @W-7293174@

WillieRuemmele:
the logic and future proofing the maxQueryLimit config value

certain queries that use `autoFetchQuery` were being limited to 10k results, now you can override queries that use that method with a config value

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -6,14 +6,16 @@
  */
 
 import { maxBy, merge } from '@salesforce/kit';
-import { asString, ensure, isString, JsonCollection, JsonMap, Optional } from '@salesforce/ts-types';
-import { Tooling as JSForceTooling } from 'jsforce';
-import { ExecuteOptions } from 'jsforce';
-import { QueryResult } from 'jsforce';
-import { RequestInfo } from 'jsforce';
-import { ConnectionOptions } from 'jsforce';
-import { Connection as JSForceConnection } from 'jsforce';
-import { Promise as JsforcePromise } from 'jsforce';
+import { asString, ensure, getNumber, isString, JsonCollection, JsonMap, Optional } from '@salesforce/ts-types';
+import {
+  Connection as JSForceConnection,
+  ConnectionOptions,
+  ExecuteOptions,
+  Promise as JsforcePromise,
+  QueryResult,
+  RequestInfo,
+  Tooling as JSForceTooling
+} from 'jsforce';
 import { AuthFields, AuthInfo } from './authInfo';
 import { ConfigAggregator } from './config/configAggregator';
 import { Logger } from './logger';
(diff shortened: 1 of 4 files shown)


src/connection.ts
```diff
@@ -258,8 +261,13 @@ export class Connection extends JSForceConnection {
    * @param options The query options. NOTE: the autoFetch option will always be true.
    */
   public async autoFetchQuery<T>(soql: string, options: ExecuteOptions = {}): Promise<QueryResult<T>> {
+    const config: Config = await Config.create(Config.getDefaultOptions());
```
amphro, line 264: This needs to be `ConfigAggregator`.

src/connection.ts
```diff
@@ -269,13 +277,18 @@ export class Connection extends JSForceConnection {
       this.query<T>(soql, _options)
         .on('record', rec => records.push(rec))
         .on('error', err => reject(err))
-        .on('end', () =>
+        .on('end', () => {
+          if (records.length >= 10000 || limit < 10000) {
```
amphro, line 281: This seems wrong to me because `records.length` is only ever the size of records that are returned. It seems like what you want is the `totalSize` set on the query. Also if `totalSize` is 2000 and limit is 6000, that means there are no missing records.

You probably want something like:

```
const query = this.query<T>(soql, _options);
query.on('record', rec => records.push(rec))
query.on('error', err => reject(err))
query.on('end', () =>
    if (query.totalSize > records.length) { // This should give you if there are missing records.
```

**Note** that I didn't use limit to determine if here are missing records. I believe query is in batches of 2k records. So if you have a limit there are 6000 records, and you set the limit to 5000, it **will** return all 6k records.

src/connection.ts
```diff
@@ -269,13 +276,18 @@ export class Connection extends JSForceConnection {
       this.query<T>(soql, _options)
         .on('record', rec => records.push(rec))
         .on('error', err => reject(err))
-        .on('end', () =>
+        .on('end', () => {
+          if (records.length >= limit) {
```
amphro, line 280: I thought there was a `totalSize` on the query that would return the total number of records even if it wasn't retrieved. For example. I run a query and it matches 20k records on Salesforce. The first response will return a `totalSize` of 20k, with 2k records, and query locators to get the next 18k records in 9 more responses. That would mean if the limit was 10k, `totalSize` would be 20k, `records.length` would be 10k, and `limit` would be 10k. So the `totalSize > records.length` would tell is if missing data is there.

However, that isn't what I saw when I ran a few tests manually... so I must have something wrong or jsforce is masking something. In any case, this is close enough to equivalent that it is probably fine.

WillieRuemmele: I think that the `totalSize` field we're used to seeing in toolbelt is coming from below. `totalSize: records.length`

src/connection.ts
```diff
@@ -258,25 +260,31 @@ export class Connection extends JSForceConnection {
...
-            done: true,
-            totalSize: records.length,
-            records
-          })
-        )
-    );
+    let returnedRecords = 0;
+    const query = this.query<T>(soql, _options)
+      .on('record', () => returnedRecords++)
+      .on('error', err => Promise.reject(err))
```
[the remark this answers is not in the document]
WillieRuemmele: It's been removed

PR #259, opened 2020-07-24
Fixed config key validation and unset error handling

colbymatthiassonpublic:
1. Added config key validation for ISV_Debugger_URL and ISV_Debugger_ISD
2. Added Error handling for configStore unset

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -255,14 +255,30 @@ export class Config extends ConfigFile<ConfigFile.Options> {
           hidden: true,
           input: {
             // If a value is provided validate it otherwise no value is unset.
-            validator: value => isString(value) && sfdc.validateApiVersion(value),
+            validator: value => value == null || (isString(value) && sfdc.validateApiVersion(value)),
             failedMessage: Config.messages.getMessage('InvalidApiVersion')
           }
         },
         { key: Config.DEFAULT_DEV_HUB_USERNAME },
         { key: Config.DEFAULT_USERNAME },
-        { key: Config.ISV_DEBUGGER_SID, encrypted: true },
-        { key: Config.ISV_DEBUGGER_URL },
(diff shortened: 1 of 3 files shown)


src/config/configStore.ts
```diff
@@ -128,6 +129,11 @@ export abstract class BaseConfigStore<T extends BaseConfigStore.Options> extends
    * @param key The key.
    */
   public unset(key: string): boolean {
+    const property = Config.getAllowedProperties().find(allowedProp => allowedProp.key === key);
```
amphro, line 132: It feels weird to have this here but not in any of the other methods. What was the big rationale for moving this here? To keep it out of the config plugin?

colbymatthiassonpublic: Yes. Also config.set does this from within core (in config.ts) so I thought it would make sense to have a similar check for config.unset as well

src/config/config.ts
```diff
@@ -255,14 +255,30 @@ export class Config extends ConfigFile<ConfigFile.Options> {
...
-            validator: value => isString(value) && sfdc.validateApiVersion(value),
+            validator: value => value == null || (isString(value) && sfdc.validateApiVersion(value)),
             failedMessage: Config.messages.getMessage('InvalidApiVersion')
           }
         },
         { key: Config.DEFAULT_DEV_HUB_USERNAME },
         { key: Config.DEFAULT_USERNAME },
-        { key: Config.ISV_DEBUGGER_SID, encrypted: true },
-        { key: Config.ISV_DEBUGGER_URL },
+        {
```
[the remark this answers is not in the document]
colbymatthiassonpublic: DevHub username and default username have added validation checks from Org, so it seemed like ISV_Debugger were the only keys missing validation - not totally necessary though

PR #265, opened 2020-08-05
resolve project path sync

WillieRuemmele:
adds a `resolveProjectPathSync` method, which mimics `resolveProjectPath` but synchronously

### DIFF
--- a/test/unit/util/fsTest.ts
+++ b/test/unit/util/fsTest.ts
@@ -113,6 +113,43 @@ describe('util/fs', () => {
       expect(path).to.be.undefined;
     });
   });
+
+  describe('traverseForFileSync', () => {
+    let statFileStub;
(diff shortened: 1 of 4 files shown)


test/unit/util/fsTest.ts
```diff
@@ -5,6 +5,7 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 import { expect } from 'chai';
+import * as fsLib from 'fs';
```
nramyasri-sf, line 8: Shouldn't this be` import * as fsLib from 'src/util/fs'`?

WillieRuemmele: I could add `statSync` to `util/fs`

PR #272, opened 2020-08-10
feat: extend logger to suppport logfmt (key=value) format

elbandito:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@salesforce/core",
-  "version": "2.9.3",
+  "version": "2.9.4",
```
amphro, line 3: We use semantic release here so this automatically happens, along with the CHANGELOG.md.

src/logger.ts
```diff
@@ -315,10 +330,31 @@ export class Logger {
...
+    // If the log format is LOGFMT, we need to convert any stream(s) into a LOGFMT type stream.
+    if (this.format === LoggerFormat.LOGFMT && options.stream) {
+      const ls: LoggerStream = this.createLogFmtFormatterStream({ stream: options.stream });
+      options.stream = ls.stream;
+    }
+    if (this.format === LoggerFormat.LOGFMT && options.streams) {
+      const logFmtConvertedStreams: LoggerStream[] = [];
+      options.streams.forEach((ls: LoggerStream) => {
+        logFmtConvertedStreams.push(this.createLogFmtFormatterStream(ls));
+        options.streams = logFmtConvertedStreams;
```
amphro, line 347: Any reason this is in the loop rather than right after the loop?

PR #273, opened 2020-08-10
feat: support logfmt style formatting

elbandito:


src/logger.ts
```diff
@@ -315,10 +330,31 @@ export class Logger {
...
+    // If the log format is LOGFMT, we need to convert any stream(s) into a LOGFMT type stream.
+    if (this.format === LoggerFormat.LOGFMT && options.stream) {
+      const ls: LoggerStream = this.createLogFmtFormatterStream({ stream: options.stream });
+      options.stream = ls.stream;
+    }
+    if (this.format === LoggerFormat.LOGFMT && options.streams) {
+      const logFmtConvertedStreams: LoggerStream[] = [];
+      options.streams.forEach((ls: LoggerStream) => {
+        logFmtConvertedStreams.push(this.createLogFmtFormatterStream(ls));
+        options.streams = logFmtConvertedStreams;
```
amphro, line 347: Seems like this should be out of the loop.

elbandito: Good catch!  Fixed.

PR #275, opened 2020-08-12
fix: added fileExistsSync method to core

WillieRuemmele:


amphro:
How quickly do you need this?

WillieRuemmele:
No Major rush, the WI will still be blocked by DRL

tnoonan-salesforce:
@WillieRuemmele @amphro generally checking for a fileExisting is lack luster. It's better to check for read or write access instead.

amphro:
Closing in favor of #284

PR #283, opened 2020-08-25
feat: support array of messages when using getMessage

mdonnalley:
Support array of messages, e.g.

```json
"examples": [
    "sfdx config:get defaultusername",
    "sfdx config:get defaultusername defaultdevhubusername instanceUrl",
    "sfdx config:get defaultusername defaultdevhubusername --verbose"
  ]
```

### DIFF
--- a/src/messages.ts
+++ b/src/messages.ts
@@ -11,8 +11,18 @@
  */
 
 import { NamedError } from '@salesforce/kit';
-import { AnyJson, asString, ensureJsonMap, ensureString, isAnyJson, isObject, Optional } from '@salesforce/ts-types';
+import {
+  AnyJson,
+  asString,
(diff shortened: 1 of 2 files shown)


src/messages.ts
```diff
@@ -339,7 +348,13 @@ export class Messages {
...
-    const msg = ensureString(map.get(key));
-    return util.format(msg, ...tokens);
+    const msg = map.get(key);
+    const messages = (isArray(msg) ? msg : [msg]) as string[];
+    return messages
+      .map(message => {
+        ensureString(message);
+        return util.format(message, ...tokens);
+      })
+      .join('\n');
```
amphro, line 358: This should probably use os.EOL but probably doesn't matter all that much.

PR #220, opened 2020-04-23
feat: add scope to authorization url

amphro:

PR #289, opened 2020-08-28
feat: add scope options

amphro:


amphro:
This has already been reviewed in #220

PR #290, opened 2020-08-31
chore: add pr validation

mdonnalley:
### What does this PR do?
Updates config.yml to include PR validation

@W-7978550@


.circleci/config.yml
```diff
@@ -6,6 +6,10 @@ workflows:
   version: 2
   test-and-release:
     jobs:
+      - release-management/validate-pr:
```
amphro, line 9: What if we had this as a separate workflow?

PR #291, opened 2020-09-03
fix: query for username during refresh token flow

mdonnalley:
Queries for username when username is undefined in the `buildRefreshTokenConfig`. Required to support `sfdx auth:sfdxurl:store`

@W-7978524@


src/authInfo.ts
```diff
@@ -929,6 +934,29 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
     };
   }
 
+  private async queryForUsername(
```
amphro, line 937: nit: It's not really running a query. More like `retrieveUsername`.

amphro (approved):

shetzel (approved):

PR #292, opened 2020-09-04
feat: allow custom config properties

mdonnalley:
@W-7187253@

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -6,56 +6,15 @@
  */
 
 import { keyBy, set } from '@salesforce/kit';
-import { Dictionary, ensure, isNumber, isString } from '@salesforce/ts-types';
+import { Dictionary, ensure, isNumber, isString, Nullable } from '@salesforce/ts-types';
 import { Crypto } from '../crypto';
 import { Messages } from '../messages';
 import { SfdxError } from '../sfdxError';
 import { sfdc } from '../util/sfdc';
-import { ConfigFile } from './configFile';
+import { ConfigFile, ConfigPropertyMeta } from './configFile';
 import { ConfigContents, ConfigValue } from './configStore';
 
 const SFDX_CONFIG_FILE_NAME = 'sfdx-config.json';
-
-/**
- * Interface for meta information about config properties
- */
-export interface ConfigPropertyMeta {
-  /**
-   *  The config property name.
-   */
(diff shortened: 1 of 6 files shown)


src/config/configFile.ts
```diff
@@ -15,10 +15,51 @@ import { Messages } from '../messages';
...
 import { resolveProjectPath, resolveProjectPathSync } from '../util/internal';
-import { BaseConfigStore, ConfigContents } from './configStore';
+import { BaseConfigStore, ConfigContents, ConfigValue } from './configStore';
 
 Messages.importMessagesDirectory(pathJoin(__dirname));
 
+/**
+ * Interface for meta information about config properties
+ */
+export interface ConfigPropertyMeta {
```
amphro, line 25: I don't know if these belong in ConfigFile.

src/config/config.ts
```diff
@@ -184,68 +149,70 @@ export class Config extends ConfigFile<ConfigFile.Options> {
...
-            failedMessage: Config.messages.getMessage('InvalidBooleanConfigValue')
-          }
-        },
-        {
-          key: Config.MAX_QUERY_LIMIT,
-          input: {
-            validator: value => isNumber(value),
-            failedMessage: Config.messages.getMessage('InvalidNumberConfigValue')
-          }
+    Config.allowedProperties = [
```
amphro, line 152: It seems like we can run into problems when resetting this every time since this class doesn't cache. i.e. a plugin adds some config and another plugin creates one of these manually, will it clear out the other plugins properties?

src/config/config.ts
```diff
@@ -184,68 +149,70 @@ export class Config extends ConfigFile<ConfigFile.Options> {
...
+        key: Config.MAX_QUERY_LIMIT,
+        input: {
+          validator: value => isNumber(value),
+          failedMessage: Config.messages.getMessage('InvalidNumberConfigValue')
+        }
+      }
+    ];
+
+    if (options && options.customConfigProperties) {
+      Config.allowedProperties.push(...options.customConfigProperties);
```
amphro, line 215: I think we might want a global method to do this rather than in the constructor. That method might have to protect against plugin versions too... we will have to guard against version differences but I think we will run into more problems without it.

amphro (changes requested):

PR #293, opened 2020-09-09
feat: add sobject describe api support

jgellin-sf:
This PR adds a service for retrieving SObject metadata in a general purpose way. It is intended for consumption by upstream tooling, and to eventually replace several ad hoc solutions already in existence. 

Given a connection, the service makes a SObject [describe](https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/resources_sobject_describe.htm) request for a single SObject or [batched](https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/resources_composite_batch.htm) describe requests for multiple SObjects. The requests are slightly different, in that the single request optionally supports the "If-Modified-Since" header, which can improve performance for upstream consumers, while the batch requests cannot.


src/describe/sobjectApi.ts
```diff
@@ -0,0 +1,319 @@
...
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { RequestInfo } from 'jsforce';
+import { Connection } from '../connection';
+import { Logger } from '../logger';
+
+/* tslint:disable:no-any */
+export interface SObject {
```
lcampos, line 13: We should move the types/interfaces/enums to a types specific file so they can be published to consumers of the library.

src/describe/sobjectApi.ts
```diff
@@ -0,0 +1,319 @@
...
+  private readonly TARGETVERSION = '49.0';
+  private readonly VERSIONPREFIX = 'v';
+  private readonly SOBJECTS: string = 'sobjects';
+  private readonly BATCH: string = 'composite/batch';
+  private readonly DESCRIBE: string = 'describe';
+
+  private readonly CLIENT_ID: string = 'sfdx-vscode';
+
+  private readonly commonHeaders = {
+    'User-Agent': 'salesforcedx-extension',
```
lcampos, line 196:
```suggestion
    'User-Agent': 'salesforcedx-core',
```

Or something specific to this library, @amphro is there a user agent already used for this repo ?

src/describe/sobjectApi.ts
```diff
@@ -0,0 +1,319 @@
...
+  // URL constants
+  private readonly SERVICESPATH: string = 'services/data';
+  // the targetVersion should be consistent with the Cli even if only using REST calls
+  private readonly TARGETVERSION = '49.0';
+  private readonly VERSIONPREFIX = 'v';
+  private readonly SOBJECTS: string = 'sobjects';
+  private readonly BATCH: string = 'composite/batch';
+  private readonly DESCRIBE: string = 'describe';
+
+  private readonly CLIENT_ID: string = 'sfdx-vscode';
```
lcampos, line 193: Same as for user-agent, let's use something specific to the library

src/describe/sobjectApi.ts
```diff
@@ -0,0 +1,319 @@
...
+};
+export type BatchResponse = { hasErrors: boolean; results: SubResponse[] };
+
+export class SObjectDescribeAPI {
+  private readonly connection: Connection;
+
+  // URL constants
+  private readonly SERVICESPATH: string = 'services/data';
+  // the targetVersion should be consistent with the Cli even if only using REST calls
+  private readonly TARGETVERSION = '49.0';
```
amphro, line 187: Why are you hardcoding the version? This should just use `connection.getApiVersion`. https://github.com/forcedotcom/sfdx-core/blob/develop/src/connection.ts#L199

amphro: Also if you are part of the connection class, there are some methods that build part of this request for you, so you don't have to piece it all together, like services/data, etc.

test/unit/describe/sobjectApiTest.ts
```diff
@@ -0,0 +1,261 @@
...
+    expect(actualValue).equals(expectedValue);
+  });
+
+  it('should create the correct sobject describe request (when last refresh date not specified)', () => {
+    const sObjectName = 'George';
+    const expectedValue = {
+      method: 'GET',
+      url: sObjectAPI.doBuildSObjectDescribeURL(sObjectName, true),
+      headers: {
+        'Sforce-Call-Options': 'client=sfdx-vscode',
```
amphro, line 106: Based on the comments above, don't forget to remove these too.

amphro (changes requested):

src/describe/sobjectService.ts
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  public async describeSObjects(names: string[]): Promise<SObject[]> {
+    // TODO: make cancellable?
+    const describeAPI = new SObjectDescribeAPI(this.connection);
+    let fetchedResults: DescribeSObjectResult[] = [];
+    let j = 0;
+    while (j < names.length) {
+      try {
+        fetchedResults = fetchedResults.concat(await describeAPI.describeSObjectBatch(names, j));
```
xyc, line 35: @jgellin-sf Would you like to consider parallelizing fetching SObject as part of this effort?

We've found out that it would offer significant performance improvement (26.9s vs 3.6s). Work item W-8035201 has detailed description and link to sample code (just around 30 lines of changes). This is currently under my name but let me know you are interested in taking on it.

src/describe/sobjectApi.ts
```diff
@@ -0,0 +1,319 @@
...
+};
+export type SubResponse = { statusCode: number; result: SObject };
+export type BatchRawResponse = {
+  statusCode: number;
+  headers: { [key: string]: string };
+  body: string;
+};
+export type BatchResponse = { hasErrors: boolean; results: SubResponse[] };
+
+export class SObjectDescribeAPI {
```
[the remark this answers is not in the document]
jgellin-sf: @amphro I wrote these classes as an independently consumable library, and basically just moved them over. I had gotten used to these methods in this context, and so had not considered adding to Connection. I'll refactor these methods over there and we can see whether it makes sense.

src/describe/sobjectService.ts
```diff
@@ -0,0 +1,56 @@
...
+import { Connection } from '../connection';
+import { DescribeSObjectResult, SObject, SObjectDescribeAPI } from './sobjectApi';
+
+export enum SObjectType {
+  ALL,
+  STANDARD,
+  CUSTOM
+}
+
+export class SObjectService {
```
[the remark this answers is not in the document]
jgellin-sf: We haven't quite figured out whether the caching is going to happen in the service itself or as implemented by an upstream library. For the time being the SOQL tooling is only grabbing SObject metadata lazily, one object at a time, which is relatively quick, so we have kicked the cache can down the road, to when we have more information about what metadata would be useful to cache in which format, as well as which upstream tools are also going to make use of this interface and potential cache.

jgellin-sf:
Changes moved to jsforce library. https://github.com/jsforce/jsforce/pull/1063

PR #294, opened 2020-09-10
feat: support plugin-auth

mdonnalley:
### What does this PR do?
- add AuthRemover class
- add listAllAuthorizations method to AuthInfo
- add setAsDefault method to AuthInfo
- add setAlias method to AuthInfo

@W-7978524@

### DIFF
--- a/messages/auth.json
+++ b/messages/auth.json
@@ -1,0 +1,8 @@
+{
+  "defaultOrgNotFound": "No %s org found",
+  "defaultOrgNotFoundActions": [
+    "Run the \"sfdx auth\" commands with --setdefaultusername to connect to an org and set it as your default org.",
+    "Run \"force:org:create\" with --setdefaultusername to create a scratch org and set it as your default org.",
+    "Run \"sfdx config:set defaultusername=<username>\" to set your default username."
+  ]
+}
--- a/src/authInfo.ts
+++ b/src/authInfo.ts
@@ -28,8 +28,11 @@ import { OAuth2, OAuth2Options, TokenResponse } from 'jsforce';
 import * as Transport from 'jsforce/lib/transport';
 import * as jwt from 'jsonwebtoken';
 import { resolve as pathResolve } from 'path';
+import { basename, extname } from 'path';
 import { parse as urlParse } from 'url';
+import { Aliases } from './config/aliases';
 import { AuthInfoConfig } from './config/authInfoConfig';
+import { Config } from './config/config';
 import { ConfigAggregator } from './config/configAggregator';
 import { Connection, SFDX_HTTP_HEADERS } from './connection';
 import { Crypto } from './crypto';
@@ -66,6 +69,16 @@ export interface AuthFields {
   userProfileName?: string;
   expirationDate?: string;
 }
+
+export type Authorization = {
+  alias: Optional<string>;
+  username: Optional<string>;
+  orgId: Optional<string>;
+  instanceUrl: Optional<string>;
+  accessToken?: string;
+  oauthMethod?: 'jwt' | 'web' | 'token' | 'unknown';
+  error?: string;
+};
 
 /**
  * Options for access token flow.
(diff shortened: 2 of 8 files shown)


src/authInfo.ts
```diff
@@ -603,6 +659,44 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
...
+  public async setAsDefault(options: { defaultUsername?: boolean; defaultDevhubUsername?: boolean }) {
+    let config: Config;
+    // if we fail to create the local config, default to the global config
+    try {
+      config = await Config.create({ isGlobal: false });
+    } catch {
+      config = await Config.create({ isGlobal: true });
+    }
+
+    const value = ensureString(this.getFields().alias || this.getUsername());
```
amphro, line 677: Wait, `alias` is on the config fields?

src/authInfo.ts
```diff
@@ -603,6 +659,44 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
...
+  }
+
+  /**
+   * Sets the provided alias to the username
+   * @param alias alias to set
+   */
+  public async setAlias(alias: string) {
+    const authFields = this.getFields();
+    await Aliases.parseAndUpdate([`${alias}=${authFields.username}`]);
+    const updated = Object.assign({}, authFields, { alias });
```
amphro, line 696: I didn't think we stored the alias in the stored file... that seems wrong to me.

On a separate note, doesn't `this.update({ alias })` do all the merging for you?

src/config/configAggregator.ts
```diff
@@ -287,6 +287,9 @@ export class ConfigAggregator extends AsyncOptionalCreatable<JsonMap> {
    */
   private async loadProperties(): Promise<void> {
     this.resolveProperties(await this.globalConfig.read(), this.localConfig && (await this.localConfig.read()));
+    // const localConfig = (this.localConfig && await this.localConfig.exists()) ? await this.localConfig.read() : undefined;
```
amphro, line 290: Remove?

src/authRemover.ts
```diff
@@ -0,0 +1,230 @@
...
+ * const authConfigs = await remover.findAuthConfigs(
+ *  example@mycompany.com
+ * );
+ * const usernames = [...authConfigs.keys()];
+ * for (const username of usernames) {
+ *   await remover.removeAuth(username);
+ * }
+ * ```
+ */
+export class AuthRemover extends AsyncOptionalCreatable {
```
amphro, line 46: Nice job with this class. I like the API.

amphro (changes requested):

src/authInfo.ts
mdonnalley: @amphro It exists on the AuthFields interface (https://github.com/forcedotcom/sfdx-core/pull/294/files#diff-394ba1de20ae85e42959125d8ebc62b6R50) so it _could_ be there. However, I agree that the `Aliases` class should be the single source of truth for aliases. I'll change this accordingly

messages/auth.json
```diff
@@ -0,0 +1,8 @@
+{
+  "defaultOrgNotFound": "No %s org found",
+  "defaultOrgNotFoundAction": [
```
amphro, line 3: nit: Actions -< pluaral

src/authInfo.ts
```diff
@@ -603,6 +659,44 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
...
+    await config.write();
+  }
+
+  /**
+   * Sets the provided alias to the username
+   * @param alias alias to set
+   */
+  public async setAlias(alias: string) {
+    const username = this.getUsername();
+    await Aliases.parseAndUpdate([`${alias}=${username}`]);
```
amphro, line 697: You could add a new method called `Aliases.update(alias: string, value: string)` if you want. This method was mainly to help with command line arguments but I'm also fine if you just leave it like this.

amphro (approved):

src/authRemover.ts
shetzel: Yeah, this is awesome!

shetzel (approved):
👍

PR #311, opened 2020-10-01
fix: stub should call resolver at runtime

amphro:


amphro:
@WillieRuemmele this will fix the tests in toolbelt

WillieRuemmele (approved):

PR #313, opened 2020-10-06
feat: add WebOauthServer for web based auth flow

mdonnalley:
@W-7824743@


src/authInfo.ts
```diff
@@ -444,8 +444,8 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
    *
    * @param options The options to generate the URL.
    */
-  public static getAuthorizationUrl(options: OAuth2Options & { scope: string }): string {
-    const oauth2 = new OAuth2WithVerifier(options);
+  public static getAuthorizationUrl(options: OAuth2Options & { scope?: string }, oauth2?: OAuth2WithVerifier): string {
+    oauth2 = oauth2 || new OAuth2WithVerifier(options);
```
rorolopetegui, line 448: I think its better to not replace/reassign the parameters from the function, can we use a const here instead? (So we can maintain immutable the parameters)

shetzel (approved):

src/authInfo.ts
amphro: I personally think for optional params this is a fair pattern. I think the code is cleaner.

src/webOAuthServer.ts
```diff
@@ -0,0 +1,374 @@
...
+
+  /**
+   * sends a response error.
+   *
+   * @param statusCode he statusCode for the response.
+   * @param message the message for the http body.
+   * @param response the response to write the error to.
+   */
+  public sendError(status: number, message: string, response: http.ServerResponse): void {
+    response.statusMessage = message;
```
amphro, line 297: We should eventually provide more information here. In the past, most people will see `localhost:1717` in the URL bar and an error like 404 and assume that the localhost is the problem because they don't have a server running. Either provide additional information or redirect them to a page that has more information on oauth failures.

src/authInfo.ts
```diff
@@ -365,6 +369,17 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
     this.options = options;
   }
 
+  /**
+   * Returns the default instance url
+   *
+   * @returns {string}
+   */
+  public static async getDefaultInstanceUrl(): Promise<string> {
+    const config = await ConfigAggregator.create({});
```
amphro, line 378: Why don't you use the new sync method? `ConfigAggregator.getValue('instanceUrl').value`

src/authInfo.ts
```diff
@@ -255,11 +255,15 @@ function parseIdUrl(idUrl: string) {
 //
 // Ideally, this would be removed at some point in the distant future
 // when all auth files now have the clientId stored in it.
-const DEFAULT_CONNECTED_APP_INFO = {
+const LEGACY_DEFAULT_CONNECTED_APP_INFO = {
   legacyClientId: 'SalesforceDevelopmentExperience',
   legacyClientSecret: '1384510088588713504',
 };
 
+export const DEFAULT_CONNECTED_APP_INFO = {
```
amphro, line 263: I feel like the LEGACY in the above is redundant. You could have just had clientIf in the above object, so `DEFAULT_CONNECTED_APP_INFO.clientId` and `DEFAULT_CONNECTED_APP_INFO.legacyClientId`.

package.json
```diff
@@ -45,24 +45,28 @@
     "@salesforce/ts-types": "^1.0.0",
     "@types/graceful-fs": "^4.1.3",
     "@types/jsforce": "1.9.20",
+
```
amphro, line 48: remove

package.json
```diff
@@ -45,24 +45,28 @@
...
+    "axios": "^0.19.2",
     "debug": "^3.1.0",
     "graceful-fs": "^4.2.4",
     "jsen": "0.6.6",
     "jsforce": "1.9.3",
     "jsonwebtoken": "8.5.0",
     "mkdirp": "1.0.4",
+    "querystring": "^0.2.0",
```
amphro, line 56: How big is this library? Is it absolutely necessary?

package.json
```diff
@@ -45,24 +45,28 @@
     "@salesforce/ts-types": "^1.0.0",
     "@types/graceful-fs": "^4.1.3",
     "@types/jsforce": "1.9.20",
+
+    "axios": "^0.19.2",
```
amphro, line 49: Doesn't jsforce already make calls to `services/oauth2/token`? Why do we need to bring in another library to do this?

src/deviceOauthService.ts
```diff
@@ -0,0 +1,208 @@
...
+
+    this.logger.debug(`this.options.clientId: ${this.options.clientId}`);
+    this.logger.debug(`this.options.loginUrl: ${this.options.loginUrl}`);
+  }
+
+  private getLoginOptions(url: string): AxiosRequestConfig {
+    return {
+      url,
+      method: 'POST',
+      data: qs.stringify({
```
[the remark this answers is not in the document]
mdonnalley: @WillieRuemmele because it's converting the object into a query string, eg `?client_id=12345&grant_type=code`

amphro (approved):

PR #317, opened 2020-10-13
feat: cache config files by default

mdonnalley:
BREAKING CHANGE: config files are now cached after the first read

@W-7246647@


amphro (approved):
This is good to put in the v3 branch. I want too put together a list of changes that should go into the v3 branch.

PR #320, opened 2020-10-19
feat: allow setting allowedProperties on config

chadian:
## What does this PR do?
This PR adds `addAllowedProperties` static method to the `Config` class to allow the dynamic addition of `ConfigPropertyMeta` values. This supports `plugin-config` to be able load files containing the `ConfigPropertyMeta` definitions and support plugins defining their own keys (https://github.com/salesforcecli/plugin-config/pull/15)

* `addAllowedProperties` will only add `ConfigPropertyMeta` for `key` values that do not already exist in `allowedProperties`
* The default set of `allowedProperties` has been moved from the being set by the constructor and is now set on `Config` directly
* `propertyConfigMap` is changed to being a getter based on the `allowedProperties`

⚠️ Following the merging of this PR the following will need to be updated to the latest version of sfdx-core:
* `plugin-config` https://github.com/salesforcecli/plugin-config/pull/15
* `@salesforce/command`

## What issues does this PR fix or reference?
@W-7187253@

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -7,6 +7,7 @@
 
 import { keyBy, set } from '@salesforce/kit';
 import { Dictionary, ensure, isNumber, isString } from '@salesforce/ts-types';
+import { Logger } from '../logger';
 import { Crypto } from '../crypto';
 import { Messages } from '../messages';
 import { SfdxError } from '../sfdxError';
@@ -14,6 +15,7 @@ import { sfdc } from '../util/sfdc';
 import { ConfigFile } from './configFile';
 import { ConfigContents, ConfigValue } from './configStore';
 
+const log = Logger.childFromRoot('core:config');
 const SFDX_CONFIG_FILE_NAME = 'sfdx-config.json';
 
 /**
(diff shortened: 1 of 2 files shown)


src/config/config.ts
```diff
@@ -106,9 +106,13 @@ export class Config extends ConfigFile<ConfigFile.Options> {
    * allows users to override the 10,000 result query limit
    */
   public static readonly MAX_QUERY_LIMIT = 'maxQueryLimit';
-  private static allowedProperties: ConfigPropertyMeta[];
+
+  private static get propertyConfigMap(): Dictionary<ConfigPropertyMeta> {
+    return keyBy(Config.allowedProperties, 'key');
+  }
+
+  private static allowedProperties: ConfigPropertyMeta[] = [];
```
chadian, line 114: Sets the initial property value to `[]`.

src/config/config.ts
```diff
@@ -118,71 +122,70 @@ export class Config extends ConfigFile<ConfigFile.Options> {
...
+        input: {
+          validator: (value) => isNumber(value),
+          failedMessage: Config.messages.getMessage('InvalidNumberConfigValue'),
         },
-      ];
-    }
+      },
+    ];
+
+    Config.allowedProperties = [...(Config.allowedProperties || []), ...defaultConfigPropertyMetas];
```
chadian, line 187: Takes any external `allowedProperties` that have been set and adds the internal defaults, and sets the combined result on `allowedProperties`.

amphro: This seems a little weird to me. Consumers are not supposed to call the constructor directly, but there is a global and local config instance, which means that `allowedProperties` would have the defaults twice?

src/config/config.ts
```diff
@@ -106,83 +108,97 @@ export class Config extends ConfigFile<ConfigFile.Options> {
...
-            validator: (value) => value == null || (isString(value) && sfdc.isSalesforceDomain(value)),
-            failedMessage: Config.messages.getMessage('InvalidInstanceUrl'),
-          },
+  private static allowedProperties: ConfigPropertyMeta[] = [
+    {
+      key: 'instanceUrl',
+      input: {
+        // If a value is provided validate it otherwise no value is unset.
+        validator: (value) => value == null || (isString(value) && sfdc.isSalesforceDomain(value)),
+        get failedMessage() {
```
amphro, line 122: Out of curiosity, any reason you changed them to property getters?

src/config/config.ts
```diff
@@ -14,6 +15,7 @@ import { sfdc } from '../util/sfdc';
 import { ConfigFile } from './configFile';
 import { ConfigContents, ConfigValue } from './configStore';
 
+const debug = Debug('@salesforce/core:config');
```
amphro, line 18: We typically use Logger for debugging in this library (which will also delegate to the debug library). i.e. `Logger.childFromRoot('core:config')`

amphro (approved):

src/config/config.ts
chadian: https://github.com/forcedotcom/sfdx-core/pull/320/files#diff-615dacd7000fa06d7afef0f22b7b6d9267c6219fbfc9fa0315e51a5f5d963755R198-R200

Making them lazy allows them to pick up `Config.messages` being set in the constructor.

PR #327, opened 2020-11-11
fix: updated various logic and QOL on User and PermissionSetAssignment

WillieRuemmele:
Fix a few small behaviors with the User and PermissionSetAssignment classes. I don't think they were actually being used yet

### DIFF
--- a/src/user.ts
+++ b/src/user.ts
@@ -29,7 +29,7 @@ import { SecureBuffer } from './secureBuffer';
 import { SfdxError } from './sfdxError';
 import { sfdc } from './util/sfdc';
 
-const PASSWORD_LENGTH = 10;
+const PASSWORD_LENGTH = 13;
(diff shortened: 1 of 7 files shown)


src/user.ts
```diff
@@ -29,7 +29,7 @@ import { SecureBuffer } from './secureBuffer';
 import { SfdxError } from './sfdxError';
 import { sfdc } from './util/sfdc';
 
-const PASSWORD_LENGTH = 10;
+const PASSWORD_LENGTH = 13;
```
WillieRuemmele, line 32: when generating a password we were using `PASSWORD_LENGTH - ALL.length`, with `const ALL = [LOWER, UPPER, NUMBERS, SYMBOLS.join('')];` this was resulting in a password length of 6 with the minimum required being 8 and the server was throwing an error

mdonnalley (approved):

shetzel (approved):

PR #330, opened 2020-11-16
fix: single package entry is defaulted to default package @W-8222081@

WillieRuemmele:
if a there's only one package entry in `sfdx-project.json` and it doesn't include `default: true` we will now assume that it is the default

### DIFF
--- a/src/sfdxProject.ts
+++ b/src/sfdxProject.ts
@@ -248,8 +248,17 @@ export class SfdxProjectJson extends ConfigFile<ConfigFile.Options> {
       return Object.assign({}, packageDir, { name, path, fullPath });
     });
 
-    const defaultDirs = packageDirs.filter((packageDir) => packageDir.default);
+    // If we only have one package entry, it must be the default even if not explicitly labelled
+    if (packageDirs.length === 1) {
+      if (packageDirs[0].default === false) {
+        // we have one package but it is explicitly labelled as default=false
+        throw new SfdxError(this.messages.getMessage('SingleNonDefaultPackage'));
+      }
+      // add default=true to the package
+      packageDirs[0].default = true;
+    }
 
+    const defaultDirs = packageDirs.filter((packageDir) => packageDir.default);
     // Don't throw about a missing default path if we are in the global file.
     // Package directories are not really meant to be set at the global level.
     if (defaultDirs.length === 0 && !this.isGlobal()) {
--- a/test/unit/projectTest.ts
+++ b/test/unit/projectTest.ts
@@ -8,6 +8,7 @@ import { join, sep } from 'path';
 import { assert, expect } from 'chai';
 
 import { env } from '@salesforce/kit';
+import { SfdxError } from '../../src/exported';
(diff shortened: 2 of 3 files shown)


src/sfdxProject.ts
```diff
@@ -248,14 +248,19 @@ export class SfdxProjectJson extends ConfigFile<ConfigFile.Options> {
...
+      // Don't throw about a missing default path if we are in the global file.
+      // Package directories are not really meant to be set at the global level.
+      if (defaultDirs.length === 0 && !this.isGlobal() && packageDirs.length > 1) {
+        throw new SfdxError(this.messages.getMessage('MissingDefaultPath'));
+      } else if (defaultDirs.length > 1) {
+        throw new SfdxError(this.messages.getMessage('MultipleDefaultPaths'));
+      }
+    } else {
+      // add default=true to the package
+      packageDirs[0].default = true;
```
WillieRuemmele, line 263: Not sure if this is needed - but I don't want to alter consumer's expectations

mdonnalley (approved):

src/sfdxProject.ts
```diff
@@ -248,14 +248,19 @@ export class SfdxProjectJson extends ConfigFile<ConfigFile.Options> {
...
-    if (defaultDirs.length === 0 && !this.isGlobal()) {
-      throw new SfdxError(this.messages.getMessage('MissingDefaultPath'));
-    } else if (defaultDirs.length > 1) {
-      throw new SfdxError(this.messages.getMessage('MultipleDefaultPaths'));
+    // If we only have one package entry, it must be the default even if not explicitly labelled
+    if (packageDirs.length > 1) {
+      const defaultDirs = packageDirs.filter((packageDir) => packageDir.default);
+      // Don't throw about a missing default path if we are in the global file.
+      // Package directories are not really meant to be set at the global level.
+      if (defaultDirs.length === 0 && !this.isGlobal() && packageDirs.length > 1) {
```
amphro, line 256: Remove `packageDirs.length > 1` here since it is checked above.

src/sfdxProject.ts
amphro: We definitely want to set that if it isn't defined. A couple of things though. 

1. `packageDirs` may be empty here... so you should check for that case. This was handled before by throwing the MissingDefaultPath if the filter came back empty, which would return empty on an empty array.
1. What happens if there is only one `packageDirectories` entry but they set `default: false`?  Do we case? I feel like we should still throw in that case too because they were explicit about saying it wasn't the default. Thoughts?

test/unit/projectTest.ts
```diff
@@ -406,6 +406,25 @@ describe('SfdxProject', () => {
       expect(actual).to.deep.equal(expected);
     });
 
+    it('should set the a single package entry as default', async () => {
```
amphro, line 409: Add another test for one with no packages and you should see the index error I pointed out above.

amphro (changes requested):

amphro (approved):

PR #334, opened 2020-11-19
feat: support plugin-data

mdonnalley:
Adds support for plugin-data

@W-8236232@

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -6,7 +6,16 @@
  */
 
 import { maxBy, merge } from '@salesforce/kit';
-import { asString, ensure, getNumber, isString, JsonCollection, JsonMap, Optional } from '@salesforce/ts-types';
+import {
+  asString,
+  ensure,
+  getNumber,
+  isString,
+  JsonCollection,
+  JsonMap,
+  Optional,
+  AnyJson,
+} from '@salesforce/ts-types';
 import {
   Connection as JSForceConnection,
   ConnectionOptions,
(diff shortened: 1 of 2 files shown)


src/connection.ts
```diff
@@ -16,12 +25,70 @@ import {
...
+//   export interface Metric {
+//     requestPath: string;
+//     perfMetrics: AnyJson;
+//   }
+
+//   export interface Response {
+//     headers: AnyJson & { perfmetrics?: string };
+//     req: { path: string };
+//   }
+// }
```
WillieRuemmele, line 69: probably remove all of this

src/connection.ts
```diff
@@ -16,12 +25,70 @@ import {
   RequestInfo,
   Tooling as JSForceTooling,
 } from 'jsforce';
+// eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+// @ts-ignore
```
WillieRuemmele, line 29: why ts-ignore? Just no typings from jsforce? Just add a comment about why

WillieRuemmele (changes requested):

src/connection.ts
```diff
@@ -16,12 +25,70 @@ import {
...
+//     req: { path: string };
+//   }
+// }
+
+interface Response {
+  headers: AnyJson & { perfmetrics?: string };
+  req: { path: string };
+}
+
+// Unfortunately, we have to monkey patch HttpApi.prototype.request in order
```
amphro, line 76: I thought we talked about having this contained in the data plugin for now? Just have a `DataConnection extends Connection` class in the data plugin to do all this. For the auth info, you can always just create a new one if you need to.

amphro (changes requested):

src/connection.ts
mdonnalley: @amphro overriding the jsforce function doesn't work if I do it from plugin-data

WillieRuemmele (approved):

PR #338, opened 2020-12-05
Sm/jwt instance url

mshanemc:
[do not merge.  I want to thoroughly test this via the auth command to feel ok].

primary: ignoring the instanceUrl flag in auth:jwt:grant and referencing non-existent properties

- https://org62.lightning.force.com/lightning/r/0D50M00004IAqlISAT/view
- https://github.com/forcedotcom/cli/issues/217
- https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000006bEIUIA2/view
- https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008fosXIAQ/view

secondary: core prioritizing env var over flag

additional: there's usually a multi-minute delay replicating the the "generic" login servers (login.salesforce.com, test.salesforce.com) so attempting to jwt:grant right after creating a scratch org will fail.  By letting the jwt:grant command use the mydomain instanceUrl, the request goes straight to an org that can successfully auth immediately.

additional: we don't treat internal or public cloud instances differently.

### DIFF
--- a/src/authInfo.ts
+++ b/src/authInfo.ts
@@ -198,42 +198,6 @@ export enum SfdcUrl {
   PRODUCTION = 'https://login.salesforce.com',
 }
 
-const INTERNAL_URL_PARTS = [
-  '.internal.',
-  '.vpod.',
-  'stm.salesforce.com',
-  '.blitz.salesforce.com',
-  'mobile1.t.salesforce.com',
-];
-
-function isInternalUrl(loginUrl = ''): boolean {
-  return loginUrl.startsWith('https://gs1.') || INTERNAL_URL_PARTS.some((part) => loginUrl.includes(part));
-}
-
-function getJwtAudienceUrl(options: OAuth2Options) {
-  // default audience must be...
-  let audienceUrl: string = SfdcUrl.PRODUCTION;
-  const loginUrl = getString(options, 'loginUrl', '');
-  const createdOrgInstance = getString(options, 'createdOrgInstance', '').trim().toLowerCase();
-
-  if (process.env.SFDX_AUDIENCE_URL) {
-    audienceUrl = process.env.SFDX_AUDIENCE_URL;
-  } else if (isInternalUrl(loginUrl)) {
-    // This is for internal developers when just doing authorize;
-    audienceUrl = loginUrl;
-  } else if (
-    createdOrgInstance.startsWith('cs') ||
-    createdOrgInstance.endsWith('s') ||
-    urlParse(loginUrl).hostname === 'test.salesforce.com'
-  ) {
-    audienceUrl = SfdcUrl.SANDBOX;
-  } else if (createdOrgInstance.startsWith('gs1')) {
-    audienceUrl = 'https://gs1.salesforce.com';
-  }
-
-  return audienceUrl;
-}
-
 // parses the id field returned from jsForce oauth2 methods to get
 // user ID and org ID.
 function parseIdUrl(idUrl: string) {
@@ -921,12 +885,11 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
   // Build OAuth config for a JWT auth flow
   private async buildJwtConfig(options: OAuth2Options): Promise<AuthFields> {
     const privateKeyContents = await fs.readFile(ensure(options.privateKey), 'utf8');
-    const audienceUrl = getJwtAudienceUrl(options);
     const jwtToken = jwt.sign(
       {
         iss: options.clientId,
         sub: this.getUsername(),
-        aud: audienceUrl,
+        aud: options.loginUrl ?? process.env.SFDX_AUDIENCE_URL ?? SfdcUrl.PRODUCTION,
         exp: Date.now() + 300,
       },
       privateKeyContents,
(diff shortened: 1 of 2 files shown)


src/authInfo.ts
```diff
@@ -198,42 +198,6 @@ export enum SfdcUrl {
...
-
-function isInternalUrl(loginUrl = ''): boolean {
-  return loginUrl.startsWith('https://gs1.') || INTERNAL_URL_PARTS.some((part) => loginUrl.includes(part));
-}
-
-function getJwtAudienceUrl(options: OAuth2Options) {
-  // default audience must be...
-  let audienceUrl: string = SfdcUrl.PRODUCTION;
-  const loginUrl = getString(options, 'loginUrl', '');
-  const createdOrgInstance = getString(options, 'createdOrgInstance', '').trim().toLowerCase();
```
mshanemc, line 217: createdOrgInstance isn't even a member of `OAuth2Options`.  Using this getString seems to defeat the purpose of typescript.

src/authInfo.ts
```diff
@@ -198,42 +198,6 @@ export enum SfdcUrl {
...
-
-  if (process.env.SFDX_AUDIENCE_URL) {
```
mshanemc, line 219: flags should be a priority over env vars.  [Maybe we could use the oclif standard `env` option to default the flag to that?]

src/authInfo.ts
```diff
@@ -198,42 +198,6 @@ export enum SfdcUrl {
...
-    audienceUrl = process.env.SFDX_AUDIENCE_URL;
-  } else if (isInternalUrl(loginUrl)) {
-    // This is for internal developers when just doing authorize;
-    audienceUrl = loginUrl;
-  } else if (
-    createdOrgInstance.startsWith('cs') ||
```
mshanemc, line 225: all moot because createdOrgInstance doesn't exist

src/authInfo.ts
```diff
@@ -921,12 +885,11 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
   // Build OAuth config for a JWT auth flow
   private async buildJwtConfig(options: OAuth2Options): Promise<AuthFields> {
     const privateKeyContents = await fs.readFile(ensure(options.privateKey), 'utf8');
-    const audienceUrl = getJwtAudienceUrl(options);
     const jwtToken = jwt.sign(
       {
         iss: options.clientId,
         sub: this.getUsername(),
-        aud: audienceUrl,
+        aud: options.loginUrl ?? process.env.SFDX_AUDIENCE_URL ?? SfdcUrl.PRODUCTION,
```
mshanemc, line 892: without having to handle internal urls and looking at the created org instance, not sure we need this in other methods at all.  This hopefully makes clear the prioritization.

The original didn't seem to reference it for audiences, but I think we should include the loginUrl from the project file here, too: https://github.com/forcedotcom/schemas/blob/c0df4d670cd368d2cf760064983b8fa2ca04a4a8/sfdx-project.schema.json#L99

src/authInfo.ts
amphro: Ohh it is for sure getting past typescript =P But the majority of this code existed before typescript and I don't think we wanted to update the jsforce types since it is specific to us. Granted, we should have fixed the typing in this method and added a comment.

This is needed because of scratch org creations. 

In the header, you can change it too `function getJwtAudienceUrl(options: OAuth2Options & { createdOrgInstance?: string }) {` if you want to fix the typing.

src/authInfo.ts
amphro: The audience url is not a flag and is something completely different than what is passed in as the instanceUrl.

src/authInfo.ts
amphro: Not true, it exists during scratch org creations. Removing this will break scratch org authentication.

src/authInfo.ts
```diff
@@ -198,42 +198,6 @@ export enum SfdcUrl {
...
-  const createdOrgInstance = getString(options, 'createdOrgInstance', '').trim().toLowerCase();
-
-  if (process.env.SFDX_AUDIENCE_URL) {
-    audienceUrl = process.env.SFDX_AUDIENCE_URL;
-  } else if (isInternalUrl(loginUrl)) {
-    // This is for internal developers when just doing authorize;
-    audienceUrl = loginUrl;
-  } else if (
-    createdOrgInstance.startsWith('cs') ||
-    createdOrgInstance.endsWith('s') ||
```
amphro, line 226: This is needed for falcon.

src/authInfo.ts
```diff
@@ -198,42 +198,6 @@ export enum SfdcUrl {
...
-
-function getJwtAudienceUrl(options: OAuth2Options) {
-  // default audience must be...
-  let audienceUrl: string = SfdcUrl.PRODUCTION;
-  const loginUrl = getString(options, 'loginUrl', '');
-  const createdOrgInstance = getString(options, 'createdOrgInstance', '').trim().toLowerCase();
-
-  if (process.env.SFDX_AUDIENCE_URL) {
-    audienceUrl = process.env.SFDX_AUDIENCE_URL;
-  } else if (isInternalUrl(loginUrl)) {
```
amphro, line 221: This will break most localhost and internal environment usage.

src/authInfo.ts
amphro: I'm almost positive this order will not work for the majority of orgs.

amphro (changes requested):

mshanemc:
closing...I found a better way to do this

PR #339, opened 2020-12-08
fix: audiences use instanceURL

mshanemc:
@W-8543971@ primary: ignoring the instanceUrl flag in auth:jwt:grant and referencing non-existent properties

https://org62.lightning.force.com/lightning/r/0D50M00004IAqlISAT/view
forcedotcom/cli#217
https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000006bEIUIA2/view
https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008fosXIAQ/view

leaves the env as an override @amphro 

here's a file for QAing from plugin-auth.  requires a hub connected via jwt-key with some environment variables set.  https://github.com/salesforcecli/plugin-auth/blob/sm/qa-for-jwt-branches/test/commands/auth/jwt/grant.e2e.ts

### DIFF
--- a/src/authInfo.ts
+++ b/src/authInfo.ts
@@ -228,6 +228,7 @@ function getJwtAudienceUrl(options: OAuth2Options & { createdOrgInstance?: strin
     options.loginUrl?.includes('sandbox.my.salesforce.com') || // enhanced domains >= 230
     options.loginUrl?.match(/([Cc][Ss][0-9]+.my.salesforce.com)/g) || // my domains
     options.loginUrl?.match(/([Cc][Ss][0-9]+.salesforce.com)/g) || // instance without my domain
+    options.loginUrl?.match(/((usa)[0-9]+s\..+\.salesforce\.com)/g) || // falcon sandbox ex: usa2s.sfdc-whatever.salesforce.com
     (options.loginUrl && urlParse(options.loginUrl).hostname === 'test.salesforce.com')
   ) {
     return SfdcUrl.SANDBOX;
--- a/src/authInfo.ts
+++ b/src/authInfo.ts
@@ -226,7 +226,7 @@ function getJwtAudienceUrl(options: OAuth2Options & { createdOrgInstance?: strin
     createdOrgInstance.startsWith('cs') ||
     createdOrgInstance.endsWith('s') ||
     options.loginUrl?.includes('sandbox.my.salesforce.com') || // enhanced domains >= 230
-    options.loginUrl?.match(/(cs[0-9]+.my.salesforce.com)/g) ||
+    options.loginUrl?.match(/([Cc][Ss][0-9]+.my.salesforce.com)/g) ||
     (options.loginUrl && urlParse(options.loginUrl).hostname === 'test.salesforce.com')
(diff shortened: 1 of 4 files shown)


peternhale (approved):
@mshanemc Looks good to me. Just one thing you may want to check and nothing to do with the changes you are doing. But, there was a big impact announcement regarding URL changes that will be introduced as 230 rolls out. These may affect URL matching.

https://gus.lightning.force.com/lightning/r/0D5B000001EobxiKAB/view

shetzel (changes requested):
Can you please add a unit test or 2 to authInfoTest.ts for the changes?

mshanemc:
> Can you please add a unit test or 2 to authInfoTest.ts for the changes?

I couldn't think of any other domains to test for (there's a lot of permutations covered already) until I saw @peternhale 's super-useful catch about new domain formats coming.  I made some changes for those which required test examples.

shetzel (approved):

amphro (approved):

rorolopetegui (approved):
lgtm

src/authInfo.ts
```diff
@@ -229,7 +229,7 @@ function getJwtAudienceUrl(options: OAuth2Options & { createdOrgInstance?: strin
     createdOrgInstance.endsWith('s') ||
     loginUrlLowercased?.includes('sandbox.my.salesforce.com') || // enhanced domains >= 230
     loginUrlLowercased?.match(/(cs[0-9]+(\.my|)\.salesforce\.com)/g) || // my domains on CS instance OR CS instance without my domain
-    loginUrlLowercased?.match(/(usa[0-9]+s\..+\.salesforce\.com)/g) || // falcon sandbox ex: usa2s.sfdc-whatever.salesforce.com
+    loginUrlLowercased?.match(/((usa|ind|aus)[0-9]+s\..+\.salesforce\.com)/g) || // falcon sandbox ex: usa2s.sfdc-whatever.salesforce.com
```
peternhale, line 232: @mshanemc specifying region names is fragile. This will need to be adjusted over time.

peternhale (changes requested):

src/authInfo.ts
mshanemc: What's a good way to not do it?  How do I keep a prod url for some hipster vinyl business like vintage78s.my.salesforce.com from being treated as a sandbox because of falcon?

peternhale: That's a good point, but there has to be a way to distinguish a my domain URL from an instance login URL.

amphro: I agree, this is going to be a huge pain to update this list every time a new falcon instance is added.

I would rather to take our chances with the hipsters, or even something like `/([a-z]{3}[0-9]+s\..+\.salesforce\.com)/gi)` Note the `i` for case insenstivity.

mshanemc: Pete had more patterns over in slack (sfdc-something) that may help.  Stand by for another push.

peternhale (approved):

PR #340, opened 2020-12-08
fix(windows): throw error when file is not accessible

mdonnalley:
Throw an actual error object when windows can't access a file

Work Item: 
@W-8550536@


WillieRuemmele (approved):

amphro (approved):

amphro:
Doh. Test?

PR #344, opened 2020-12-15
Sm/acccesstoken works as username

mshanemc:
@W-7885939@

By fixing this accessToken issue in Core, it might also solve some other things like
- https://github.com/forcedotcom/cli/issues/290

### DIFF
--- a/src/authInfo.ts
+++ b/src/authInfo.ts
@@ -42,6 +42,7 @@ import { Global } from './global';
 import { Logger } from './logger';
 import { SfdxError, SfdxErrorConfig } from './sfdxError';
 import { fs } from './util/fs';
+import { sfdc } from './util/sfdc';
 
 /**
  * Fields for authorization, org, and local information.
@@ -758,8 +759,7 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
     this.fields.username = this.options.username || getString(options, 'username') || undefined;
 
     // If the username is an access token, use that for auth and don't persist
-    const accessTokenMatch = isString(this.fields.username) && this.fields.username.match(/^(00D\w{12,15})![.\w]*$/);
-    if (accessTokenMatch) {
+    if (isString(this.fields.username) && sfdc.matchesAccessToken(this.fields.username)) {
       // Need to initAuthOptions the logger and authInfoCrypto since we don't call init()
       this.logger = await Logger.child('AuthInfo');
       this.authInfoCrypto = await AuthInfoCrypto.create({
@@ -772,7 +772,7 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
       this.update({
         accessToken: this.options.username,
         instanceUrl,
-        orgId: accessTokenMatch[1],
+        orgId: this.fields.username.split('!')[0],
       });
 
       this.usingAccessToken = true;
--- a/src/user.ts
+++ b/src/user.ts
@@ -76,7 +76,11 @@ async function retrieveUserFields(logger: Logger, username: string): Promise<Use
   });
 
   if (sfdc.matchesAccessToken(username)) {
+    logger.debug('received an accessToken for the username.  Converting...');
     username = (await connection.identity()).username;
+    logger.debug(`accessToken converted to ${username}`);
+  } else {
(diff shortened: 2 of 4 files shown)


src/authInfo.ts
```diff
@@ -758,8 +759,7 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
     this.fields.username = this.options.username || getString(options, 'username') || undefined;
 
     // If the username is an access token, use that for auth and don't persist
-    const accessTokenMatch = isString(this.fields.username) && this.fields.username.match(/^(00D\w{12,15})![.\w]*$/);
-    if (accessTokenMatch) {
+    if (isString(this.fields.username) && sfdc.matchesAccessToken(this.fields.username)) {
```
mshanemc, line 762: used in multiple places, so moved to the sfdc utils

src/user.ts
```diff
@@ -75,6 +75,14 @@ async function retrieveUserFields(logger: Logger, username: string): Promise<Use
     authInfo: await AuthInfo.create({ username }),
   });
 
+  if (sfdc.matchesAccessToken(username)) {
```
WillieRuemmele, line 78: should resolving the username like this be used in creating the Connection above?

src/user.ts
```diff
@@ -75,6 +75,14 @@ async function retrieveUserFields(logger: Logger, username: string): Promise<Use
...
+    logger.debug('received an accessToken for the username.  Converting...');
+    username = (await connection.identity()).username;
+    logger.debug(`accessToken converted to ${username}`);
+  } else {
+    logger.debug('not a accessToken');
+  }
+
   const fromFields = Object.keys(REQUIRED_FIELDS).map(upperFirst);
```
peternhale, line 86: Not part of this change, but I do not see the sense in the transformation of required fields to name case.

@shetzel, is there a reason for this?

src/authInfo.ts
```diff
@@ -758,8 +759,7 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
     this.fields.username = this.options.username || getString(options, 'username') || undefined;
```
peternhale, line 759: Is last `|| undefined` necessary, given getString either returns the property or undefined?

src/util/sfdc.ts
```diff
@@ -111,4 +111,13 @@ export const sfdc = {
...
     return key;
   },
+
+  /**
+   * Tests whether a given string is an access token
+   *
+   * @param value
+   */
+  matchesAccessToken: (value: string): boolean => {
+    return value.match(/^(00D\w{12,15})![.\w]*$/) !== null;
```
peternhale, line 121: Could use `/^(00D\w{12,15})![.\w]*$/.test(value);`

peternhale (approved):
With some suggestions.

src/util/sfdc.ts
mshanemc: agree!

src/user.ts
mshanemc: I don't understand Auth well enough to know, so I'm limiting the change to fixing the bug.

It definitely works OK to get the Connection via accessToken.

src/authInfo.ts
shetzel: I agree that the `|| undefined` suffix looks unnecessary based on how `getString` works

src/user.ts
shetzel: No we must use the access token for the connection if it's passed in.  If we used the username it would fail because there's not associated auth file.

src/user.ts
shetzel: There was a reason for it at the time.  I'm guessing it was something to do with the User API.

shetzel (approved):

src/authInfo.ts
mshanemc: TS complains:
_Type 'Optional<string | null>' is not assignable to type 'string | undefined'.
  Type 'null' is not assignable to type 'string | undefined'._

going to leave it for now

shetzel: Ahh right.  `getString` returns `string | null` not `string | undefined`.  null !== undefined

PR #347, opened 2020-12-22
Sm/w 8206747

mshanemc:
@W-8206747@

I hate everything about this fix.  From what I can tell, the stuff leaking into the stdout isn't from us, it's from jsforce > request > asap

Screen Shot 2020-12-22 at 4 00 26 PM

Asap hasn't been updated in years and request is deprecated.  This'll happen whenever jsforce.request tries to use a domain that no longer exists and gets a DNS ENOTFOUND.

You can most easily repo by creating a sandbox, authing to it, and then deleting it, and then waiting for a while for it to be destroyed.

After a few tries, the best option I could think of is to pre-check a DNS using our existing domainResolver.  This required a number of updates to unit tests (where instanceUrl must exist to get past the pre-check).  It was either that or a serious rewrite the tests in a major way.

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -4,7 +4,7 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
-
+import { URL } from 'url';
 import { maxBy, merge } from '@salesforce/kit';
 import { asString, ensure, getNumber, isString, JsonCollection, JsonMap, Optional } from '@salesforce/ts-types';
 import {
@@ -16,7 +16,10 @@ import {
   RequestInfo,
   Tooling as JSForceTooling,
 } from 'jsforce';
+import { Duration } from '@salesforce/kit';
 import { AuthFields, AuthInfo } from './authInfo';
+import { MyDomainResolver } from './status/myDomainResolver';
+
 import { ConfigAggregator } from './config/configAggregator';
 import { Logger } from './logger';
 import { SfdxError } from './sfdxError';
@@ -119,8 +122,12 @@ export class Connection extends JSForceConnection {
 
     const conn = new this(options);
     await conn.init();
-    if (!versionFromConfig) {
-      await conn.useLatestApiVersion();
+    if (!versionFromConfig && conn.options?.connectionOptions?.instanceUrl) {
+      try {
+        await conn.useLatestApiVersion();
+      } catch (e) {
+        conn.logger.error(`Cannot connect to the org at ${conn.options?.connectionOptions?.instanceUrl}`);
+      }
     }
     return conn;
   }
@@ -178,11 +185,28 @@ export class Connection extends JSForceConnection {
    * Retrieves the highest api version that is supported by the target server instance.
    */
   public async retrieveMaxApiVersion(): Promise<string> {
-    type Versioned = { version: string };
-    const versions = (await this.request(`${this.instanceUrl}/services/data`)) as Versioned[];
-    this.logger.debug(`response for org versions: ${versions}`);
-    const max = ensure(maxBy(versions, (version: Versioned) => version.version));
-    return max.version;
+    // errors go to stdout from within jsforce (maybe even further down the stack) if org's url doesn't exist.  (not a 404 from the domain but a DNS ENOTFOUND)
+    // We need to verify that the org can be connected to BEFORE asking about it's api version.  This will throw on DNS errors
+    if (!this.options.connectionOptions?.instanceUrl) {
+      throw new Error('Connection has no instanceUrl');
+    }
+    const resolver = await MyDomainResolver.create({
+      url: new URL(this.options.connectionOptions.instanceUrl),
+      timeout: Duration.minutes(0),
+    });
+    const ipAddress = await resolver.resolve();
+
+    if (ipAddress) {
+      type Versioned = { version: string };
+      const versions = (await this.request(
+        `${this.options.connectionOptions.instanceUrl}/services/data`
+      )) as Versioned[];
+      this.logger.debug(`response for org versions: ${versions}`);
+      const max = ensure(maxBy(versions, (version: Versioned) => version.version));
+      return max.version;
+    }
+
+    throw new Error();
   }
   /**
    * Use the latest API version available on `this.instanceUrl`.
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -187,8 +187,9 @@ export class Connection extends JSForceConnection {
     await this.isResolvable();
     type Versioned = { version: string };
     const versions = (await this.request(`${this.instanceUrl}/services/data`)) as Versioned[];
-    this.logger.debug(`response for org versions: ${versions}`);
+    this.logger.debug(`response for org versions: ${versions.map((item) => item.version).join(',')}`);
     const max = ensure(maxBy(versions, (version: Versioned) => version.version));
+
     return max.version;
   }
   /**
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -38,6 +38,8 @@ export const SFDX_HTTP_HEADERS = {
   'user-agent': clientId,
 };
 
+const ORG_NOT_FOUND_ERROR_NAME = 'Org Not Found';
+
 // This interface is so we can add the autoFetchQuery method to both the Connection
 // and Tooling classes and get nice typing info for it within editors.  JSForce is
 // unlikely to accept a PR for this method, but that would be another approach.
@@ -196,7 +198,7 @@ export class Connection extends JSForceConnection {
     try {
       this.setApiVersion(await this.retrieveMaxApiVersion());
     } catch (err) {
-      if (err.name === 'Org Not Found') {
+      if (err.name === ORG_NOT_FOUND_ERROR_NAME) {
         throw err; // throws on DNS connection errors
       }
       // Don't fail if we can't use the latest, just use the default
@@ -222,7 +224,7 @@ export class Connection extends JSForceConnection {
       await resolver.resolve();
       return true;
     } catch (e) {
-      throw new SfdxError('The org cannot be found', 'Org Not Found', [
+      throw new SfdxError('The org cannot be found', ORG_NOT_FOUND_ERROR_NAME, [
         'Verify that the org still exists',
         'If your org is newly created, wait a minute and run your command again',
       ]);
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -227,7 +227,7 @@ export class Connection extends JSForceConnection {
       throw new SfdxError('The org cannot be found', DNS_ERROR_NAME, [
         'Verify that the org still exists',
         'If your org is newly created, wait a minute and run your command again',
-        "If you deployed or updated the org's My Domain, update your instanceUrl",
+        "If you deployed or updated the org's My Domain, logout from the CLI and authenticate again",
       ]);
     }
   }
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -5,7 +5,7 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 import { URL } from 'url';
-import { maxBy, merge } from '@salesforce/kit';
+import { Duration, maxBy, merge } from '@salesforce/kit';
 import { asString, ensure, getNumber, isString, JsonCollection, JsonMap, Optional } from '@salesforce/ts-types';
 import {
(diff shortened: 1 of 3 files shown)


src/connection.ts
```diff
@@ -119,8 +122,12 @@ export class Connection extends JSForceConnection {
 
     const conn = new this(options);
     await conn.init();
-    if (!versionFromConfig) {
-      await conn.useLatestApiVersion();
+    if (!versionFromConfig && conn.options?.connectionOptions?.instanceUrl) {
```
mshanemc, line 125: if there is no instance...I'm not sure how we'd check for its existence.

src/connection.ts
```diff
@@ -178,11 +185,28 @@ export class Connection extends JSForceConnection {
...
-    const versions = (await this.request(`${this.instanceUrl}/services/data`)) as Versioned[];
-    this.logger.debug(`response for org versions: ${versions}`);
-    const max = ensure(maxBy(versions, (version: Versioned) => version.version));
-    return max.version;
+    // errors go to stdout from within jsforce (maybe even further down the stack) if org's url doesn't exist.  (not a 404 from the domain but a DNS ENOTFOUND)
+    // We need to verify that the org can be connected to BEFORE asking about it's api version.  This will throw on DNS errors
+    if (!this.options.connectionOptions?.instanceUrl) {
+      throw new Error('Connection has no instanceUrl');
+    }
+    const resolver = await MyDomainResolver.create({
```
mshanemc, line 193: re-use existing domainResolver, because it has handling for internal/local urls that DNS may not like.

src/connection.ts
```diff
@@ -178,11 +185,28 @@ export class Connection extends JSForceConnection {
...
+      type Versioned = { version: string };
+      const versions = (await this.request(
+        `${this.options.connectionOptions.instanceUrl}/services/data`
+      )) as Versioned[];
+      this.logger.debug(`response for org versions: ${versions}`);
+      const max = ensure(maxBy(versions, (version: Versioned) => version.version));
+      return max.version;
+    }
+
+    throw new Error();
```
mshanemc, line 209: handled by the caller.  I could find no other references, but because it's public, there should be another bit of text in the Error saying what happened?

test/unit/connectionTest.ts
```diff
@@ -27,9 +30,19 @@ describe('Connection', () => {
     getConnectionOptions: () => testConnectionOptions,
   };
 
+  const testAuthInfoWithDomain = {
```
mshanemc, line 33: tweak of the original version to add instanceUrl

test/unit/connectionTest.ts
```diff
@@ -27,9 +30,19 @@ describe('Connection', () => {
...
+    getConnectionOptions: () => ({
+      ...testConnectionOptions,
+      instanceUrl: 'https://connectionTest/instanceUrl',
+    }),
+  };
+
   beforeEach(() => {
     $$.SANDBOXES.CONNECTION.restore();
     $$.SANDBOX.stub(process, 'emitWarning');
+    $$.SANDBOX.stub(MyDomainResolver.prototype, 'resolve').resolves(TEST_IP);
```
mshanemc, line 44: get past the DNS check without actually calling out

src/connection.ts
```diff
@@ -119,8 +122,12 @@ export class Connection extends JSForceConnection {
...
     const conn = new this(options);
     await conn.init();
-    if (!versionFromConfig) {
-      await conn.useLatestApiVersion();
+    if (!versionFromConfig && conn.options?.connectionOptions?.instanceUrl) {
+      try {
+        await conn.useLatestApiVersion();
+      } catch (e) {
+        conn.logger.error(`Cannot connect to the org at ${conn.options?.connectionOptions?.instanceUrl}`);
+      }
```
peternhale, line 130: useLatestApiVersion also has a try/catch and it does not rethrow, so this try/catch seems unnecessary.

src/connection.ts
```diff
@@ -178,11 +185,28 @@ export class Connection extends JSForceConnection {
...
-    const max = ensure(maxBy(versions, (version: Versioned) => version.version));
-    return max.version;
+    // errors go to stdout from within jsforce (maybe even further down the stack) if org's url doesn't exist.  (not a 404 from the domain but a DNS ENOTFOUND)
+    // We need to verify that the org can be connected to BEFORE asking about it's api version.  This will throw on DNS errors
+    if (!this.options.connectionOptions?.instanceUrl) {
+      throw new Error('Connection has no instanceUrl');
+    }
+    const resolver = await MyDomainResolver.create({
+      url: new URL(this.options.connectionOptions.instanceUrl),
+      timeout: Duration.minutes(0),
```
peternhale, line 195: Should there be a non-zero timeout?

peternhale (changes requested):

src/connection.ts
mshanemc: Did some updates after Steve's feedback and a lot more e2e testing.  That try/catch in `useLatestApiVersion` is now more error-specific for DNS problems.

src/connection.ts
mshanemc: done

mshanemc:
Looking better already!

Screen Shot 2020-12-29 at 5 54 11 PM

src/connection.ts
mshanemc: Not sure what other scenario would prevent us from being able to determine the version, though.

src/connection.ts
```diff
@@ -191,11 +198,39 @@ export class Connection extends JSForceConnection {
...
+   */
+  public async isResolvable(): Promise<boolean> {
+    if (!this.options.connectionOptions?.instanceUrl) {
+      throw new SfdxError('Connection has no instanceUrl', 'NoInstanceUrl', [
+        'Make sure the instanceUrl is set in your command or config',
+      ]);
+    }
+    const resolver = await MyDomainResolver.create({
+      url: new URL(this.options.connectionOptions.instanceUrl),
+      timeout: Duration.seconds(10),
```
peternhale, line 220: @shetzel is there a user controllable duration that can be applied here? I see env var `SFDX_DOMAIN_RETRY` but that seems specific to waits for Lightning domain resolution during scratch org create, so not sure that can reasonably be used here.

test/unit/connectionTest.ts
```diff
@@ -27,15 +30,36 @@ describe('Connection', () => {
...
+
     initializeStub = $$.SANDBOX.stub(jsforce.Connection.prototype, 'initialize').returns();
     requestMock = $$.SANDBOX.stub(jsforce.Connection.prototype, 'request')
       .onFirstCall()
       .returns(Promise.resolve([{ version: '42.0' }]));
   });
 
+  it('create() should throw on DNS errors', async () => {
```
peternhale, line 52: 👍

peternhale:
> Looking better already!
> 
> Screen Shot 2020-12-29 at 5 54 11 PM

@mshanemc, while the new error message "Org Not Found" is a great clarification, it does change the existing output. Should Claire/Juliet be consulted on this change to the output?

src/connection.ts
```diff
@@ -36,6 +38,8 @@ export const SFDX_HTTP_HEADERS = {
   'user-agent': clientId,
 };
 
+const ORG_NOT_FOUND_ERROR_NAME = 'Org Not Found';
```
peternhale, line 41: Please consider changing this to "Domain Not Found", which reflects the actual error encountered. While a low risk, it is possible that the org's domain name was changed and the local user json file reflects the old domain name in instance Url. [Rename Your My Domain](https://help.salesforce.com/articleView?id=domain_name_rename.htm&type=5)

src/connection.ts
```diff
@@ -191,11 +198,39 @@ export class Connection extends JSForceConnection {
...
+      url: new URL(this.options.connectionOptions.instanceUrl),
+      timeout: Duration.seconds(10),
+      frequency: Duration.seconds(10),
+    });
+    try {
+      await resolver.resolve();
+      return true;
+    } catch (e) {
+      throw new SfdxError('The org cannot be found', ORG_NOT_FOUND_ERROR_NAME, [
+        'Verify that the org still exists',
```
peternhale, line 228: Add hint regarding possibility of domain name rename.

src/connection.ts
mshanemc: That should not be used here--it'll delay results for no gain.  Newly created scratch orgs have a version so this code doesn't run.

I think "does DNS resolve?" is a more binary question where there's not a lot of value in retries.  I wanted the Resolver to not retry at all (that original zero-second version) but I don't think it supports that.

src/connection.ts
mshanemc: Good point.

src/connection.ts
peternhale: I believe the polling code waits after trying, so if the intent is to only try once, then zero is appropriate.

src/connection.ts
mshanemc: What would you say for them to do here?

"If you deployed or updated the org's My Domain, update your instanceUrl". <-- I'm not sure how to do that...editing auth files directly?  

You could modify `SFDX_INSTANCE_URL`...not sure how to apply it to a specific org  in a non-local-project context ( this bug is originally coming from`force:org:list`)

I don't think we have a clean command to modify an instance on an existing username auth file, right?

src/connection.ts
mshanemc: It didn't seem to work that way.  With `0` every org would come back with an error.

I've been using this https://github.com/lifeomic/attempt for a few years and have enjoyed its flexibility.

peternhale: 👍 for 10 seconds then.

src/connection.ts
peternhale: I think just the hint that the domain may have changed an logout/auth again.

BTW, where are these hints showing up? I did not see it in the sample output you shared.

mshanemc: > I think just the hint that the domain may have changed and logout/auth again.

Cool, will do.

> BTW, where are these hints showing up? I did not see it in the sample output you shared.

I guess that's up to whoever is using this.  In this bug, `force:org:list` applies `error.name` into the table.  I guess we *could* modify that command's output over in toolbelt, but it's not really part of the bug.

The hints are there because lots of commands call Connection.create() and might not be handling errors, in which case the hint would go all the way to the user (if I understand `sfdxError` properly).

peternhale: Makes sense.

src/connection.ts
```diff
@@ -17,6 +17,8 @@ import {
   Tooling as JSForceTooling,
 } from 'jsforce';
 import { AuthFields, AuthInfo } from './authInfo';
+import { MyDomainResolver } from './status/myDomainResolver';
+
```
peternhale, line 21: NIT: remove extra line

test/unit/connectionTest.ts
```diff
@@ -197,7 +226,9 @@ describe('Connection', () => {
   it('autoFetch() should reject the promise upon query error', async () => {
     const errorMsg = 'QueryFailed';
     requestMock.onSecondCall().throws(new Error(errorMsg));
-    const conn = await Connection.create({ authInfo: testAuthInfo as AuthInfo });
+    const conn = await Connection.create({
```
peternhale, line 229: Can these connection creates be safely moved to beforeEach?

peternhale (changes requested):

test/unit/connectionTest.ts
mshanemc: Their config (authInfo) varies a bit between tests, and would probably vary even more if we wanted to test more permutations.  So I'd say...probably not?

mshanemc: the other option would be organizing them in another describe by authInfo, and putting a beforeEach to cover everything at that level?

peternhale (approved):

src/connection.ts
```diff
@@ -185,28 +182,12 @@ export class Connection extends JSForceConnection {
...
-      this.logger.debug(`response for org versions: ${versions}`);
-      const max = ensure(maxBy(versions, (version: Versioned) => version.version));
-      return max.version;
-    }
-
-    throw new Error();
+    await this.isResolvable();
+    type Versioned = { version: string };
+    const versions = (await this.request(`${this.instanceUrl}/services/data`)) as Versioned[];
+    this.logger.debug(`response for org versions: ${versions}`);
```
shetzel, line 190: I think this will output something like, `response for org versions: [object Object], [object Object]...`  Maybe we just want to output the max?

shetzel (approved):
Approve pending comment resolution

src/connection.ts
mshanemc: now, it logs all versions as an array of string

PR #349, opened 2020-12-28
Wrong coverage reported to codecov

rorolopetegui:
## What does this PR do?
This repo is uploading wrong coverage to codecov

## Way to test
Run yarn test, check if the coverage is bigger than codecov.

## What issues does this PR fix or reference?
W-5542402

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "postcompile": "tsc -p test; tsc -p typedocExamples",
     "prepack": "sf-build",
     "pretest": "sf-compile-test",
-    "test": "sf-test"
+    "test": "sf-test --require ts-node/register"
   },
   "keywords": [
     "force",


package.json
```diff
@@ -17,7 +17,7 @@
     "postcompile": "tsc -p test; tsc -p typedocExamples",
     "prepack": "sf-build",
     "pretest": "sf-compile-test",
-    "test": "sf-test"
+    "test": "sf-test --require ts-node/register"
```
peternhale, line 20: @rorolopetegui This project defines a .mocharc.json, which has a require property that includes ts-node/register. Is sf-test not picking up that file?

rorolopetegui: Sorry for late response. It was not picking up, the solution that i found for this was just by using the require flag.

shetzel (approved):

PR #356, opened 2021-01-20
Windows env added to circle ci

rorolopetegui:
@W-8623693 - Windows env added to Circle Ci
It runs the test flow using circleci/windows orb


rorolopetegui:
Going to close this PR, instead i'll add the win job on the orb that we have

PR #362, opened 2021-01-22
feat: adjustable DNS timeout

mshanemc:
After a bunch of real-world testing, 3 generally behaves better than 10, so it should be the default.

I'd rather that be adjustable in the ENV because it might not be right for slower connections?

This a follow-on to @W-8206747@

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -5,7 +5,7 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 import { URL } from 'url';
-import { Duration, maxBy, merge } from '@salesforce/kit';
+import { Duration, maxBy, merge, Env } from '@salesforce/kit';
 import { asString, ensure, getNumber, isString, JsonCollection, JsonMap, Optional } from '@salesforce/ts-types';
 import {
   Connection as JSForceConnection,
@@ -39,7 +39,7 @@ export const SFDX_HTTP_HEADERS = {
 };
 
 export const DNS_ERROR_NAME = 'Domain Not Found';
-const DNS_TIMEOUT = parseInt(process.env.SFDX_DNS_TIMEOUT ?? '3', 10) || 3;
+const DNS_TIMEOUT = Math.min(3, new Env().getNumber('SFDX_DNS_TIMEOUT') || 3);
 
 // This interface is so we can add the autoFetchQuery method to both the Connection
 // and Tooling classes and get nice typing info for it within editors.  JSForce is
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -39,6 +39,7 @@ export const SFDX_HTTP_HEADERS = {
 };
 
 export const DNS_ERROR_NAME = 'Domain Not Found';
+const DNS_TIMEOUT = process.env.SFDX_DNS_TIMEOUT ? parseInt(process.env.SFDX_DNS_TIMEOUT, 10) : 3;
 
 // This interface is so we can add the autoFetchQuery method to both the Connection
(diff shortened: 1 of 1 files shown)


src/connection.ts
```diff
@@ -39,6 +39,7 @@ export const SFDX_HTTP_HEADERS = {
 };
 
 export const DNS_ERROR_NAME = 'Domain Not Found';
+const DNS_TIMEOUT = process.env.SFDX_DNS_TIMEOUT ? parseInt(process.env.SFDX_DNS_TIMEOUT, 10) : 3;
```
peternhale, line 42: @mshanemc parseInt will return NaN if the value is invalid. Are there any downstream effects that need to be considered?

Also should there be some sort of reasonableness check for range of allowable values?

peternhale (changes requested):

src/connection.ts
mshanemc: NaN...good point.  I'm fixing that.

Reasonableness...is that a thing that any other env do?  Or do we just let users decide what's best?

peternhale: @mshanemc Huge debate on how much latitude to give customers for these choices. I do know that negative numbers should not be allowed and previous experiments with 0 timeout did not yield good results. You have also determined that 3 seconds is a good default and maybe a good minimum value.

Upper boundary is the fuzziest to me, but I think it deserves to be set. Maybe set it to 30 seconds, the default for [domain resolver](https://github.com/forcedotcom/sfdx-core/blob/sm/adjustable-dns-timeout/src/status/myDomainResolver.ts#L89) or 60.

Guiding principle for me is, "How do we keep the customer from calling support to get help extracting themselves from a bad decision, either our bad decision or theirs?"

Is isResolvable function's limited right now? Seems it would be useful when `force:org:create` moves to OSS. There are similar checks being done when waiting for a scratch org domains to resolve, but wait times there can be in the minutes and there is no current upper bound.

My suggestion is `3<=SFDX_DNS_TIMEOUT<=60 (maybe unbounded to support future reuse)`

peternhale: @amphro Not to make this more difficult, but what is the policy WRT adding new env vars versus adding a new well known configuration property?

mshanemc: The only current use of isResolvable is used is to prevent that stdout-leak in request/asap that you raised previously on `org:list`.  Optimistically, future changes to jsforce and/or its dependencies might make it uneeded.

it's re-using MyDomainResolver, which is what other commands should use, too.  That has some logic for short-circuiting on local/internal instances, etc.

My intent with the env is to have Juliet doc it if we decide to go that way.

mshanemc: If we did want to bound it, how do we do that?

1. override the env set by the customer, and give them a warning?
2. throw on out-of-bounds and ask them to change it?

peternhale: @mshanemc Given that we hope that this code is short lived and can be retired, I think the simplest thing to do would be override and write a message to the log file only. I don't think this code has access to ux, so to inform customer directly would require throwing an error.

src/connection.ts
```diff
@@ -39,6 +39,7 @@ export const SFDX_HTTP_HEADERS = {
 };
 
 export const DNS_ERROR_NAME = 'Domain Not Found';
+const DNS_TIMEOUT = Math.min(3, new Env().getNumber('SFDX_DNS_TIMEOUT') || 3);
```
peternhale, line 42: @mshanemc should be `Math.max`. Also getNumber supports supplying a default with second parameter.

peternhale (changes requested):

peternhale (approved):

PR #363, opened 2021-01-26
refactor: move isInternalUrl to sfdc

mshanemc:
@W-8792196@ 

centralizes internal domains and exports via sfdc for all plugins to use

### DIFF
--- a/src/util/sfdc.ts
+++ b/src/util/sfdc.ts
@@ -120,4 +120,21 @@ export const sfdc = {
   matchesAccessToken: (value: string): boolean => {
     return /^(00D\w{12,15})![.\w]*$/.test(value);
   },
+
+  /**
+   * Tests whether a given url is an internal Salesforce domain
+   *
+   * @param url
+   */
+  isInternalUrl: (url: string): boolean => {
+    const INTERNAL_URL_PARTS = [
+      '.internal.',
+      '.vpod.',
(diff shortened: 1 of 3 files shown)


amphro (approved):

src/util/sfdc.ts
```diff
@@ -120,4 +120,21 @@ export const sfdc = {
...
+   * Tests whether a given url is an internal Salesforce domain
+   *
+   * @param url
+   */
+  isInternalUrl: (url: string): boolean => {
+    const INTERNAL_URL_PARTS = [
+      '.internal.',
+      '.vpod.',
+      'stm.salesforce.com',
+      '.blitz.salesforce.com',
```
peternhale, line 134: I don't think these test environments are available anymore; blitz* and mobile*. Blitz* was replaced with STEAM stmf* envs and mobile* with STREAM stmp* envs.

peternhale (approved):

src/util/sfdc.ts
mshanemc: should we update them?  Have you got a list like the others you sent me for falcon domains?

peternhale: I searched configs and blitz is still present, but mobile is not. I say remove mobile, then QA.

peternhale (approved):

PR #365, opened 2021-01-27
fix: send oauth error to browser

amphro:
Right now, the websever oauth flow will kill the server if an auth code exchange fails, leaving a no response in the browser. This change sends the error to the browser instead.

image

To reproduce an auth code problem, you can deny a connected app or hack the code to introduce a character to the auth code.

@W-8789899@

### DIFF
--- a/messages/auth.json
+++ b/messages/auth.json
@@ -10,5 +10,6 @@
   "invalidRequestMethod": "Invalid request method: %s",
   "invalidRequestUri": "Invalid request uri: %s",
   "pollingTimeout": "The device authorization request timed out. After executing force:auth:device:login, you must approve access to the device within 10 minutes.  This can happen if the URL wasn’t copied into the browser, login was not attempted, or the 2FA process was not completed within 10 minutes. Request authorization again.",
-  "ServerErrorHTMLResponse": "<h1>%s</h1><br />This is most likely <b>not</b> an error with the Salesforce CLI. Please ensure all information is accurate and try again."
+  "serverErrorHTMLResponse": "<h1>%s</h1><br />This is most likely <b>not</b> an error with the Salesforce CLI. Please ensure all information is accurate and try again.",
+  "missingAuthCode": "No authentication code found on login response."
 }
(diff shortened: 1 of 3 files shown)


src/webOAuthServer.ts
```diff
@@ -189,6 +195,7 @@ export class WebOAuthServer extends AsyncCreatable<WebOAuthServer.Options> {
         this.logger.debug(`Successfully obtained auth code: ...${authCode.substring(authCode.length - 5)}`);
       } else {
         this.logger.debug('Expected an auth code but could not find one.');
+        throw SfdxError.create('@salesforce/core', 'auth', 'missingAuthCode');
```
amphro, line 198: I'm not sure why we would ever continue without an authcode. You get `Value of undefined` if there isn't an auth code.

peternhale (approved):

mdonnalley (approved):

messages/auth.json
```diff
@@ -9,5 +9,7 @@
   "PortInUseAction": "Kill the process running on port %s or use a custom connected app and update OauthLocalPort in the sfdx-project.json file.",
   "invalidRequestMethod": "Invalid request method: %s",
   "invalidRequestUri": "Invalid request uri: %s",
-  "pollingTimeout": "The device authorization request timed out. After executing force:auth:device:login, you must approve access to the device within 10 minutes.  This can happen if the URL wasn’t copied into the browser, login was not attempted, or the 2FA process was not completed within 10 minutes. Request authorization again."
+  "pollingTimeout": "The device authorization request timed out. After executing force:auth:device:login, you must approve access to the device within 10 minutes.  This can happen if the URL wasn’t copied into the browser, login was not attempted, or the 2FA process was not completed within 10 minutes. Request authorization again.",
+  "serverErrorHTMLResponse": "<h1>%s</h1><br />This is most likely <b>not</b> an error with the Salesforce CLI. Please ensure all information is accurate and try again.",
```
jshackell-sfdc, line 13: Edit: "This is most likely <b>not</b> an error with Salesforce CLI. Ensure all information is accurate and try again."

PR #366, opened 2021-01-28
fix: myDomainResolver shouldn't check localhost

mshanemc:
@W-8818195@

### DIFF
--- a/src/status/myDomainResolver.ts
+++ b/src/status/myDomainResolver.ts
@@ -13,6 +13,7 @@ import { ensureString } from '@salesforce/ts-types';
 
 import { AsyncOptionalCreatable, Duration } from '@salesforce/kit';
 import { Logger } from '../logger';
+import { sfdc } from '../util/sfdc';
 import { StatusResult } from './client';
 import { PollingClient } from './pollingClient';
 
@@ -65,7 +66,7 @@ export class MyDomainResolver extends AsyncOptionalCreatable<MyDomainResolver.Op
 
         try {
           self.logger.debug(`Attempting to resolve url: ${host}`);
-          if (host?.includes('.internal.salesforce.com') || host?.includes('localhost')) {
+          if (sfdc.isLocalUrl(host)) {
             return {
               completed: true,
               payload: '127.0.0.1',
--- a/src/status/myDomainResolver.ts
+++ b/src/status/myDomainResolver.ts
@@ -65,7 +65,7 @@ export class MyDomainResolver extends AsyncOptionalCreatable<MyDomainResolver.Op
 
         try {
           self.logger.debug(`Attempting to resolve url: ${host}`);
-          if (host && host.includes('.internal.salesforce.com')) {
+          if (host?.includes('.internal.salesforce.com') || host?.includes('localhost')) {
             return {
               completed: true,
               payload: '127.0.0.1',
--- a/src/util/sfdc.ts
+++ b/src/util/sfdc.ts
@@ -127,7 +127,13 @@ export const sfdc = {
    * @param url
    */
   isInternalUrl: (url: string): boolean => {
-    const INTERNAL_URL_PARTS = ['.internal.', '.vpod.', 'stm.salesforce.com', '.blitz.salesforce.com'];
+    const INTERNAL_URL_PARTS = [
+      '.internal.',
+      '.vpod.',
+      'stm.salesforce.com',
+      '.blitz.salesforce.com',
+      'localhost.sfdcdev.',
+    ];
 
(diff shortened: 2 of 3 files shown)


src/status/myDomainResolver.ts
```diff
@@ -65,7 +65,7 @@ export class MyDomainResolver extends AsyncOptionalCreatable<MyDomainResolver.Op
 
         try {
           self.logger.debug(`Attempting to resolve url: ${host}`);
-          if (host && host.includes('.internal.salesforce.com')) {
+          if (host?.includes('.internal.salesforce.com') || host?.includes('localhost')) {
```
peternhale, line 68: This is scattering knowledge about localhost between here and isInternalUrl.

Can this be consolidated?

src/util/sfdc.ts
```diff
@@ -127,7 +127,13 @@ export const sfdc = {
    * @param url
    */
   isInternalUrl: (url: string): boolean => {
-    const INTERNAL_URL_PARTS = ['.internal.', '.vpod.', 'stm.salesforce.com', '.blitz.salesforce.com'];
+    const INTERNAL_URL_PARTS = [
```
peternhale, line 130: Maybe add a function that qualifies localhost URLs

peternhale (changes requested):

src/status/myDomainResolver.ts
mshanemc: yes...but I'm not sure they're exactly the same.  This is more isLocal...link things we don't want to dns check.

you can have internal steam/vpod/etc that still need to verify DNS resolution, correct?

src/util/sfdc.ts
```diff
@@ -127,7 +127,13 @@ export const sfdc = {
...
+      '.internal.',
+      '.vpod.',
+      'stm.salesforce.com',
```
peternhale, line 133: I am seeing different URLs for steam this morning on stmfa

on dev hub `https://inttestdevhub02-dev-ed.lightning.stmfa.stm.force.com/`

peternhale (changes requested):

src/status/myDomainResolver.ts
mshanemc: proposed: isLocal, which is both exported for use by myDomainResolver and plugins AND used by isInternal

src/util/sfdc.ts
mshanemc: so much fun going on

src/status/myDomainResolver.ts
peternhale: @mshanemc internal, non local need to be checked, but having an isLocal check for just localhost seems reasonable to me.

src/util/sfdc.ts
```diff
@@ -127,8 +127,19 @@ export const sfdc = {
    * @param url
    */
   isInternalUrl: (url: string): boolean => {
-    const INTERNAL_URL_PARTS = ['.internal.', '.vpod.', 'stm.salesforce.com', '.blitz.salesforce.com'];
+    const INTERNAL_URL_PARTS = ['.vpod.', 'stm.salesforce.com', 'stm.force.com', '.blitz.salesforce.com'];
+    return (
+      url.startsWith('https://gs1.') || sfdc.isLocalUrl(url) || INTERNAL_URL_PARTS.some((part) => url.includes(part))
```
amphro, line 132: Does anyone remember the context of `gs1` check? It feels so out of place to me.

src/util/sfdc.ts
```diff
@@ -127,8 +127,19 @@ export const sfdc = {
...
+    );
+  },
 
-    return url.startsWith('https://gs1.') || INTERNAL_URL_PARTS.some((part) => url.includes(part));
+  /**
+   * Tests whether a given internal url runs on a local machine
+   *
+   * @param url
+   */
+  isLocalUrl: (url: string): boolean => {
```
amphro, line 141: I like this a lot better.

amphro (approved):

src/util/sfdc.ts
mshanemc: kept it because it was here: https://github.com/forcedotcom/sfdx-core/blob/10fc422fe87b2ba2fb940abd07d0742f78f044f3/src/authInfo.ts#L231

for jwt audience stuff

src/status/myDomainResolver.ts
mshanemc: updated

src/util/sfdc.ts
mshanemc: added that, too.

peternhale (approved):

PR #368, opened 2021-02-02
fix: add space option to write json

amphro:
@W-8721794@

### DIFF
--- a/src/util/fs.ts
+++ b/src/util/fs.ts
@@ -240,8 +240,8 @@ export const fs = Object.assign({}, fsLib, {
    * @param jsonPath The path of the file to write.
    * @param data The JSON object to write.
    */
-  writeJson: async (jsonPath: string, data: AnyJson): Promise<void> => {
-    const fileData: string = JSON.stringify(data, null, 4);
+  writeJson: async (jsonPath: string, data: AnyJson, space: string | number | undefined = 2): Promise<void> => {
+    const fileData: string = JSON.stringify(data, null, space);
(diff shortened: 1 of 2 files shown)


src/util/fs.ts
```diff
@@ -254,8 +254,8 @@ export const fs = Object.assign({}, fsLib, {
    * @param jsonPath The path of the file to write.
    * @param data The JSON object to write.
    */
-  writeJsonSync: (jsonPath: string, data: AnyJson): void => {
-    const fileData: string = JSON.stringify(data, null, 4);
+  writeJsonSync: (jsonPath: string, data: AnyJson, space: string | number | undefined = 2): void => {
```
WillieRuemmele, line 257: would it make sense to add an optional third parameter which could include spaces, encoding, other meta stuff? Just for when/if we need to add something else to this method

WillieRuemmele (approved):

src/util/fs.ts
mshanemc: Not arguing for `fs-extra` here, but  suggesting that we could copy their API.  And maybe not implement all the options, but do spaces with in an options object like they did.  

writeJsonSync( jsonPath: string, data: AnyJson, options: {spaces: })
https://github.com/jprichardson/node-fs-extra/blob/master/docs/writeJson.md

mshanemc (approved):

src/util/fs.ts
mshanemc: in case ever want more options (line endings, a JSONreplacer that strips comments, etc)

PR #370, opened 2021-02-08
Update sfdxProjectJson docs

RodEsp:
GUS WI: [@W-8612474@](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008m6F6IAI/view)

This switches out the old `.retrieve()` method for the new `retrieveSfdxProjectJson()` method in the documentation's example.

### DIFF
--- a/src/sfdxProject.ts
+++ b/src/sfdxProject.ts
@@ -72,7 +72,7 @@ export type ProjectJson = ConfigContents & {
  * be in a top level property that represents your project or plugin.
  *
  * ```
- * const project = await SfdxProjectJson.retrieve();
+ * const project = await retrieveSfdxProjectJson();
  * const myPluginProperties = project.get('myplugin') || {};
  * myPluginProperties.myprop = 'someValue';
  * project.set('myplugin', myPluginProperties);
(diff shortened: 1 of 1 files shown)


RodEsp:
@mdonnalley do you know if adding @forcedotcom/platform-cli notify anyone or should I go with individual users?

src/sfdxProject.ts
```diff
@@ -72,7 +72,7 @@ export type ProjectJson = ConfigContents & {
  * be in a top level property that represents your project or plugin.
  *
  * ```
- * const project = await SfdxProjectJson.retrieve();
+ * const project = await retrieveSfdxProjectJson();
```
peternhale, line 75: @RodEsp Since `retrieveSfdxProjectJson` is an instance method of `SfdxProject`, this example should include how to construct the SfdxProject instance. Also, the docs for `retrieveSfdxProjectJson` state one should use `resolveProjectConfig` instead.

@mdonnalley I grabbed this snippet from `SfdxProject` API docs. Will this suffice?

```
const project = await SfdxProject.resolve();
const projectJson = await project.resolveProjectConfig();
```

peternhale (changes requested):

src/sfdxProject.ts
RodEsp: You're totally right. I accidentally missed those belonged to two separate classes.

peternhale (approved):
I like it.

PR #373, opened 2021-02-11
perf: small queries to identify a devHub

mshanemc:
related to work for @W-8881661@

there's no reason to query every scratchOrgInfo (on a busy hub that someone's authenticating a new machine to, that could be a lot of data.  

The sobject either exists or doesn't so this'll be faster.

### DIFF
--- a/src/org.ts
+++ b/src/org.ts
@@ -233,7 +233,7 @@ export class Org extends AsyncCreatable<Org.Options> {
     const conn = this.getConnection();
     let isDevHub = false;
     try {
-      await conn.query('SELECT Id FROM ScratchOrgInfo');
+      await conn.query('SELECT Id FROM ScratchOrgInfo limit 1');
       isDevHub = true;
     } catch (err) {
       /* Not a dev hub */


src/org.ts
```diff
@@ -233,7 +233,7 @@ export class Org extends AsyncCreatable<Org.Options> {
     const conn = this.getConnection();
     let isDevHub = false;
     try {
-      await conn.query('SELECT Id FROM ScratchOrgInfo');
+      await conn.query('SELECT Id FROM ScratchOrgInfo limit 1');
```
peternhale, line 236: What's cheaper a soql query or sobject describe?

peternhale (approved):

amphro (approved):

src/org.ts
WillieRuemmele: couldn't you use `singleRecordQuery`? If the query returning 2+ results is a bug we should have a test to verify we don't break this in the future

mshanemc: query

mshanemc: we don't care how many come back.  0 records is fine for a virgin devhub...
we only care that the obj exists at all and doesn't throw an error

mshanemc: I'm setting it to `1` so busy hubs don't have to return 50000 or whatever.

shetzel (approved):

PR #374, opened 2021-02-16
fix: pass decrypted client secret for oauth

peternhale:
@W-8894798@ - client secret is being passed as an encrypted value, now passing as decrypted value

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -260,11 +260,9 @@ export class Connection extends JSForceConnection {
 
   /**
    * Getter for the AuthInfo.
-   *
-   * @param decrypt Decrypt the fields
    */
-  public getAuthInfoFields(decrypt?: boolean): AuthFields {
-    return this.options.authInfo.getFields(decrypt);
+  public getAuthInfoFields(): AuthFields {
+    return this.options.authInfo.getFields();
   }
 
   /**
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -260,9 +260,11 @@ export class Connection extends JSForceConnection {
 
   /**
    * Getter for the AuthInfo.
+   *
+   * @param decrypt Decrypt the fields
    */
-  public getAuthInfoFields(): AuthFields {
-    return this.options.authInfo.getFields();
+  public getAuthInfoFields(decrypt?: boolean): AuthFields {
+    return this.options.authInfo.getFields(decrypt);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/connection.ts
```diff
@@ -260,9 +260,11 @@ export class Connection extends JSForceConnection {
 
   /**
    * Getter for the AuthInfo.
+   *
+   * @param decrypt Decrypt the fields
    */
-  public getAuthInfoFields(): AuthFields {
-    return this.options.authInfo.getFields();
+  public getAuthInfoFields(decrypt?: boolean): AuthFields {
```
peternhale, line 266: Add decrypt param function - default of undefined protects backward compatibility.

src/user.ts
```diff
@@ -337,7 +337,7 @@ export class User extends AsyncCreatable<User.Options> {
     } = await this.createUserInternal(fields);
 
     // Create the initial auth info
-    const adminUserAuthFields: AuthFields = this.org.getConnection().getAuthInfoFields();
+    const adminUserAuthFields: AuthFields = this.org.getConnection().getAuthInfoFields(true);
```
peternhale, line 340: decrypted fields needed here to pass along to OAuth token exchange.

mshanemc (approved):

src/connection.ts
amphro: We originally tried to not expose the decrypted values outside of the core library. I know we did that for `getFields` but I don't remember if we looked at other options.

Is there another option here? The answer may be no, but I just want to make sure we looked at other options. For example, maybe there is a way for user to use a protected method to get the authInfo rather than adding decrypt to another public method.

Obviously, anyone using the core library can decrypt if they really want to. The idea was to not make it super easy so consumers don't accidentally log/display/keep in memory when they didn't need to. Adding it to authInfo makes it a little easier. Adding it to the connection makes it really easy. At that point, we might as well just expose the authInfo.

amphro (changes requested):
Should we have a test for this?

src/connection.ts
mshanemc: the other option would be to use AuthInfo directly where it's already exposed.

peternhale:
@amphro I had originally chanced the impl to create the authinfo from org.getUsername() and then fetched the fields from results. I can switch it to that to make things not so easy ;).

amphro (approved):

amphro:
QA notes: Confirmed fix of web:login using a custom connected app. I also verified that user:create still works using jwt, auth url, and web using global connected app.

PR #375, opened 2021-02-16
fix: revert dns polling timeout to 30 and frequency to 10

mdonnalley:
This PR does the following:
- adds env var for retry frequency (`SFDX_DNS_RETRY_FREQUENCY`), which defaults to 1 second
- adds env var for disabling DNS lookup (`SFDX_DISABLE_DNS_CHECK`)

Hopefully addresses https://github.com/forcedotcom/cli/issues/820

@W-8896111@

Testing:
- edit `instanceUrl` in `~/.sfdx/<username>.json` to invalid url
- run `sfdx force:org:display -u <username>`. Command should error and show suggested actions:
```
ERROR running force:org:display:  The org cannot be found

Try this:
Verify that the org still exists
If your org is newly created, wait a minute and run your command again
If you deployed or updated the org's My Domain, logout from the CLI and authenticate again
If you are running in a CI environment with a DNS that blocks external IPs, try setting SFDX_DISABLE_DNS_CHECK=true
```
- run `SFDX_DISABLE_DNS_CHECK=true sfdx force:org:display -u <username>`. Command will error like this:
```
Error: getaddrinfo ENOTFOUND su0503.my.fsalesforce.com
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:67:26) {
  errno: -3008,
  code: 'ENOTFOUND',
  syscall: 'getaddrinfo',
  hostname: 'su0503.my.fsalesforce.com'
}
ERROR:  Error: getaddrinfo ENOTFOUND su0503.my.fsalesforce.com
ERROR running force:org:display:  getaddrinfo ENOTFOUND su0503.my.fsalesforce.com
```

That error is okay to have. We're assuming that people setting `SFDX_DISABLE_DNS_CHECK=true` are in a CI with a dns that blocks external IPs

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -39,7 +39,10 @@ export const SFDX_HTTP_HEADERS = {
 };
 
 export const DNS_ERROR_NAME = 'Domain Not Found';
-const DNS_TIMEOUT = Math.max(3, new Env().getNumber('SFDX_DNS_TIMEOUT') || 3);
+// Timeout for DNS lookup polling defaults to 30 seconds and should always be at least 3 seconds
+const DNS_TIMEOUT = Math.max(3, new Env().getNumber('SFDX_DNS_TIMEOUT', 30) as number);
+// Retry frequency for DNS lookup polling should be at least 10 seconds
+const DNS_RETRY_FREQ = Math.max(10, new Env().getNumber('SFDX_DNS_RETRY_FREQUENCY', 10) as number);
 
 // This interface is so we can add the autoFetchQuery method to both the Connection
 // and Tooling classes and get nice typing info for it within editors.  JSForce is
@@ -220,7 +223,7 @@ export class Connection extends JSForceConnection {
     const resolver = await MyDomainResolver.create({
       url: new URL(this.options.connectionOptions.instanceUrl),
       timeout: Duration.seconds(DNS_TIMEOUT),
-      frequency: Duration.seconds(DNS_TIMEOUT),
+      frequency: Duration.seconds(DNS_RETRY_FREQ),
     });
     try {
       await resolver.resolve();
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -39,10 +39,10 @@ export const SFDX_HTTP_HEADERS = {
 };
 
 export const DNS_ERROR_NAME = 'Domain Not Found';
-// Timeout for DNS lookup polling defaults to 30 seconds and should always be at least 3 seconds
-const DNS_TIMEOUT = Math.max(3, new Env().getNumber('SFDX_DNS_TIMEOUT', 30) as number);
-// Retry frequency for DNS lookup polling should be at least 10 seconds
-const DNS_RETRY_FREQ = Math.max(10, new Env().getNumber('SFDX_DNS_RETRY_FREQUENCY', 10) as number);
+// Timeout for DNS lookup polling defaults to 3 seconds and should always be at least 3 seconds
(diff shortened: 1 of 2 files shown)


src/connection.ts
```diff
@@ -39,7 +39,10 @@ export const SFDX_HTTP_HEADERS = {
 };
 
 export const DNS_ERROR_NAME = 'Domain Not Found';
-const DNS_TIMEOUT = Math.max(3, new Env().getNumber('SFDX_DNS_TIMEOUT') || 3);
+// Timeout for DNS lookup polling should default to 30 seconds
+const DNS_TIMEOUT = new Env().getNumber('SFDX_DNS_TIMEOUT', 30) as number;
+// Retry frequency for DNS lookup polling should be at least 10 seconds
```
mshanemc, line 44: Pete has strong feelings about boundary conditions on this.  I'll let him comment

mshanemc: this PR has context: https://github.com/forcedotcom/sfdx-core/pull/362

mshanemc (approved):

src/connection.ts
mshanemc: and what's taking 30 seconds?  something about the bamboo proxy caching?

mdonnalley: @mshanemc updated

src/connection.ts
```diff
@@ -39,7 +39,10 @@ export const SFDX_HTTP_HEADERS = {
 };
 
 export const DNS_ERROR_NAME = 'Domain Not Found';
-const DNS_TIMEOUT = Math.max(3, new Env().getNumber('SFDX_DNS_TIMEOUT') || 3);
+// Timeout for DNS lookup polling defaults to 30 seconds and should always be at least 3 seconds
+const DNS_TIMEOUT = Math.max(3, new Env().getNumber('SFDX_DNS_TIMEOUT', 30) as number);
+// Retry frequency for DNS lookup polling should be at least 10 seconds
+const DNS_RETRY_FREQ = Math.max(10, new Env().getNumber('SFDX_DNS_RETRY_FREQUENCY', 10) as number);
```
amphro, line 45: Is there a downsize to have the frequency be 3 seconds as opposed to 10 seconds?

mdonnalley: @amphro The reason it's an issue here is that the timeout and retry frequency were _both_ set to 3, which meant that it only tried the lookup once and then failed.

Now that the timeout defaults to 30, I'm not sure if retrying every 3 seconds would be an issue or not. To be safe I just changed it back to what it used to be in versions 2.16.6 and older

mshanemc (approved):

src/connection.ts
```diff
@@ -39,7 +39,10 @@ export const SFDX_HTTP_HEADERS = {
 };
 
 export const DNS_ERROR_NAME = 'Domain Not Found';
-const DNS_TIMEOUT = Math.max(3, new Env().getNumber('SFDX_DNS_TIMEOUT') || 3);
+// Timeout for DNS lookup polling defaults to 3 seconds and should always be at least 3 seconds
+const DNS_TIMEOUT = Math.max(3, new Env().getNumber('SFDX_DNS_TIMEOUT', 3) as number);
```
amphro, line 43: These are generic envvars. What are your thoughts about also putting these in myDomainResolver:98 so all DNS checks follow it? Otherwise, maybe they should be `SFDX_CONNECTION_DNS_XX`?

src/connection.ts
peternhale: @mdonnalley Not sure what affect negative numbers have, but its simple to eliminate those using Math.max. Also give one the ability to enforce a minimum.

peternhale (approved):

PR #376, opened 2021-02-18
fix: meet more password requirements

mshanemc:
@W-8904350@
fixes https://github.com/forcedotcom/cli/issues/858

increases generated password length to 13
ensures that password has one of each of (uppercase, lowercase, number, and special character)

### DIFF
--- a/src/user.ts
+++ b/src/user.ts
@@ -216,13 +216,23 @@ export class User extends AsyncCreatable<User.Options> {
    * Generate default password for a user. Returns An encrypted buffer containing a utf8 encoded password.
    */
   public static generatePasswordUtf8(): SecureBuffer<void> {
-    // Fill an array with random characters from random requirement sets
-    const pass = Array(PASSWORD_LENGTH - ALL.length)
-      .fill(9)
-      .map(() => {
+    // Fill an array with the correct length
+    let pass = Array(PASSWORD_LENGTH).fill('x');
+    const meetsAllRequirements = (candidate: string[]): boolean => {
+      return (
+        candidate.some((char) => NUMBERS.includes(char)) &&
+        candidate.some((char) => SYMBOLS.includes(char)) &&
+        candidate.some((char) => UPPER.includes(char)) &&
+        candidate.some((char) => LOWER.includes(char))
+      );
+    };
+
+    while (!meetsAllRequirements(pass)) {
+      pass = pass.map(() => {
         const set = ALL[rand(ALL)];
         return set[rand(set)];
       });
+    }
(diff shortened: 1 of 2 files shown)


src/user.ts
```diff
@@ -216,13 +216,23 @@ export class User extends AsyncCreatable<User.Options> {
...
+    const meetsAllRequirements = (candidate: string[]): boolean => {
+      return (
+        candidate.some((char) => NUMBERS.includes(char)) &&
+        candidate.some((char) => SYMBOLS.includes(char)) &&
+        candidate.some((char) => UPPER.includes(char)) &&
+        candidate.some((char) => LOWER.includes(char))
+      );
+    };
+
+    while (!meetsAllRequirements(pass)) {
```
amphro, line 230: These type of whiles always make me a little nervous =) even though it seems like the average iterations would be low. Should we have a MAX_ITERATION count that throws an error, just to be safe?

WillieRuemmele: I feel like there has to be some password generation algorithm that can guarantee the presence of certain characters in a deterministic way that gets rid of the `while` loop maybe something like
```
const lower = ['a', 'b'...]
const upper = ['A'...]
const numbers = [1,2..]
const symbols = [...]

const combined = [lower, upper, numbers, symbols]
then shuffle the combined array
then access a random index, and index of that to get a character
and build a password up from that?
```

peternhale: https://www.npmjs.com/package/generate-password seems pretty popular

mshanemc: @WillieRuemmele that's what the original version was doing.  By getting it randomly from that combined list, it would **mostly** work, except sometimes it doesn't catch a number in there, thus the retries.

Probability-wise, you've got 13 chances to get at least one of the 10 numerals out of 77 possible characters, so that's (67/77)^13 (16%), better than the current (67/77)^9 (28%)

But that would only work for orgs that left the pword requirements for orgs who left it as the default alphanumeric.  So adding the symbols checks compounds that a bit, but each of those has a lower probably (because there's more characters out of the 77 that can meet the requirement)

I definitely would have used a 3rd party dependency if that were a thing we did...my thinking was it might be harder given the security implications on this feature.

@amphro I can add a loop limit if it makes you more comfortable--but that's definitely leave scripts as broken as they currently are.  What's the right number of iterations?  The unit test does 1000 in about a half second.  Seem reasonable?

amphro (approved):

WillieRuemmele (approved):

amphro (approved):

PR #378, opened 2021-02-24
fix: issue proper message when get of User fails during web:login

peternhale:
@W-8926345@

Web login flow will fail when the target org does not have REST API enabled. The failure occurs when the User record is fetched from the target org. 

This PR improves the messaging when User record fetch fails.

### DIFF
--- a/src/authInfo.ts
+++ b/src/authInfo.ts
@@ -10,7 +10,8 @@ import * as dns from 'dns';
 import { resolve as pathResolve } from 'path';
 import { basename, extname } from 'path';
 import { parse as urlParse } from 'url';
-import { AsyncCreatable, cloneJson, env, isEmpty, parseJsonMap, set } from '@salesforce/kit';
+import * as os from 'os';
+import { AsyncCreatable, cloneJson, env, isEmpty, parseJson, parseJsonMap, set } from '@salesforce/kit';
 import {
   AnyFunction,
   AnyJson,
@@ -19,6 +20,7 @@ import {
   ensureJsonMap,
   ensureString,
   getString,
+  isArray,
   isPlainObject,
   isString,
   JsonMap,
@@ -1085,10 +1087,42 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
     try {
       this.logger.info(`Sending request for Username after successful auth code exchange to URL: ${url}`);
       const response = await new Transport().httpRequest({ url, headers });
-      return asString(parseJsonMap(response.body).Username);
+      if (response.statusCode >= 400) {
+        this.throwUserGetException(response as never);
+      } else {
+        return asString(parseJsonMap(response.body).Username);
+      }
     } catch (err) {
       throw SfdxError.create('@salesforce/core', 'core', 'AuthCodeUsernameRetrievalError', [orgId, err.message]);
     }
+  }
+
+  /**
+   * Given an error while getting the User object, handle different possibilities of response.body.
+   *
+   * @param response
+   * @private
+   */
+  private throwUserGetException(response: never) {
+    let messages = '';
+    let body: AnyJson | string = getString(
+      response,
+      'body',
+      JSON.stringify({ message: 'UNKNOWN', errorCode: 'UNKNOWN' })
+    );
+    try {
+      body = parseJson(getString(response, 'body', JSON.stringify({ message: 'UNKNOWN', errorCode: 'UNKNOWN' })));
+      if (isArray(body)) {
+        messages = body
+          .map((line) => getString(line, 'message') ?? getString(line, 'errorCode', 'UNKNOWN'))
+          .join(os.EOL);
+      } else {
+        messages = getString(body, 'message') ?? getString(body, 'errorCode', 'UNKNOWN');
+      }
+    } catch (err) {
+      messages = `${body}`;
+    }
+    throw new SfdxError(messages);
   }
 
   // See https://nodejs.org/api/dns.html#dns_dns_lookup_hostname_options_callback
--- a/src/authInfo.ts
+++ b/src/authInfo.ts
@@ -1105,13 +1105,9 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
    */
   private throwUserGetException(response: never) {
     let messages = '';
-    let body: AnyJson | string = getString(
-      response,
-      'body',
-      JSON.stringify({ message: 'UNKNOWN', errorCode: 'UNKNOWN' })
-    );
+    const bodyAsString = getString(response, 'body', JSON.stringify({ message: 'UNKNOWN', errorCode: 'UNKNOWN' }));
     try {
-      body = parseJson(getString(response, 'body', JSON.stringify({ message: 'UNKNOWN', errorCode: 'UNKNOWN' })));
+      const body = parseJson(bodyAsString);
       if (isArray(body)) {
         messages = body
           .map((line) => getString(line, 'message') ?? getString(line, 'errorCode', 'UNKNOWN'))
@@ -1120,7 +1116,7 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
         messages = getString(body, 'message') ?? getString(body, 'errorCode', 'UNKNOWN');
       }
     } catch (err) {
-      messages = `${body}`;
+      messages = `${bodyAsString}`;
     }
     throw new SfdxError(messages);
   }
--- a/src/authInfo.ts
+++ b/src/authInfo.ts
@@ -1088,7 +1088,7 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
       this.logger.info(`Sending request for Username after successful auth code exchange to URL: ${url}`);
       const response = await new Transport().httpRequest({ url, headers });
       if (response.statusCode >= 400) {
-        this.throwUserGetException(response as never);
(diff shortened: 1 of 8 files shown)


src/authInfo.ts
```diff
@@ -1085,12 +1087,44 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
     try {
       this.logger.info(`Sending request for Username after successful auth code exchange to URL: ${url}`);
       const response = await new Transport().httpRequest({ url, headers });
-      return asString(parseJsonMap(response.body).Username);
+      if (response.statusCode >= 400) {
```
peternhale, line 1090: check if request failed and then delegate to handler

src/authInfo.ts
```diff
@@ -1085,12 +1087,44 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
...
 
+  /**
+   * Given an error while getting the User object, handle different possibilities of response.body.
+   *
+   * @param response
+   * @private
+   */
+  private throwUserGetException(response: never) {
+    let messages = '';
+    let body: AnyJson | string = getString(
```
peternhale, line 1108: extract body from response, if not present provide a parsable string that represents an UNKNOWN failure

src/authInfo.ts
```diff
@@ -1085,12 +1087,44 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
...
+      response,
+      'body',
+      JSON.stringify({ message: 'UNKNOWN', errorCode: 'UNKNOWN' })
+    );
+    try {
+      body = parseJson(getString(response, 'body', JSON.stringify({ message: 'UNKNOWN', errorCode: 'UNKNOWN' })));
```
peternhale, line 1114: parse the body and the based on result build messages string.

src/authInfo.ts
```diff
@@ -1085,12 +1087,44 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
...
+      if (isArray(body)) {
+        messages = body
+          .map((line) => getString(line, 'message') ?? getString(line, 'errorCode', 'UNKNOWN'))
+          .join(os.EOL);
+      } else {
+        messages = getString(body, 'message') ?? getString(body, 'errorCode', 'UNKNOWN');
+      }
+    } catch (err) {
+      messages = `${body}`;
```
peternhale, line 1123: If parse fails, just use string representation of body. Server errors >= 500 may return an html block instead of JSON, so the message will be ugly, but will at least provide the needed info to user.

src/authInfo.ts
```diff
@@ -1085,12 +1087,44 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
...
+    }
+    throw new SfdxError(messages);
```
peternhale, line 1125: Throw the resulting message

test/unit/authInfoTest.ts
```diff
@@ -2008,4 +2008,90 @@ describe('AuthInfo', () => {
...
+        body: JSON.stringify([
+          {
+            message: 'The REST API is not enabled for this Organization',
+            errorCode: 'RESTAPINOTENABLED',
+          },
+        ]),
+      };
+      stubMethod($$.SANDBOX, Transport.prototype, 'httpRequest').returns(Promise.resolve(responseBody));
+      try {
+        await AuthInfo.create({ oauth2Options: authCodeConfig });
```
peternhale, line 2040: I tried use `expect(await AuthInfo.create({ oauth2Options: authCodeConfig })).to.throw....` but I could not get it to work. Suggestions welcome.

src/authInfo.ts
amphro: NIT: I would prefer to have two different variables rather than sharing a type. Maybe `const possibleJsonBody` here and `const body` on line 1114?

src/authInfo.ts
amphro: Ya, those html pages being returned from the server are annoying. We can have follow-ups work items to clean those up if we start getting a bunch of them.

test/unit/authInfoTest.ts
amphro: Maybe `expect(AuthInfo.create({ oauth2Options: authCodeConfig })).to.reject`?

amphro: I think this is fine though, but please add an `assert(false, 'should throw')` below this line, otherwise, the test will pass if it doesn't throw.

test/unit/authInfoTest.ts
```diff
@@ -2008,4 +2008,90 @@ describe('AuthInfo', () => {
...
+      const responseBody = {
+        statusCode: 403,
+        body: JSON.stringify({
+          message: 'The REST API is not enabled for this Organization',
+          errorCode: 'RESTAPINOTENABLED',
+        }),
+      };
+      stubMethod($$.SANDBOX, Transport.prototype, 'httpRequest').returns(Promise.resolve(responseBody));
+      try {
+        await AuthInfo.create({ oauth2Options: authCodeConfig });
```
amphro, line 2055: Ditto on all of these, you need to assert failure if you can't get the `.reject` to work.

amphro (approved):

test/unit/authInfoTest.ts
```diff
@@ -2008,4 +2008,90 @@ describe('AuthInfo', () => {
...
+            message: 'The REST API is not enabled for this Organization',
+            errorCode: 'RESTAPINOTENABLED',
+          },
+        ]),
+      };
+      stubMethod($$.SANDBOX, Transport.prototype, 'httpRequest').returns(Promise.resolve(responseBody));
+      try {
+        await AuthInfo.create({ oauth2Options: authCodeConfig });
+      } catch (err) {
+        expect(err).to.have.property('message').to.include('The REST API is not enabled for this Organization');
```
jshackell-sfdc, line 2042: Edit: "The REST API isn't enabled in this org."

test/unit/authInfoTest.ts
```diff
@@ -2008,4 +2008,90 @@ describe('AuthInfo', () => {
...
+      }
+    });
+    it('user get returns 403 with body of json map', async () => {
+      const responseBody = {
+        statusCode: 403,
+        body: JSON.stringify({
+          message: 'The REST API is not enabled for this Organization',
```
jshackell-sfdc, line 2049: Ditto (Edit: "The REST API isn't enabled in this org.")

test/unit/authInfoTest.ts
```diff
@@ -2008,4 +2008,90 @@ describe('AuthInfo', () => {
...
+          errorCode: 'RESTAPINOTENABLED',
+        }),
+      };
+      stubMethod($$.SANDBOX, Transport.prototype, 'httpRequest').returns(Promise.resolve(responseBody));
+      try {
+        await AuthInfo.create({ oauth2Options: authCodeConfig });
+      } catch (err) {
+        expect(err).to.have.property('message').to.include('The REST API is not enabled for this Organization');
```
jshackell-sfdc, line 2057: Ditto (Edit: "The REST API isn't enabled in this org.")

test/unit/authInfoTest.ts
```diff
@@ -2008,4 +2008,90 @@ describe('AuthInfo', () => {
...
+      }
+    });
+    it('user get returns 403 with string body', async () => {
+      const responseBody = {
+        statusCode: 403,
+        body: 'The REST API is not enabled for this Organization',
```
jshackell-sfdc, line 2063: Ditto (Edit: "The REST API isn't enabled in this org.")

test/unit/authInfoTest.ts
```diff
@@ -2008,4 +2008,90 @@ describe('AuthInfo', () => {
...
+      };
+      stubMethod($$.SANDBOX, Transport.prototype, 'httpRequest').returns(Promise.resolve(responseBody));
+      try {
+        await AuthInfo.create({ oauth2Options: authCodeConfig });
+      } catch (err) {
+        expect(err).to.have.property('message').to.include('The REST API is not enabled for this Organization');
```
jshackell-sfdc, line 2072: Ditto (Edit: "The REST API isn't enabled in this org.")

test/unit/authInfoTest.ts
jshackell-sfdc: Ah, I see.  Hmm... Well, it would be great if the CLI could output a better error; that one from the app server is old-fashioned, and maybe even "wrong" in that it uses "organization" instead of "org".   But up to you!

peternhale: @jshackell-sfdc, if there were only one possible error from getting a user record, then I would be ok with mapping the error to a better message, but there are probably more than one to deal with and I am sure most can use a bit of polishing.

src/authInfo.ts
```diff
@@ -1085,12 +1087,40 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
...
     }
   }
 
+  /**
+   * Given an error while getting the User object, handle different possibilities of response.body.
+   *
+   * @param response
+   * @private
+   */
+  private throwUserGetException(response: never) {
```
shetzel, line 1106: I think `Unknown` is a better type here.  `never` is a function return type.

shetzel (changes requested):
Please change `never` --> `Unknown` and add test assertion suggestions.

shetzel (approved):

PR #379, opened 2021-02-24
fix: await the call to authInfo.save

shetzel:
When loading the instance API version we need to await the call to authInfo save.
@W-8927294@


peternhale (approved):

amphro (approved):

amphro:
Ugg..... that sucks. My bad.

PR #380, opened 2021-02-26
Do not add listening event handler if oauthServer is already listening

hansalves:
This fixes a bug that can cause the process to hang when trying to authenticate an org.


mdonnalley:
@hansalves Thank you so much for your PR. I've cherry-picked your changes into a different PR (#384) so that we can merge this fix faster.

PR #383, opened 2021-03-03
fix: resolve issue with WebOauthServer hanging

mdonnalley:
@W-8956885@

PR #384, opened 2021-03-03
Do not add listening event handler if oauthServer is already listening

mdonnalley:
cherry-picking https://github.com/forcedotcom/sfdx-core/pull/380 so that we can merge this faster. 

Thanks to @hansalves for fixing this for us.

@W-8956885@
Closes #380 #383


RodEsp (approved):
👍🏼

WillieRuemmele (approved):

PR #387, opened 2021-03-12
fix(auth): Allow for parsing sfdxAuthUrls from json files

RodEsp:
Allows for getting sfdxAuthUrls from json files like stated in the [documentation for `auth:sfdxurl:store`](https://developer.salesforce.com/docs/atlas.en-us.sfdx_cli_reference.meta/sfdx_cli_reference/cli_reference_auth_sfdxurl.htm#cli_reference_auth_sfdxurl_store).

@W-8993725@

### DIFF
--- a/src/authInfo.ts
+++ b/src/authInfo.ts
@@ -515,9 +515,12 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
    * @param sfdxAuthUrl
    */
   public static parseSfdxAuthUrl(sfdxAuthUrl: string) {
-    const match = sfdxAuthUrl.match(
-      /^force:\/\/([a-zA-Z0-9._-]+):([a-zA-Z0-9._-]*):([a-zA-Z0-9._-]+)@([a-zA-Z0-9._-]+)/
-    );
+    let url;
+    try {
+      url = JSON.parse(sfdxAuthUrl).sfdxAuthUrl;
+    } catch (error) {
+      url = sfdxAuthUrl;
+    }
 
     if (!match) {
       throw new SfdxError(
--- a/src/authInfo.ts
+++ b/src/authInfo.ts
@@ -521,7 +521,9 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
     } catch (error) {
       url = sfdxAuthUrl;
     }
-
+    
+    const match = url.match(/^force:\/\/([a-zA-Z0-9._-]+):([a-zA-Z0-9._-]*):([a-zA-Z0-9._-]+)@([a-zA-Z0-9._-]+)/);
+    
     if (!match) {
       throw new SfdxError(
         'Invalid sfdx auth url. Must be in the format `force://<clientId>:<clientSecret>:<refreshToken>@<loginUrl>`. The instanceUrl must not have the protocol set.',
--- a/src/authInfo.ts
+++ b/src/authInfo.ts
@@ -521,9 +521,9 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
     } catch (error) {
       url = sfdxAuthUrl;
     }
-    
+
     const match = url.match(/^force:\/\/([a-zA-Z0-9._-]+):([a-zA-Z0-9._-]*):([a-zA-Z0-9._-]+)@([a-zA-Z0-9._-]+)/);
-    
+
     if (!match) {
       throw new SfdxError(
         'Invalid sfdx auth url. Must be in the format `force://<clientId>:<clientSecret>:<refreshToken>@<loginUrl>`. The instanceUrl must not have the protocol set.',
--- a/test/unit/authInfoTest.ts
+++ b/test/unit/authInfoTest.ts
@@ -2014,6 +2014,19 @@ describe('AuthInfo', () => {
       expect(options.loginUrl).to.equal('https://test.my.salesforce.com');
     });
 
+    it('should parse the correct url when passed in as json', () => {
+      const options = AuthInfo.parseSfdxAuthUrl(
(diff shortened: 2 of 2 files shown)


peternhale (approved):

peternhale:
@RodEsp do existing tests adequately cover this change?

RodEsp:
@peternhale yeah, there are tests for everything else related to this function, just not one for passing in a json string which I added.

I'm trying to add one to https://github.com/salesforcecli/plugin-auth as well but having trouble running its unit tests locally.

test/unit/authInfoTest.ts
```diff
@@ -2014,6 +2014,19 @@ describe('AuthInfo', () => {
       expect(options.loginUrl).to.equal('https://test.my.salesforce.com');
     });
 
+    it('should parse the correct url when passed in as json', () => {
```
WillieRuemmele, line 2017: do we have a test for the case it's a just a string?

src/authInfo.ts
```diff
@@ -515,9 +515,14 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
    * @param sfdxAuthUrl
    */
   public static parseSfdxAuthUrl(sfdxAuthUrl: string) {
-    const match = sfdxAuthUrl.match(
-      /^force:\/\/([a-zA-Z0-9._-]+):([a-zA-Z0-9._-]*):([a-zA-Z0-9._-]+)@([a-zA-Z0-9._-]+)/
-    );
+    let url;
+    try {
+      url = JSON.parse(sfdxAuthUrl).sfdxAuthUrl;
```
mdonnalley, line 520: This should be fixed in plugin-auth: https://github.com/salesforcecli/plugin-auth/blob/main/src/commands/auth/sfdxurl/store.ts#L55

Something like this:

```typescript
const sfdxAuthUrl = this.flags.sfdxurlfile.endsWith('.json')
  ? (await fs.readJson(this.flags.sfdxurlfile)).sfdxAuthUrl
  : await fs.readFile(this.flags.sfdxurlfile, 'utf8');
```

WillieRuemmele (approved):

mdonnalley (changes requested):

test/unit/authInfoTest.ts
RodEsp: Yes, every other test for this function uses just a string.
For example, https://github.com/forcedotcom/sfdx-core/pull/387/files#diff-72e39279bdd6c1fae17308917208a797a82022b0150c99572b109abcffe4449fR1990

src/authInfo.ts
RodEsp: @mdonnalley I did try that but it got pretty messy because of types.
You have to end up doing something like this
```
// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call
const sfdxAuthUrl = this.flags.sfdxurlfile.endsWith('.json')
  ? (await fs.readJson(this.flags.sfdxurlfile)).sfdxAuthUrl
  : await fs.readFile(this.flags.sfdxurlfile, 'utf8');
```

And even then `tsc` doesn't like it so it won't compile
image


I just gave it another shot though and this is what I've ended up with.
```
    let sfdxAuthUrl;
    // eslint-disable-next-line @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-member-access
    if (this.flags.sfdxurlfile.endsWith('json')) {
      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
      sfdxAuthUrl = await fs.readJson(this.flags.sfdxurlfile);
      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access
      sfdxAuthUrl = sfdxAuthUrl.sfdxAuthUrl;
    } else {
      sfdxAuthUrl = await fs.readFile(this.flags.sfdxurlfile, 'utf8');
    }
```    

I would rather avoid something that ugly, do you have any other suggestions on how to do this without having  to resort to that mess?

RodEsp: Here's a PR with those changes for easier viewing
https://github.com/salesforcecli/plugin-auth/pull/81

mdonnalley: @RodEsp 

```typescript
 import { AnyJson } from '@salesforce/ts-types';

 type AuthUrlJson = AnyJson & {
  sfdxUrlJson: string;
 };

 public async run(): Promise<AuthFields> {
    const authFile = this.flags.sfdxurlfile as string;
    const sfdxAuthUrl = authFile.endsWith('.json')
      ? await this.getUrlFromJson(authFile)
      : await this.getUrlFromFile(authFile);
    // etc...  
  }


  private async getUrlFromJson(authFile: string): Promise<string> {
    const contents = (await fs.readJson(authFile)) as AuthUrlJson;
    return contents.sfdxUrlJson;
  }

  private async getUrlFromFile(authFile: string): Promise<string> {
    const contents = await fs.readFile(authFile, 'utf8');
    return contents;
  }
```

RodEsp: Oh man, do we really want to create a whole 'nother type wiht private methods just for reading a simple JSON file? `sfdx-cli` is my first real foray into TS but is this just "the typescript way"?
This is what I meant [here](https://platformcloud.slack.com/archives/G4Q6L9KDH/p1615575817258600?thread_ts=1615575689.258400&cid=G4Q6L9KDH).

Thoughts? @WillieRuemmele @amphro @peternhale @mshanemc

peternhale: Since this is so simple I would use `Dictionary` instead of `AnyJson`

so 
```
const contents = (await fs.readJson(authFile)) as Disctiony;
return contents.sfdxUrlJson;
```

shetzel: I like Mike's approach, except I don't think the `getUrlFromFile` method is necessary;  Just do the `await readFile`.  And the AuthUrlJson type can just be `{ sfdxUrlJson: string };`

shetzel (approved):

RodEsp:
Ok, closing this in favor of https://github.com/salesforcecli/plugin-auth/pull/81

PR #389, opened 2021-03-25
deployRecentValidation

WillieRuemmele:
@W-9014822@ 

adds the `deployRecentValidation` method to do a quick deploy with REST or SOAP

compare with https://github.com/salesforcecli/toolbelt/blob/main/src/lib/mdapi/mdApiUtil.ts#L36
and https://github.com/salesforcecli/toolbelt/blob/main/src/lib/mdapi/mdApiUtil.ts#L51

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -214,15 +214,12 @@ export class Connection extends JSForceConnection {
   /**
    * Will deploy a recently validated deploy request
    *
-   * @param options.id = the id of the deploy to quickly validate
+   * @param options.id = the deploy ID that's been validated already from a previous checkOnly deploy request
    * @param options.rest = a boolean whether or not to use the REST API
    */
   public async deployRecentValidation(options: recentValidationOptions): Promise<JsonCollection> {
     if (options.rest) {
-      const url = `${this.instanceUrl.replace(
-        /\/$/,
-        ''
-      )}/services/data/v${this.getApiVersion()}/metadata/deployRequest`;
+      const url = `${this.baseUrl()}/metadata/deployRequest`;
       const messageBody = JSON.stringify({
         validatedDeployRequestId: options.id,
       });
@@ -234,6 +231,7 @@ export class Connection extends JSForceConnection {
       const requestOptions = { headers: 'json' };
       return this.request(requestInfo, requestOptions);
     } else {
+      // the _invoke is private in jsforce, we can call the SOAP deployRecentValidation like this
       // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
       // @ts-ignore
       return this.metadata['_invoke']('deployRecentValidation', {
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -221,7 +221,9 @@ export class Connection extends JSForceConnection {
    * @param options.rest = a boolean whether or not to use the REST API
    */
   public async deployRecentValidation(options: recentValidationOptions): Promise<JsonCollection> {
-    if (options.rest) {
+    const rest = options.rest;
+    delete options.rest;
+    if (rest) {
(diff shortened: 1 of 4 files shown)


src/connection.ts
```diff
@@ -210,6 +211,36 @@ export class Connection extends JSForceConnection {
     return super._baseUrl();
   }
 
+  /**
+   * Will deploy a recently validated deploy request
+   *
+   * @param options.id = the id of the deploy to quickly validate
```
shetzel, line 217: This param description isn't quite accurate.  It's the deploy ID that's been validated already from a previous checkOnly deploy request.   See this doc for lots of details: https://developer.salesforce.com/docs/atlas.en-us.api_meta.meta/api_meta/meta_deployRecentValidation.htm

src/connection.ts
```diff
@@ -210,6 +211,36 @@ export class Connection extends JSForceConnection {
...
+      const messageBody = JSON.stringify({
+        validatedDeployRequestId: options.id,
+      });
+      const requestInfo = {
+        method: 'POST',
+        url,
+        body: messageBody,
+      };
+      const requestOptions = { headers: 'json' };
+      return this.request(requestInfo, requestOptions);
```
shetzel, line 235: This doesn't auto-refresh the auth.  Not sure it's a huge deal since people will generally run this immediately after a successful checkOnly deploy, but it would be a bug.  That could be addressed in another work item though.

test/unit/connectionTest.ts
```diff
@@ -273,6 +273,17 @@ describe('Connection', () => {
     }
   });
 
+  describe('deployRecentValidation', () => {
```
shetzel, line 276: I think we need 2 tests.  1 for REST.  1 for SOAP.

shetzel (changes requested):

src/connection.ts
```diff
@@ -210,6 +211,36 @@ export class Connection extends JSForceConnection {
...
+   *
+   * @param options.id = the id of the deploy to quickly validate
+   * @param options.rest = a boolean whether or not to use the REST API
+   */
+  public async deployRecentValidation(options: recentValidationOptions): Promise<JsonCollection> {
+    if (options.rest) {
+      const url = `${this.instanceUrl.replace(
+        /\/$/,
+        ''
+      )}/services/data/v${this.getApiVersion()}/metadata/deployRequest`;
```
peternhale, line 225:
```suggestion
      const url = `${this.baseUrl()}/metadata/deployRequest`;
```

peternhale (changes requested):

shetzel (approved):

src/connection.ts
```diff
@@ -210,6 +211,34 @@ export class Connection extends JSForceConnection {
     return super._baseUrl();
   }
 
+  /**
+   * Will deploy a recently validated deploy request
+   *
+   * @param options.id = the deploy ID that's been validated already from a previous checkOnly deploy request
+   * @param options.rest = a boolean whether or not to use the REST API
+   */
+  public async deployRecentValidation(options: recentValidationOptions): Promise<JsonCollection> {
```
RodEsp, line 220: I don't really know anything about this code base but it seems strange to me that this method would be inside `connection.ts` and the `Connection` class. Is this something all connections will need/use?

amphro: It belongs in JSForce but we aren't going to put it there until we move over to v2. These will be removed in the v2 of core that uses v2 of jsforce. @WillieRuemmele can you put comments on these methods you are adding saying that, and that it is just to add the rest option for existing soap operations in jsforce.

amphro (changes requested):

amphro (approved):

PR #390, opened 2021-03-29
Wr/deploy

WillieRuemmele:
@W-8943593@

see https://github.com/salesforcecli/toolbelt/blob/main/src/lib/core/force.ts#L727 and https://github.com/salesforcecli/toolbelt/blob/main/src/lib/core/force.ts#L712

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -4,7 +4,10 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
+import { Stream } from 'stream';
 import { URL } from 'url';
+import { AsyncResult, DeployOptions, DeployResultLocator } from 'jsforce/api/metadata';
+import { Callback } from 'jsforce/connection';
 import { Duration, maxBy, merge, env } from '@salesforce/kit';
 import {
   asString,
@@ -25,6 +28,7 @@ import {
   QueryResult,
   RequestInfo,
   Tooling as JSForceTooling,
+  // Metadata as JSForceMetadata,
 } from 'jsforce';
 import { AuthFields, AuthInfo } from './authInfo';
 import { MyDomainResolver } from './status/myDomainResolver';
@@ -33,6 +37,8 @@ import { ConfigAggregator } from './config/configAggregator';
 import { Logger } from './logger';
 import { SfdxError } from './sfdxError';
 import { sfdc } from './util/sfdc';
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const FormData = require('form-data');
 
 /**
  * The 'async' in our request override replaces the jsforce promise with the node promise, then returns it back to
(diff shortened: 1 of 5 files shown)


peternhale (approved):

src/keyChainImpl.ts
```diff
@@ -165,7 +165,7 @@ export class KeychainAccess implements PasswordStore {
     }
 
     // eslint-disable-next-line @typescript-eslint/no-misused-promises
-    credManager.on('close', async (code) => {
+    credManager.on('close', async (code: number) => {
```
WillieRuemmele, line 168: this was causing a build error

src/connection.ts
```diff
@@ -4,7 +4,11 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
+import * as fs from 'fs';
```
amphro, line 7: Will the `'../fs'` not work?

src/connection.ts
```diff
@@ -160,6 +169,13 @@ export class Connection extends JSForceConnection {
     return conn;
   }
 
+  private static createReadStreamFromBuffer(buffer: Buffer): fs.ReadStream {
```
amphro, line 172: Let's not do this...

Why not just do something like:

```
import { Readable } from 'stream';
...
Readable.from(buffer);
```

src/connection.ts
```diff
@@ -168,6 +184,54 @@ export class Connection extends JSForceConnection {
...
+    delete options.rest;
+    if (rest) {
+      this.logger.debug('deploy with REST');
+      const headers = {
+        Authorization: this && `OAuth ${this.accessToken}`,
+        clientId: this.oauth2 && this.oauth2.clientId,
+        'Sforce-Call-Options': 'client=sfdx-core',
+      };
+      const url = `${this.baseUrl()}/metadata/deployRequest`;
+      const request = Transport.prototype._getHttpRequestModule();
```
amphro, line 203: Ugg, we should keep this in mind with jsforce v2. I know we want this in jsforce but it would be nice to make this better for people trying to extend jsforce.

amphro (changes requested):

src/connection.ts
WillieRuemmele: oops, just chose the wrong import

src/connection.ts
WillieRuemmele: I tried transforming the `buffer` into a `ReadableStream` just about every way that I could find without any working. The api seems to only accept types of `fs.ReadStream`. I'm very open to pairing and showing you what I tried and the results they produce

src/connection.ts
WillieRuemmele: yeah, this and the above `Buffer` to `fs.ReadStream` are the two big things I don't like about how we're doing it. I wasn't sure what the best way to do the request was here

src/connection.ts
```diff
@@ -168,6 +184,54 @@ export class Connection extends JSForceConnection {
...
+
+      return await new Promise((resolve, reject) => {
```
shetzel, line 205: Outside of try/catch blocks, `return await` is redundant.   You can just `return new Promise(...)`

shetzel (approved):

src/connection.ts
```diff
@@ -49,6 +54,7 @@ export const SFDX_HTTP_HEADERS = {
 };
 
 export const DNS_ERROR_NAME = 'Domain Not Found';
+type deployOptions = DeployOptions & { rest?: boolean };
```
amphro, line 57: You should always export types used in public methods, maybe `DeployOptionsWithRest`?

amphro (approved):

PR #394, opened 2021-04-02
fix: debugEnabled should be public

shetzel:
Core's logger uses the Debug library. That library has a public field so that you can check if debug is enabled and if so, do some specific logic for debugging that could potentially degrade performance. To avoid this when debug is not enabled we need to expose this property on the Logger and have @salesforce/command library use it so that it's available for plugins.

@W-9092712@

### DIFF
--- a/test/unit/loggerTest.ts
+++ b/test/unit/loggerTest.ts
@@ -40,7 +40,7 @@ describe('Logger', () => {
       const logger1 = new Logger('testLogger');
       expect(logger1).to.be.instanceof(Logger);
       expect(logger1.getName()).to.equal('testLogger');
-      const logger2 = Logger.root();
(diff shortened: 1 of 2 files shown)


test/unit/loggerTest.ts
```diff
@@ -40,7 +40,7 @@ describe('Logger', () => {
       const logger1 = new Logger('testLogger');
       expect(logger1).to.be.instanceof(Logger);
       expect(logger1.getName()).to.equal('testLogger');
-      const logger2 = Logger.root();
+      const logger2 = await Logger.root();
```
shetzel, line 43: Pretty sure this was meant to await the result, otherwise it's just a Promise.

WillieRuemmele (approved):

PR #397, opened 2021-04-28
feat: add GlobalInfo config file

mdonnalley:
- Adds `GlobalInfo` class for managing all authorization-like data in a single file
- Adds ability for `GlobalInfo` class to read/write to both the old and new data format and location

#### TODO
- [x] Fix unit tests
- [x] Add unit tests
- [x] Fully remove `AuthInfoConfig`
- [x] Update examples directory
- [x] Fix test stubbing

@W-9103872@


src/config/sfdxDataHandler.ts
```diff
@@ -0,0 +1,195 @@
...
+    const auths = await this.listAllAuthorizations();
+    const environments = auths.reduce(
+      (x, y) => Object.assign(x, { [ensureString(y.username)]: y }),
+      {} as Authorizations
+    );
+    return { [this.sfKey]: environments };
+  }
+
+  public async write(latest: SfData, original: SfData): Promise<void> {
+    const { changed, deleted } = await this.findChanges(latest, original);
```
amphro, line 123: Nice

src/config/globalInfoConfig.ts
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  public async write(newContents?: SfData): Promise<SfData> {
+    const result = (await super.write(newContents)) as SfData;
+    if (this.enableInteroperability) await this.sfdxHandler.write(result);
+    return result;
+  }
+
+  protected async init(): Promise<void> {
+    const contents = this.enableInteroperability ? await this.mergeWithSfdxData() : await this.loadSfData();
```
amphro, line 91: Ahh I see. This would be awesome if just turning this off made all the unit tests pass =) but I see how the mocks will probably cause problems.

 We should also profile this init with interoperability to make sure it is fine. I don't think there is a performance hit but good to make sure.

examples/examples/apexLogStream/apexLogStream.ts
```diff
@@ -30,9 +30,8 @@ export async function run() {
     });
   }
 
-  // List all auth files
-  const authFiles = await AuthInfo.listAllAuthFiles();
-  const orgs = authFiles.map(authfile => authfile.replace('.json', ''));
+  const config = await GlobalInfo.getInstance();
+  const orgs = Object.keys(config.authorizations)
```
amphro, line 34: I think we need a `type` on authorizations. For example, what about the functions account? So this would be something like `config.authorizations.filter(auth => auth.type === 'org')` or something like that. OR maybe something like the following:

```
// eventually you could have 'heroku', 'globalId' etc.
type AuthorizationType = 'org' | 'functions'; 
...
// If nothing is passed in, return all of them
config.getAuthorizations(type?: AuthorizationType)
```

OR. Maybe this is through environment?
```
// eventually you could have 'heroku', 'globalId' etc. Maybe enum instead.
type AuthorizationType = 'org' | 'functions';
type EnvironmentType = 'org' | 'compute';
// If there isn't a token on the env, or they need to refresh the token, it would have to go to the authorization.
...
// If nothing is passed in, return all of them
config.getEnvironments(type?: EnvironmentType)
```

Just some thoughts.

src/authInfo.ts
```diff
@@ -77,7 +75,7 @@ export interface AuthFields {
   usernames?: string[];
   userProfileName?: string;
   expirationDate?: string;
-}
+};
 
 export type Authorization = {
```
amphro, line 80: I wonder if we will need other kinds of Authorization.

```
export type Authorization = {
   loginUrl: string;
   token: string;
   timestamp: string;
}

export type OrgAuthorization = Authorization & {
   orgId: string // should this be on environment rather than authorization?
   authMethod: ...
}

export FunctionAuthorization = Authorization & {
 ....
}
```

Do we need all types of auth specified here or do plugins have this info and it is just stored through core because of the json map?

PR #398, opened 2021-05-03
feat: add message support for markdown files @W-9210781@

amphro:
Both `.js` and `.json` message files have issues.

* `.json` doesn't support comments for variables.
* `.js` is much more dev-focused rather than doc-writer-focused, which is one of the big benefits of message files.

Both don't handle multiline very well: `.json` has really long lines with \n and `.js` uses template strings which break formatting.

This PR adds markdown support to give another option for message files. It doesn't support variables like js, but multiline support is much better, and we can do some awesome things with markdown formatting and how that translates to the terminal. For example, an array of messages for a single key could be a list.

json: (new lines in string)
```json
{
  "myKey": [
    "action 1",
    "action 2\nmore action 2"
  ]
}
```

js (broken formatting)
```javascript
module.exports = {
  "myKey": [
    "action 1",
    `action 2
more action 2`
  ]
}
```

markdown (beautiful)
```markdown
# myKey
* action 1
* action 2
  more action 2
```


src/messages.ts
```diff
@@ -38,6 +38,77 @@ class Key {
...
+  for (const section of sections) {
+    const lines = section.split('\n');
+    const firstLine = lines.shift();
+    const rest = lines.join('\n').trim();
+
+    if (firstLine && rest.length > 0) {
+      const key = firstLine.trim();
+      const nonEmptyLines = lines.filter((line) => !!line.trim());
+      // If every entry in the value is a list item, then treat this as a list. Indented lines are part of the list.
+      if (nonEmptyLines.every((line) => /^[*-]\s+|^ {2}/.exec(line))) {
```
mdonnalley, line 62: I've found it helpful when dealing with a lot of regex patterns to abstract them out to good variable names, e.g.

```typescript
const isListItem = /^[*-]\s+|^ {2}/
nonEmptyLines.every((line) => isListItem.exec(line))
```

Just for the sanity of the next developer

messages/auth.md
```diff
@@ -0,0 +1,37 @@
...
+
+No %s org found
+
+# defaultOrgNotFoundActions
+
+- Run the "sfdx auth" commands with --setdefaultusername to connect to an org and set it as your default org.
+- Run "force:org:create" with --setdefaultusername to create a scratch org and set it as your default org.
+- Run "sfdx config:set defaultusername=<username>" to set your default username.
+
+# PortInUse
```
jshackell-sfdc, line 11: tiny nit: for consistency I assume we should recommend camelCase?  i.e portInUse

messages/auth.md
```diff
@@ -0,0 +1,37 @@
...
+
+Cannot start the OAuth redirect server on port %s.
+
+# PortInUseAction
```
jshackell-sfdc, line 15: Ditto

src/messages.ts
```diff
@@ -38,6 +38,77 @@ class Key {
...
+            .map((line) => line.trim())
+            // put it back together
+            .join('\n')
+        );
+        map.set(key, values);
+      } else {
+        map.set(key, rest);
+      }
+    } else {
+      // messages.js should have no internal dependencies.
```
mdonnalley, line 79: I'm not sure this comment applies to this error

src/messages.ts
```diff
@@ -38,6 +38,77 @@ class Key {
  */
 export type LoaderFunction = (locale: string) => Messages;
 
+/**
+ * Different message file content parsers. This allows us to have js, json, and md. Maybe we will eventually support yaml, etc.
+ */
+type FileParser = (filePath: string, fileContents: string) => Map<string, AnyJson>;
+
+const markdownLoader: FileParser = (filePath: string, fileContents: string): Map<string, AnyJson> => {
+  const map = new Map<string, AnyJson>();
```
mdonnalley, line 47: `Map<string, string>`?

mdonnalley (approved):
Just a few suggestions

messages/auth.md
```diff
@@ -0,0 +1,37 @@
+# defaultOrgNotFound
+
+No %s org found
+
+# defaultOrgNotFoundActions
+
+- Run the "sfdx auth" commands with --setdefaultusername to connect to an org and set it as your default org.
+- Run "force:org:create" with --setdefaultusername to create a scratch org and set it as your default org.
+- Run "sfdx config:set defaultusername=<username>" to set your default username.
+
```
jshackell-sfdc, line 10: Question: will this message be displayed with actual bullets?  Or will the message contain the literal character "-" at the beginning of each line?  Just curious. We talked about some Markdown elements *not* rendering in the terminal output, it would be good for me to list either they stuff that does (or doesn't, depending on how long the list is) render using Markdown format in the terminal.

amphro: This would just make an array of messages, for whatever the consumer wants to do with it. So the existing errors with actions would still look exactly the same. 

Since it is an array, a consumer can do whatever they want with it, meaning vscode could display it in whatever way that makes sense for them, etc.

messages/auth.md
amphro: Ya, I can update these. Assuming everyone is ok with markdown, I can go update the rest of the messages too.

src/messages.ts
amphro: Well, you can have nested JsonMaps and JsonArrays. For example, 

```
{
  "myKey": {
    "mySubKey: ["item1", "item2"]
  }
}
```

I could do something like the following but it didn't really seem worth it at the time.

```
type MessageContainer = Map<string, string | string[] | MessageContainer>
```

I can make the change if you like that better.

src/messages.ts
amphro: It does, because you can't have SfdxError. I will make it more clear.

messages/auth.md
```diff
@@ -0,0 +1,37 @@
+# defaultOrgNotFound
+
```
jshackell-sfdc, line 2: If pressed, I'd vote for using "error.<key>" and "error.<key>.actions", just to be completely clear what these things are.  But honestly, I don't really care.  It's  pretty obvious that "defaultOrgNotFound" is an error.  So I leave that up to you.

src/messages.ts
mdonnalley: Nah it's fine as is

jshackell-sfdc (approved):

shetzel (approved):

shetzel (approved):

RodEsp:
Tested using https://github.com/salesforcecli/plugin-env/tree/re/test-markdown-messages and `yarn link @salesforce/core`.


Verified both that the `.md` files are read correctly and that the locale error for `en_US` shows up if the keys aren't found on build.

PR #403, opened 2021-05-10
chore: add loginurl as option when using access token flow

peternhale:
@W-8066452@ add auth:toke:set command

### DIFF
--- a/src/authInfo.ts
+++ b/src/authInfo.ts
@@ -628,13 +628,13 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
   public getConnectionOptions(): ConnectionOptions {
     let opts: ConnectionOptions;
 
-    const { accessToken, instanceUrl } = this.fields;
+    const { accessToken, instanceUrl, loginUrl } = this.fields;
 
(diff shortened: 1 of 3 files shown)


src/authInfo.ts
```diff
@@ -628,13 +628,13 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
...
 
-    const { accessToken, instanceUrl } = this.fields;
+    const { accessToken, instanceUrl, loginUrl } = this.fields;
 
     if (this.isAccessTokenFlow()) {
       this.logger.info('Returning fields for a connection using access token.');
 
       // Just auth with the accessToken
-      opts = { accessToken, instanceUrl };
+      opts = { accessToken, instanceUrl, loginUrl };
```
peternhale, line 637: @amphro when trying w/o setting loginUrl, I end up with a connection instance where loginUrl is pointing at login.salesforce.com, which fails when trying to hit and endpoint that is on a sandbox. Choice here was just to set the loginUrl to instance url. This satisfies the needs of the new command and does not seem to affect use of access token as username.

maggiben (approved):

PR #405, opened 2021-05-21
feat: modify authInfo to return username for access token auth

peternhale:
@W-8066452@

When creating an authInfo with `AuthInfo.create({ accessTokenOptions: { accessToken, instanceUrl, loginUrl: instanceUrl } });`  the new authInfo instance will contain the username/orgId.

### DIFF
--- a/src/authInfo.ts
+++ b/src/authInfo.ts
@@ -841,7 +841,7 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
         authConfig = options;
         const userInfo = await this.retrieveUsername(
           ensureString(options.instanceUrl),
-          ensureString(options.instanceUrl)
+          ensureString(options.accessToken)
         );
         this.fields.username = userInfo?.preferred_username;
         this.fields.orgId = userInfo?.organization_id;
--- a/src/authInfo.ts
+++ b/src/authInfo.ts
@@ -98,6 +98,13 @@ export interface AccessTokenOptions {
   instanceUrl?: string;
 }
 
+/* eslint-disable camelcase */
+type UserInfo = AnyJson & {
+  preferred_username: string;
+  organization_id: string;
+};
+/* eslint-enable camelcase */
+
 /**
  * A function to update a refresh token when the access token is expired.
  */
@@ -832,6 +839,12 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
 
       if (this.isTokenOptions(options)) {
         authConfig = options;
+        const userInfo = await this.retrieveUsername(
+          ensureString(options.instanceUrl),
+          ensureString(options.instanceUrl)
+        );
+        this.fields.username = userInfo?.preferred_username;
+        this.fields.orgId = userInfo?.organization_id;
       } else {
         if (this.options.parentUsername) {
           const parentUserFields = await this.loadAuthFromConfig(this.options.parentUsername);
@@ -1015,19 +1028,15 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
     }
 
     // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
-    // @ts-ignore TODO: need better typings for jsforce
-    const { orgId, userId } = parseIdUrl(authFieldsBuilder.id);
+    // @ts-ignore
+    const { orgId } = parseIdUrl(authFieldsBuilder.id);
 
     let username = this.getUsername();
     if (!username) {
-      username = await this.retrieveUsername(
-        // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
-        // @ts-ignore TODO: need better typings for jsforce
-        authFieldsBuilder.instance_url,
-        authFieldsBuilder.access_token,
-        ensureString(orgId),
-        ensureString(userId)
-      );
+      // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+      // @ts-ignore
+      const userInfo = await this.retrieveUsername(authFieldsBuilder.instance_url, authFieldsBuilder.access_token);
+      username = userInfo?.preferred_username;
     }
     return {
       orgId,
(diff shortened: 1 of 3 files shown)


src/authInfo.ts
```diff
@@ -98,6 +98,13 @@ export interface AccessTokenOptions {
   instanceUrl?: string;
 }
 
+/* eslint-disable camelcase */
+type UserInfo = AnyJson & {
```
peternhale, line 102: These are the relevant attributes obtained by hitting the current user API endpoint.

src/authInfo.ts
```diff
@@ -832,6 +839,12 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
 
       if (this.isTokenOptions(options)) {
         authConfig = options;
+        const userInfo = await this.retrieveUsername(
```
peternhale, line 842: For access token access, retrieve username with access token and capture username and orgid.

src/authInfo.ts
```diff
@@ -1015,19 +1028,15 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
     }
 
     // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
-    // @ts-ignore TODO: need better typings for jsforce
-    const { orgId, userId } = parseIdUrl(authFieldsBuilder.id);
+    // @ts-ignore
+    const { orgId } = parseIdUrl(authFieldsBuilder.id);
```
peternhale, line 1032: userId is no longer needed.

src/authInfo.ts
```diff
@@ -1099,27 +1104,33 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
   private async retrieveUsername(
     instanceUrl: string,
     accessToken: string,
-    orgId: string,
-    userId: string
-  ): Promise<Optional<string>> {
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
peternhale, line 1107: I need to remove these.

src/authInfo.ts
```diff
@@ -1099,27 +1104,33 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
...
+    userId?: string
+  ): Promise<Optional<UserInfo>> {
     // Make a REST call for the username directly.  Normally this is done via a connection
     // but we don't want to create circular dependencies or lots of snowflakes
     // within this file to support it.
-    const apiVersion = 'v42.0'; // hardcoding to v42.0 just for this call is okay.
     const instance = ensure(instanceUrl);
-    const url = `${instance}/services/data/${apiVersion}/sobjects/User/${userId}`;
+    const url = `${instance}/services/oauth2/userinfo`;
```
peternhale, line 1116: Hit endpoint that returns the current user's information.

src/authInfo.ts
```diff
@@ -1099,27 +1104,33 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
...
-
+    let orgId;
     try {
       this.logger.info(`Sending request for Username after successful auth code exchange to URL: ${url}`);
       const response = await new Transport().httpRequest({ url, headers });
       if (response.statusCode >= 400) {
         this.throwUserGetException(response);
       } else {
-        return asString(parseJsonMap(response.body).Username);
+        const json = parseJsonMap(response.body) as UserInfo;
```
peternhale, line 1125: Capture response as UserInfo

src/authInfo.ts
```diff
@@ -832,6 +839,12 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
 
       if (this.isTokenOptions(options)) {
         authConfig = options;
+        const userInfo = await this.retrieveUsername(
+          ensureString(options.instanceUrl),
+          ensureString(options.instanceUrl)
```
WillieRuemmele, line 844: should this be called with `options.access_token`? It's hard to tell in the UI, but it looks like it's be called with the same param twice

src/authInfo.ts
WillieRuemmele: what's the ignore for?

src/authInfo.ts
```diff
@@ -1015,19 +1028,15 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
...
-      username = await this.retrieveUsername(
-        // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
-        // @ts-ignore TODO: need better typings for jsforce
-        authFieldsBuilder.instance_url,
-        authFieldsBuilder.access_token,
-        ensureString(orgId),
-        ensureString(userId)
-      );
+      // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+      // @ts-ignore
```
WillieRuemmele, line 1037: why the ignore?

src/authInfo.ts
peternhale: Yes, good catch, thank you!

src/authInfo.ts
peternhale: @WillieRuemmele w/o ignore I get this TS error
`TS2339: Property 'instance_url' does not exist on type 'TokenResponse'.`

I see this comment at the top of this file
```
// No typings directly available for jsforce/lib/transport
// eslint-disable-next-line @typescript-eslint/ban-ts-ignore
// @ts-ignore
import * as Transport from 'jsforce/lib/transport';
```

src/authInfo.ts
peternhale: same as above

src/authInfo.ts
```diff
@@ -1096,30 +1101,29 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
     };
   }
 
-  private async retrieveUsername(
-    instanceUrl: string,
-    accessToken: string,
-    orgId: string,
-    userId: string
-  ): Promise<Optional<string>> {
+  private async retrieveUsername(instanceUrl: string, accessToken: string): Promise<Optional<UserInfo>> {
```
amphro, line 1104: Rename to `retrieveUserInfo`?

src/authInfo.ts
```diff
@@ -1096,30 +1101,29 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
...
       } else {
-        return asString(parseJsonMap(response.body).Username);
+        const json = parseJsonMap(response.body) as UserInfo;
+        orgId = asString(json.organization_id);
+        return json;
       }
     } catch (err) {
-      throw SfdxError.create('@salesforce/core', 'core', 'AuthCodeUsernameRetrievalError', [orgId, err.message]);
+      throw SfdxError.create('@salesforce/core', 'core', 'AuthCodeUsernameRetrievalError', [
+        orgId || 'unknown',
```
amphro, line 1124: You can modify the message to get rid of the orgId. I think it will always be undefined here, because if the API call fails OR the parse response fails, the orgId will never get set.

amphro (approved):
A couple small things.

WillieRuemmele (approved):

src/authInfo.ts
mshanemc: I like this unversioned endpoint better.

mshanemc (approved):
it's nice.  It preserves the no-store, username-is-accessToken, but gets around that by getting a username from the accessToken if you wanted it stored via the store command.

xyc (approved):
Looks good! thanks for taking care of it. Tested it out and the access tokens are saved and can authenticate with orgs as expected.

PR #406, opened 2021-05-21
feat: move crypto to config file @W-9225317@

amphro:
Adds [tokens to GlobalInfo](https://github.com/forcedotcom/sfdx-core/pull/406/files#diff-bdfa60433ff57d35492d726e3c75348da03bc70413ccd6da4cb761cfc53e7f0aR16) to allow to store other tokens other than just org tokens in the authorizations. Adding another property is a pain. We should think about how we can reduce the boilerplate. Also, `authorizations` doesn't feel right to me since it is only for orgs. This goes to a conversation I had with Mike to keep things specific since we don't have enough data to generalize. Because of that, I think we should change it to "orgs".

To allow functions to story auth data in GlobalInfo, I needed to move crypto to `ConfigStore` rather than `AuthInfo` which is only for orgs. The [migration md](https://github.com/forcedotcom/sfdx-core/compare/v3...td/environments?expand=1&short_path=348fc4e#diff-348fc4eb1f8e6a142dbde66c598b11095caeb92dbc6b74d22003eaae9463c9ba) shows how this is done by a consumer at the lowest level.

Because Crypt is now used in ConfigFile, Crypto can no longer use ConfigFile for [read](https://github.com/forcedotcom/sfdx-core/compare/v3...td/environments?expand=1&short_path=348fc4e#diff-e185377eac4f27bdea2c0a590cffadcb556a90abf145586ececc9d327e5582a3L461) or [write](https://github.com/forcedotcom/sfdx-core/compare/v3...td/environments?expand=1&short_path=348fc4e#diff-e185377eac4f27bdea2c0a590cffadcb556a90abf145586ececc9d327e5582a3L432).

Pay close attention to [ConfigStore](https://github.com/forcedotcom/sfdx-core/compare/v3...td/environments?expand=1&short_path=348fc4e#diff-6466f7d3421bbfa50e7be7643dfd183d6ad9ae258a205747f2e7b3ff762778b4L38) and how I did typings. 

Pay close attention to [AuthInfo](https://github.com/forcedotcom/sfdx-core/compare/v3...td/environments?expand=1&short_path=348fc4e#diff-6466f7d3421bbfa50e7be7643dfd183d6ad9ae258a205747f2e7b3ff762778b4L38) and the removed crypto and cache.


src/config/globalInfoConfig.ts
```diff
@@ -5,37 +5,64 @@
...
+} & Timestamp &
+  JsonMap;
+
+export interface Tokens {
+  [key: string]: Token;
 }
 
 export type SfData = {
-  [SfDataKeys.AUTHORIZATIONS]: Authorizations;
+  [SfDataKeys.AUTHORIZATIONS]: OrgAuthorizations;
```
mdonnalley, line 47: I agree that we should change this to `orgs`

src/config/globalInfoConfig.ts
```diff
@@ -5,37 +5,64 @@
...
-  [key: string]: SfAuthorization & Timestamp;
+export type OrgAuthorization = {
+  alias: Optional<string>;
+  username: Optional<string>;
+  orgId: Optional<string>;
+  instanceUrl: Optional<string>;
+  accessToken?: string;
+  oauthMethod?: 'jwt' | 'web' | 'token' | 'unknown';
+  error?: string;
+} & Timestamp &
```
mdonnalley, line 29: Maybe extract `Timestamp` and `JsonMap` into a new type? e.g.

```typescript
type Entry = Timestamp & JsonMap;

export type OrgAuthorization = {
    ...
} & Entry;

export type Token = {
    ...
} & Entry;
```

src/config/globalInfoConfig.ts
```diff
@@ -65,51 +101,86 @@ export class GlobalInfo extends ConfigFile<ConfigFile.Options> {
     };
   }
 
-  public getContents(): SfData {
-    return this['contents'] as SfData;
-  }
-
-  public getAuthorizations(): Authorizations {
-    return this.getContents()[SfDataKeys.AUTHORIZATIONS];
+  public getAuthorizations(decrypt = false): OrgAuthorizations {
```
mdonnalley, line 104: We'll have to update these methods too if we change it to `orgs`

src/config/configStore.ts
```diff
@@ -249,10 +465,16 @@ export namespace BaseConfigStore {
   /**
    * Options for the config store.
    */
+  // eslint-disable-next-line @typescript-eslint/no-empty-interface
```
mdonnalley, line 468: I don't think you need this rule since it's not empty

src/config/configStore.ts
```diff
@@ -185,8 +261,12 @@ export abstract class BaseConfigStore<T extends BaseConfigStore.Options>
    *
    * @param contents The contents.
    */
-  public setContents(contents?: ConfigContents): void {
-    this.contents = contents || {};
+  public setContents(contents?: P): void {
```
mdonnalley, line 264: I think you can add a default value and it'll inherit the type, e.g.

`setContents(contents: P = {})`

src/config/configStore.ts
```diff
@@ -32,17 +36,23 @@ export type ConfigEntry = [string, ConfigValue];
...
 /**
  * An interface for a config object with a persistent store.
  */
-export interface ConfigStore {
+export interface ConfigStore<P extends ConfigContents = ConfigContents> {
   // Map manipulation methods
   entries(): ConfigEntry[];
-  get(key: string): Optional<ConfigValue>;
-  getKeysByValue(value: ConfigValue): string[];
+  get<K extends Key<P>>(key: K, decrypt: boolean): P[K];
```
mdonnalley, line 47: Love all these type changes 👍

mdonnalley (approved):

RodEsp (commented):
This looks good to me but it seems like `SfOrg` and `SfOrgs` aren't being exported correctly here https://github.com/forcedotcom/sfdx-core/pull/406/files#diff-3f54a992c2e8d9e1f41a84f603da1fb18ea9155d5c58167dd89293b260a30c5bR21 because when I run `yarn build` locally the resulting `lib\exported.js` doesn't contain them.

mdonnalley (approved):

PR #411, opened 2021-05-28
fix: authinfo returning only lowercase usernames.

peternhale:
This PR fixes this [GH issue](https://github.com/forcedotcom/cli/issues/1027), a breaking change to authinfo is causing only lowercase usernames.

@W-9332288@

### DIFF
--- a/src/authInfo.ts
+++ b/src/authInfo.ts
@@ -107,7 +107,10 @@ type UserInfo = AnyJson & {
 type UserInfoResult = AnyJson & {
   preferred_username: string;
   organization_id: string;
-  user_id?: string;
+  user_id: string;
+  urls: AnyJson & {
+    sobjects: string;
+  };
 };
 /* eslint-enable camelcase */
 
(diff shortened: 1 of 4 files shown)


src/authInfo.ts
```diff
@@ -1063,21 +1082,17 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
...
-        // @ts-ignore TODO: need better typings for jsforce
-        authFields.instance_url,
-        authFields.access_token,
-        ensureString(orgId),
-        ensureString(userId)
-      );
+      // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+      // @ts-ignore
+      const userInfo = await this.retrieveUserInfo(authFields.instance_url, authFields.access_token);
+      username = userInfo?.username;
```
mshanemc, line 1095: what happens if we get an undefined username?

mshanemc: or would the error be handled nicely enough in `retrieveUserInfo` ?

src/authInfo.ts
```diff
@@ -1096,30 +1111,33 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
...
-    accessToken: string,
-    orgId: string,
-    userId: string
-  ): Promise<Optional<string>> {
+  private async retrieveUserInfo(instanceUrl: string, accessToken: string): Promise<Optional<UserInfo>> {
     // Make a REST call for the username directly.  Normally this is done via a connection
     // but we don't want to create circular dependencies or lots of snowflakes
     // within this file to support it.
-    const apiVersion = 'v42.0'; // hardcoding to v42.0 just for this call is okay.
+    const apiVersion = 'v51.0'; // hardcoding to v51.0 just for this call is okay.
```
mshanemc, line 1118: suggestion: that userinfo response contains a urls objects with versions in it.  we could not hardcode at all in this new approach

https://help.salesforce.com/articleView?id=sf.remoteaccess_using_userinfo_endpoint.htm&type=5

mshanemc: it gives you back the url for the /sobjects call

<sobjects>https://yourInstance.salesforce.com/services/data/v{version}/sobjects/</sobjects>

mshanemc (approved):
comments are suggestions only.

src/authInfo.ts
amphro: 💯

peternhale: I'll look at that possibility

amphro (approved):

src/authInfo.ts
```diff
@@ -1117,16 +1120,17 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
     // within this file to support it.
     const apiVersion = 'v51.0'; // hardcoding to v51.0 just for this call is okay.
     const instance = ensure(instanceUrl);
-    const userInfoUrl = `${instance}/services/oauth2/userinfo`;
+    const baseUrl = new URL(instance);
+    const userInfoUrl = `${baseUrl}services/oauth2/userinfo`;
```
mshanemc, line 1124: is there a slash or not (vs. line 1133)

src/authInfo.ts
peternhale: @amphro  grabbing the url still requires inserting a version number.

src/authInfo.ts
peternhale: @mshanemc since the function was using the value sent from the user it could end up with a double slash, which is why I passed it through new URL, which guarantees the URL will have a slash on the end.

PR #413, opened 2021-06-03
fix: works with TS4+ and jsforce types

mshanemc:
moves core to ts4.  There were problems with jsforce typings (Never) on plugins that were using ts4+ (all of them)

Main change is in the typing for Connection.request (supports generics but defaults to JsonCollection we were using)

I had to update some devDeps to get around some eslint bugs (https://github.com/typescript-eslint/typescript-eslint/issues/2446) which is probably good but caused a lot of linter noise.  

There are some actual type changes (object => Record<string,unknown>)
And I turned off eslint rules where current standard are way stricter than what was in core.

@W-0@

I tested this locally with plugin-user (where we were seeing the ts compilation errors).


peternhale (approved):
@mshanemc are you comfortable with test coverage?

WillieRuemmele:
a lot more changes than I would've expected, what's your comfort level with this?

WillieRuemmele (approved):

PR #415, opened 2021-06-07
feat: exported function checkLightningDomain()

maggiben:
### What does this PR do?
Exports a function `checkLightningDomain` to check if an org domain is ready DNS-wise.

@W-9236226@


mshanemc (approved):
I like having this in `core`.  Can you also add a WI to update `org:open` to use, too (int eng).

mshanemc (approved):

PR #419, opened 2021-06-15
Sm/polling fix

mshanemc:
@W-9467623@
old pollingClient was using setInterval (issue a request every x units until a satisfactory response is received, without waiting for the result of the previous request)

this one gets a response and then waits the given units before requesting again.


package.json
```diff
@@ -49,7 +49,8 @@
     "jsforce": "^1.10.1",
     "jsonwebtoken": "8.5.0",
     "mkdirp": "1.0.4",
-    "sfdx-faye": "^1.0.9"
+    "sfdx-faye": "^1.0.9",
+    "ts-retry-promise": "^0.6.0"
```
WillieRuemmele, line 53: Is this 3PP approved?

src/status/pollingClient.ts
```diff
@@ -60,74 +54,34 @@ export class PollingClient extends AsyncOptionalCreatable<PollingClient.Options>
...
+        result = await this.options.poll();
+      } catch (error) {
+        errorInPollingFunction = error;
+        // there was an actual error thrown, so we don't want to keep retrying
+        throw new NotRetryableError(error.name);
+      }
+      if (result.completed) {
+        return result.payload;
+      }
+      throw new Error('Operation did not complete.  Retrying...'); // triggers a retry
```
WillieRuemmele, line 71: Are there any tests for this change?

package.json
mshanemc: not yet.  Wanted to get agreement on the change first.

src/status/pollingClient.ts
mshanemc: the tests are the same (it gets hit by its own tests plus myDomainResolver).

We **could** try to simulate the SDR bug (have a slow-returning test function) to prove that calls are only made after responses come back.  Seem like a good use of time?

package.json
WillieRuemmele: I got that `retry` library approved overnight, might as well submit it now

src/status/pollingClient.ts
WillieRuemmele: it could be worth it, just to prove that we've actually fixed it, and ensure it won't happen again. Maybe timebox it?

it would take the PR from good to great

package.json
peternhale: @mshanemc @WillieRuemmele I checked 3pp this morning and I could not find ts-retry-promise in the approved list. However, the [p-retry](https://www.npmjs.com/package/p-retry) module that Shane and I discuss on slack has already been approved. I would like to add that p-retry has millions of weekly downloads vs the relatively new ts-retry-promise, which has weekly activity < 100k.

src/status/pollingClient.ts
mshanemc: add one UT.  Could be useful when the next person feels like swapping out the retry library, etc.

src/status/pollingClient.ts
```diff
@@ -60,74 +54,34 @@ export class PollingClient extends AsyncOptionalCreatable<PollingClient.Options>
...
+    };
+    const finalResult = retryDecorator(doPoll, {
+      timeout: this.options.timeout.milliseconds,
+      delay: this.options.frequency.milliseconds,
```
shetzel, line 75: I like this much more.  It would be cool to expose the backoff options that ts-retry-promise allows.

shetzel (approved):

package.json
mshanemc: approved, thanks to Willie's magic touch.

src/status/pollingClient.ts
mshanemc: yeah...I'm **real** hesitant to change the shape of the API in core.  V3 seems like the time to do it, but I'd change a lot more than just adding those options.

WillieRuemmele (approved):

PR #420, opened 2021-06-17
Core URL class SfdcUrl

maggiben:
# WIP

### What does this PR do?
A new `SfdcUrl` class derived from WHATWG URL API with some extensions that consolidate a lot of url utilities into a single class.

### What functionality is included in this class ?
- [x]  `getJwtAudienceUrl`
- [x]  `isSandboxUrl`
- [x]  `isSalesforceDomain`
- [x]  `isInternalUrl`
- [x]  `isLocalUrl`
- [x]  `checkLightningDomain`
- [x] `lookup` 

### Backward compatibility:
- [x] `getJwtAudienceUrl`
- [x] `sfdc.isInternalUrl`

@W-9457855@


src/util/sfdcUrl.ts
```diff
@@ -0,0 +1,155 @@
...
+export class SfdcUrl extends URL {
+  /**
+   * Salesforce URLs.
+   */
+  public static SANDBOX = 'https://test.salesforce.com';
+  public static PRODUCTION = 'https://login.salesforce.com';
+
+  public constructor(input: string, base?: string | URL) {
+    super(input, base);
+    if (this.protocol !== 'https:') {
```
peternhale, line 21: @maggiben Thomas had mentioned trying to limit the number of warnings emitted to just one per command execution, just to keep the noise to user down. We should consider finding a way to accomplish this. Possibly a cache of some sort with the key being the origin.

peternhale (changes requested):
@maggiben @mshanemc I was expecting to see changes in core that would reflect the availability of this URL class whereever node URL is being used. Will that be forthcoming?

src/util/sfdcUrl.ts
maggiben: I decided to keep this branch "clean" (only the class and it's tests) until everyone feels happy about the functionality and then move on to refactor phase.

mshanemc: 2 PRs seems fine.  It's good to add a note to the PR description if you're planning to do that, so the reviewers understand your intent.

mshanemc: A cache is a good idea.  As more things use this class, getting lots of http warnings on the same command would be painful.
A list of urls that have already been warned about (and maybe a timestamp?) could be a nice touch.  Since it would only get executed when people are doing something insecure, it would never hit most users with any perf problems.  

This is mostly for internals, anyway I think.  

Maybe we can do it via a singleton (ex https://medium.com/swlh/node-js-and-singleton-pattern-7b08d11c726a) before resorting to fs since it's perhaps not important to persist across multiple command runs.

src/util/sfdcUrl.ts
```diff
@@ -0,0 +1,169 @@
...
+      url: new URL(domain),
+      timeout,
+      frequency: new Duration(1, Duration.Unit.SECONDS),
+    });
+
+    await resolver.resolve();
+    return true;
+  }
+
+  public async lookup(): Promise<string> {
```
RodEsp, line 129: Can you add a comment that describes what this method is supposed to return?

RodEsp (approved):

test/unit/util/sfdcUrlTest.ts
```diff
@@ -108,7 +108,7 @@ describe('util/sfdcUrl', () => {
       emitWarningSpy.restore();
     });
 
-    it('listens for the insecure http signal', () => {
+    it('emits the insecure http signal', () => {
```
peternhale, line 111: @maggiben should a test be added to make sure emit is only called once?

peternhale (dismissed):

test/unit/util/sfdcUrlTest.ts
mshanemc: yes...maybe instantiate the SfdcUrl again and verify the count is still 1.

maggiben: Good idea, test created for single call-count

src/authInfo.ts
```diff
@@ -972,7 +916,10 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
   // Build OAuth config for a JWT auth flow
   private async buildJwtConfig(options: OAuth2Options): Promise<AuthFields> {
     const privateKeyContents = await fs.readFile(ensure(options.privateKey), 'utf8');
-    const audienceUrl = await getJwtAudienceUrl(options);
+    const { loginUrl = '' } = options;
```
RodEsp, line 919: Why make an empty string the default?

Won't that throw an error like this one if `loginUrl` isn't defined in `options`?
image

And if so, could we at least show a slightly better error message like "loginUrl is required"?

maggiben: This is a good catch, the original version of `getJwtAudienceUrl` would return `SfdcUrl.PRODUCTION` if no `loginUrl` is given, so we have that option, or we can throw with a good message, what would be our best option ?

RodEsp: Ah, I'll default to someone with more historical context than I here.
@amphro or @mdonnalley thoughts?

I personally would be fine with returning `SfdxUrl.PRODUCTION` though.

src/authInfo.ts
```diff
@@ -916,7 +916,7 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
   // Build OAuth config for a JWT auth flow
   private async buildJwtConfig(options: OAuth2Options): Promise<AuthFields> {
     const privateKeyContents = await fs.readFile(ensure(options.privateKey), 'utf8');
-    const { loginUrl = '' } = options;
+    const { loginUrl = SfdcUrl.PRODUCTION } = options;
     const url = new SfdcUrl(loginUrl);
     const createdOrgInstance = getString(options, 'createdOrgInstance', '').trim().toLowerCase();
     const audienceUrl = await url.getJwtAudienceUrl(createdOrgInstance);
```
peternhale, line 922: @maggiben Audience URL might be a candidate for insecure check.

maggiben: All refactors in pace

test/unit/util/sfdcUrlTest.ts
```diff
@@ -0,0 +1,226 @@
...
+  describe('Insecure HTTP warning', () => {
+    let emitWarningSpy: SinonSpy;
+    beforeEach(() => {
+      $$.SANDBOX.stub(MyDomainResolver.prototype, 'getCnames').resolves(TEST_CNAMES);
+      emitWarningSpy = spyMethod($$.SANDBOX, SfdcUrl.prototype, 'emitWarning');
+    });
+
+    afterEach(() => {
+      $$.SANDBOX.restore();
+      emitWarningSpy.restore();
```
shetzel, line 110: we shouldn't need either of the above calls since it's done for us by using `$$.SANDBOX`

test/unit/util/sfdcUrlTest.ts
```diff
@@ -0,0 +1,226 @@
...
+  });
+
+  describe('getJwtAudienceUrl', () => {
+    const env = new Env();
+    beforeEach(() => {
+      $$.SANDBOX.stub(MyDomainResolver.prototype, 'getCnames').resolves(TEST_CNAMES);
+    });
+
+    afterEach(() => {
+      $$.SANDBOX.restore();
```
shetzel, line 142: This sandbox is restored for us so we can remove this.

test/unit/util/sfdcUrlTest.ts
```diff
@@ -0,0 +1,226 @@
...
+    });
+
+    it('return the jwt audicence url for sandbox domains', async () => {
+      const url = new SfdcUrl('https://organization.sandbox.my.salesforce.com');
+      const response = await url.getJwtAudienceUrl();
+      expect(response).to.be.equal('https://test.salesforce.com');
+    });
+
+    it('should use the correct audience URL for createdOrgInstance beginning with "gs1"', async () => {
+      $$.SANDBOX.restore();
```
shetzel, line 164: shouldn't be necessary to restore here

test/unit/util/sfdcUrlTest.ts
```diff
@@ -0,0 +1,226 @@
...
+      $$.SANDBOX.stub(MyDomainResolver.prototype, 'getCnames').resolves([]);
+      const url = new SfdcUrl('https://foo.bar.baz');
+      const respose = await url.getJwtAudienceUrl('gs1');
+      expect(respose).to.be.equal('https://gs1.salesforce.com');
+    });
+
+    it('should return production URL if domain cannot be resolved', async () => {
+      $$.SANDBOX.restore();
```
shetzel, line 172: shouldn't be necessary to restore here

test/unit/util/sfdcUrlTest.ts
```diff
@@ -0,0 +1,226 @@
...
+      const url = new SfdcUrl('https://organization.sandbox.my.salesforce.com');
+      const response = await url.getJwtAudienceUrl();
+      expect(response).to.be.equal('https://test.salesforce.com');
+    });
+
+    it('should use the correct audience URL for createdOrgInstance beginning with "gs1"', async () => {
+      $$.SANDBOX.restore();
+      $$.SANDBOX.stub(MyDomainResolver.prototype, 'getCnames').resolves([]);
+      const url = new SfdcUrl('https://foo.bar.baz');
+      const respose = await url.getJwtAudienceUrl('gs1');
```
shetzel, line 167: nit: typo.  `response`.  Bunch of the same below too.

test/unit/util/sfdcUrlTest.ts
```diff
@@ -0,0 +1,226 @@
...
+    });
+  });
+
+  describe('lookup', () => {
+    beforeEach(() => {
+      $$.SANDBOX.stub(MyDomainResolver.prototype, 'resolve').resolves(TEST_IP);
+    });
+
+    afterEach(() => {
+      $$.SANDBOX.restore();
```
shetzel, line 193: shouldn't need to restore

src/util/sfdcUrl.ts
```diff
@@ -0,0 +1,179 @@
...
+    }
+  }
+
+  /**
+   * Returns the appropiate jwt audience url for this url.
+   */
+  public async getJwtAudienceUrl(createdOrgInstance?: string): Promise<string> {
+    // environment variable is used as an override
+    if (process.env.SFDX_AUDIENCE_URL) {
+      return process.env.SFDX_AUDIENCE_URL;
```
shetzel, line 36: We might want to consider adding debug output here that the audience URL is being overridden by the env var, and what the value of the env var is.

src/util/sfdcUrl.ts
```diff
@@ -0,0 +1,179 @@
...
+   * Salesforce URLs.
+   */
+  public static SANDBOX = 'https://test.salesforce.com';
+  public static PRODUCTION = 'https://login.salesforce.com';
+
+  public constructor(input: string | URL, base?: string | URL) {
+    super(input.toString(), base);
+    if (this.protocol !== 'https:' && !cache.has(this.origin)) {
+      cache.set(this.origin, {
+        timestamp: new Date().getTime(),
```
mshanemc, line 24: I know we discussed timestamping these, but I think that was before we used the singleton-cache.  We're not doing anything with the timestamps, so that key value could be empty?

Probably doesn't matter.

mshanemc (approved):

src/util/cache.ts
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright (c) 2021, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+class Cache extends Map {}
+
+export default new Cache();
```
shetzel, line 10: Cache is not equal to a Map.  There needs to be a lot more to this class for it to be a proper cache.  This is only used by SfdcUrl right?  Can we move it there?

src/util/sfdcUrl.ts
```diff
@@ -0,0 +1,179 @@
...
+      });
```
shetzel, line 25: This isn't really a cache.  We could just use a `Set` for this to keep track of "already warned origins".

shetzel (changes requested):

src/util/sfdcUrl.ts
maggiben: Using Logger class to debug overriden value

src/util/sfdcUrl.ts
```diff
@@ -0,0 +1,179 @@
...
+      return true;
+    }
+    const myDomainResolver = await MyDomainResolver.create({ url: this });
+    const cnames: string[] = (await myDomainResolver.getCnames()) ?? [];
+    return cnames.some((cname) => new SfdcUrl(cname).isSandboxUrl());
+  }
+
+  private emitWarning(warning: string): void {
+    process.emitWarning(warning);
+  }
```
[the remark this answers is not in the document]
maggiben: Makes it easier for testing, plus we might want to log this event elsewhere (telemety?)

test/unit/status/myDomainResolverTest.ts
```diff
@@ -46,6 +46,22 @@ describe('myDomainResolver', () => {
     expect(lookupAsyncSpy.callCount).to.be.equal(CALL_COUNT);
   });
 
+  describe('disable dns check', () => {
```
maggiben, line 49: I added this test because it was missing

src/util/cache.ts
```diff
@@ -0,0 +1,8 @@
+/*
+ * Copyright (c) 2021, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+export default new Set();
```
shetzel, line 8: This file can be deleted now.

src/util/sfdcUrl.ts
```diff
@@ -0,0 +1,197 @@
...
+import { MyDomainResolver } from '../status/myDomainResolver';
+import { Logger } from '../logger';
+
+export class SfdcUrl extends URL {
+  /**
+   * Salesforce URLs
+   */
+  public static readonly SANDBOX = 'https://test.salesforce.com';
+  public static readonly PRODUCTION = 'https://login.salesforce.com';
+  public static readonly cache: Set<string> = new Set();
```
shetzel, line 20: This should be private.  I can't think of a reason why this would be useful outside of this class.

test/unit/util/sfdcUrlTest.ts
```diff
@@ -0,0 +1,218 @@
...
+
+    it('should be able to do dns lookup', async () => {
+      const url = new SfdcUrl('https://foo.bar.baz');
+      const ip = await url.lookup();
+      expect(ip).to.be.equal(TEST_IP);
+    });
+
+    it('shoult throw on dns resulution failure', async () => {
+      resolveStub.restore();
+      resolveStub = stubMethod($$.SANDBOX, MyDomainResolver.prototype, 'resolve').rejects();
```
shetzel, line 185: Since there are just 2 tests for `lookup` and they each use a different stub I think we should remove the beforeEach and afterEach and just define the stubs in the tests.  That would simplify it a bit.

shetzel (changes requested):

src/util/sfdcUrl.ts
maggiben: I only needed access to the cache to clear it in the tests but, instead of doing that I just a different entry in the set

src/util/cache.ts
maggiben: must have slipped in

shetzel (approved):

mshanemc:
QA notes

Repro
from this branch locally `yarn link` 
from `main` branch on plugin-auth, `yarn link "@salesforce/core"`

set all the required envs 
* TESTKIT_HUB_USERNAME
* TESTKIT_JWT_CLIENT_ID
* TESTKIT_JWT_KEY (one long no-line-break string of the contents of the key)
* TESTKIT_HUB_INSTANCE
* TESTKIT_AUTH_URL

`yarn build && yarn test:nuts`

You can verify that the NUTs are passing on regular core by doing `yarn link "@salesforce/core"` and then the suggested install command.

```


  auth:logout NUTs
    1) "before each" hook for "should remove the org specified by the -u flag (json)"


  auth:accesstoken:store NUTs
    2) "before all" hook: prepare session and ensure environment variables for "should authorize an org using access token (json)"


  auth:jwt:grant NUTs
    3) should authorize an org using jwt (json)
    4) "after each" hook for "should authorize an org using jwt (json)"


  auth:list NUTs
    ✓ should list auth files (json)
    ✓ should list auth files (human readable)


  auth:sfdxurl:store NUTs
    ✓ should authorize an org using sfdxurl (json) (4797ms)
    ✓ should authorize an org using sfdxurl (human readable) (4093ms)


  verify discovery/id of scratch org
    ✓ should have the scratch org in auth files
    ✓ should logout from the org)
    ✓ should NOT have the scratch org in auth files
    5) should login to the org via jwt grant
    6) should have the devhubUsername in the auth file
    7) "after all" hook for "should have the devhubUsername in the auth file"


  7 passing (48s)
  7 failing
```

mshanemc:
QA Notes: plugin-org (open, myDomainResolver stuff) NUTs worked fine.

maggiben:
I was able to successfully run the NUTs on `plugin-auth`:
image

mshanemc (approved):

mshanemc (changes requested):
QA notes: 

good: plugin config is catching bad URLs.  
bad: toolbelt still has an import of jwtAudienceUrl until  this merges: https://github.com/salesforcecli/toolbelt/pull/74

because of https://github.com/salesforcecli/toolbelt/blob/4d5cf46b231702741905d431edd99060267011ef/src/lib/core/force.ts#L36

I didn't catch this in review, only noticed it while getting read to QA stuff.  That would be a breaking change.  Let's avoid that...it'll be a nice puzzle.

checkLightningDomain has the same problem, but I don't think anyone is using that one that I could find.

mshanemc (approved):
nice tests!

PR #429, opened 2021-06-29
feat: @W-9517449@ Allow signupTargetLoginUrl to be overridden via env var

### LINKED ISSUE #427: @W-9517449@ Allow signupTargetLoginUrl to be overridden via an env var
This repository only supports the Salesforce CLI Core libraries if you have a feature request that is not related to the Core Libraries but instead to the Salesforce CLI or the force commands please submit your issue to the forcedotcome/cli repo not this one. 

**Is your feature request related to a problem? Please describe.**

Internal Salesforce teams building against internal systems would like a way to override the signupTargetLoginUrl for internal systems without changing the (shared local + internal + prod) project definition file.  Currently the only avenue is to define it in the project def.  Other attributes are overridable via the env var. 

This property is currently overridable via env var for tests; we would like to also extend it to "prod".

**Describe the solution you'd like**

Allow the signupTargetLoginUrl (currently defined in the project def) to be optionally overridden by an environment variable

**Describe alternatives you've considered**

The alternative would be for internal teams to create a script which replaces the signupTargetLoginUrl on execution with whatever new value for each env; this was deemed to be:

1. Error-prone (if someone forgets to run the script beforehand, they may accidentally deploy to the wrong env)
2. Less user friendly (the script would require OS-independent shell scripting, which means requiring non-devs to install shells or npm just to override a json value... and then this goes back to "more error prone")

**Additional context**

Please see details on: W-9517449

jfeng-salesforce:
Fixes #427 @W-9517449@

### DIFF
--- a/src/sfdxProject.ts
+++ b/src/sfdxProject.ts
@@ -646,6 +646,11 @@ export class SfdxProject {
       if (process.env.FORCE_SFDC_LOGIN_URL) {
         this.projectConfig.sfdcLoginUrl = process.env.FORCE_SFDC_LOGIN_URL;
       }
+
+      // Allow override of signupTargetLoginUrl via env var SFDX_SCRATCH_ORG_CREATION_LOGIN_URL
+      if (process.env.SFDX_SCRATCH_ORG_CREATION_LOGIN_URL) {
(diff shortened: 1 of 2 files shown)


mshanemc (approved):
looks like the plan we came up with during Review Board

peternhale (approved):

PR #431, opened 2021-06-30
fix: set retries to INFINITELY for polling client

### LINKED ISSUE #430: PollingClient can only call the poll method a maximum of 11 times before it times out.
### Summary

Use the PollingClient

### Steps To Reproduce:

1. Create a PollingClient that will poll more than 11 times.
1. Call subscribe
1. wait

### Expected result

It keeps polling until complete.

### Actual result

Always only tries 11 times. The initial plus the default retries for the retry library you are using.

### Additional information

**SFDX CLI Version**(to find the version of the CLI engine run sfdx --version):
N/A

**SFDX plugin Version**(to find the version of the CLI plugin run sfdx plugins --core):
N/A

**OS and version**:
Linux Debian Buster.

jayree:
Fixes #430

@W-9544990@

### DIFF
--- a/src/status/pollingClient.ts
+++ b/src/status/pollingClient.ts
@@ -73,6 +73,7 @@ export class PollingClient extends AsyncOptionalCreatable<PollingClient.Options>
     const finalResult = retryDecorator(doPoll, {
       timeout: this.options.timeout.milliseconds,
       delay: this.options.frequency.milliseconds,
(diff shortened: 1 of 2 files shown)


shetzel (approved):
Thanks Jochen!

mshanemc (approved):

shetzel:
I did some testing with it and looks good!

PR #432, opened 2021-07-01
feat: update generate password functionality

vamsimundra:
What does this PR do?
Changed the functionality of password:generate command. Now generating the password with the Profile Password policies and Organization Password Policies

What issues does this PR fix or reference?
@W-5829533@

### DIFF
--- a/messages/user.json
+++ b/messages/user.json
@@ -6,5 +6,5 @@
   "missingId": "The Salesforce id for the user is required.",
   "problemsDescribingTheUserObject": "Problems occurred while attempting to describe the user object",
   "permsetNamesAreRequired": "The permission set names are missing but required.",
-  "complexityOutOffBound": "Invalid complexity value. Specify a value between 0 and 5, inclusive."
+  "complexityOutOfBound": "Invalid complexity value. Specify a value between 0 and 5, inclusive."
 }
--- a/messages/user.json
+++ b/messages/user.json
@@ -1,9 +1,10 @@
 {
-    "orgRequired": "An org instance is required.",
-    "userQueryFailed": "Failed to query for the user %s.",
-    "invalidHttpResponseCreatingUser": "An invalid http response code (%s) was received while trying to create a user.",
-    "missingFields": "The fields parameters is undefined.",
-    "missingId": "The Salesforce id for the user is required.",
-    "problemsDescribingTheUserObject": "Problems occurred while attempting to describe the user object",
-    "permsetNamesAreRequired": "The permission set names are missing but required."
+  "orgRequired": "An org instance is required.",
+  "userQueryFailed": "Failed to query for the user %s.",
+  "invalidHttpResponseCreatingUser": "An invalid http response code (%s) was received while trying to create a user.",
+  "missingFields": "The fields parameters is undefined.",
+  "missingId": "The Salesforce id for the user is required.",
+  "problemsDescribingTheUserObject": "Problems occurred while attempting to describe the user object",
+  "permsetNamesAreRequired": "The permission set names are missing but required.",
+  "complexityOutOffBound": "INVALID_COMPLEXITY: valid values are between 0 and 5"
 }
--- a/messages/user.json
+++ b/messages/user.json
@@ -6,5 +6,6 @@
   "missingId": "The Salesforce id for the user is required.",
   "problemsDescribingTheUserObject": "Problems occurred while attempting to describe the user object",
   "permsetNamesAreRequired": "The permission set names are missing but required.",
-  "complexityOutOfBound": "Invalid complexity value. Specify a value between 0 and 5, inclusive."
+  "complexityOutOfBound": "Invalid complexity value. Specify a value between 0 and 5, inclusive.",
+  "lengthOutOfBound": "Invalid length value. Specify a value between 8 and 1000, inclusive."
 }
--- a/messages/user.json
+++ b/messages/user.json
@@ -6,5 +6,5 @@
   "missingId": "The Salesforce id for the user is required.",
   "problemsDescribingTheUserObject": "Problems occurred while attempting to describe the user object",
   "permsetNamesAreRequired": "The permission set names are missing but required.",
-  "complexityOutOffBound": "INVALID_COMPLEXITY: valid values are between 0 and 5"
+  "complexityOutOffBound": "Invalid complexity value. Specify a value between 0 and 5, inclusive."
 }
--- a/src/user.ts
+++ b/src/user.ts
@@ -29,27 +29,30 @@ import { SecureBuffer } from './secureBuffer';
 import { SfdxError } from './sfdxError';
 import { sfdc } from './util/sfdc';
 
-const LOWER = 'abcdefghijklmnopqrstuvwxyz';
-const UPPER = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ';
-const NUMBERS = '1234567890';
-const SYMBOLS = ['!', '@', '#', '$', '%', '^', '&', '*', '(', ')', '_', '[', ']', '|', '-'];
-
 const rand = (len: Many<string>): number => Math.floor(Math.random() * len.length);
 
 interface Complexity {
-  LOWER: number;
-  UPPER: number;
-  NUMBERS: number;
-  SYMBOLS: number;
+  [key: string]: boolean | undefined;
+  LOWER?: boolean;
+  UPPER?: boolean;
+  NUMBERS?: boolean;
+  SYMBOLS?: boolean;
 }
 
-const PASSWORD_COMPLEXITY: { [index: string]: Complexity } = {
-  '0': { LOWER: 0, UPPER: 0, NUMBERS: 0, SYMBOLS: 0 },
-  '1': { LOWER: 1, UPPER: 0, NUMBERS: 1, SYMBOLS: 0 },
-  '2': { LOWER: 1, UPPER: 0, NUMBERS: 0, SYMBOLS: 1 },
-  '3': { LOWER: 1, UPPER: 1, NUMBERS: 1, SYMBOLS: 0 },
-  '4': { LOWER: 1, UPPER: 1, NUMBERS: 1, SYMBOLS: 1 },
-  '5': { LOWER: 1, UPPER: 0, NUMBERS: 1, SYMBOLS: 1 },
+const CHARACTERS: { [index: string]: string | string[] } = {
+  LOWER: 'abcdefghijklmnopqrstuvwxyz',
+  UPPER: 'ABCDEFGHIJKLMNOPQRSTUVWXYZ',
+  NUMBERS: '1234567890',
+  SYMBOLS: ['!', '@', '#', '$', '%', '^', '&', '*', '(', ')', '_', '[', ']', '|', '-'],
+};
+
+const PASSWORD_COMPLEXITY: { [index: number]: Complexity } = {
+  '0': { LOWER: true },
+  '1': { LOWER: true, NUMBERS: true },
+  '2': { LOWER: true, SYMBOLS: true },
+  '3': { LOWER: true, UPPER: true, NUMBERS: true },
+  '4': { LOWER: true, UPPER: true, NUMBERS: true, SYMBOLS: true },
+  '5': { LOWER: true, NUMBERS: true, SYMBOLS: true },
 };
 
 const scimEndpoint = '/services/scim/v1/Users';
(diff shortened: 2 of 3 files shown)


src/user.ts
```diff
@@ -211,20 +231,48 @@ export class User extends AsyncCreatable<User.Options> {
...
+  public static generatePasswordUtf8(
+    passwordCondition: PasswordConditions = { length: 10, complexity: 1 }
+  ): SecureBuffer<void> {
     const ALLCHARS = UPPER.concat(LOWER, NUMBERS, SYMBOLS.join(''));
-    const password = [UPPER[rand(UPPER)], LOWER[rand(LOWER)], NUMBERS[rand(NUMBERS)], SYMBOLS[rand(SYMBOLS)]]
-      .concat(
-        Array(PASSWORD_LENGTH - 4)
+    let password: string[] = [];
+
+    // Required number of Lower letter characters. one character of each of the 4 types followed by the remaining random characters, then quasi-randomize
```
mshanemc, line 244: overall, this feels like it could be simpler with booleans.  They all require 1 or 0, nothing higher than 1.

So, for example
```js
if (PASSWORD_COMPLEXITY[passwordCondition.complexity].LOWER) {
    password.push(LOWER[rand(LOWER)])
}
```

mshanemc: another idea would be to iterate the types to avoid the repetition, something like

```ts
[SYMBOLS,NUMBERS,UPPER,LOWER].forEach( charSet => {
    if (PASSWORD_COMPLEXITY[passwordCondition.complexity][charSet]) {
        password.push(charType[rand(charSet)])
    }
})
```

mshanemc: finally, we could generate the right length of password from the start (`length` randow lowercase characters) then overwrite it with whatever other characters are requied and then shuffle that.

Then there's a chance (for cases 0 and 1) where no changes have to be made at all (all the requirements were false)

src/user.ts
```diff
@@ -29,14 +29,29 @@ import { SecureBuffer } from './secureBuffer';
...
 
+interface Complexity {
+  LOWER: number;
+  UPPER: number;
+  NUMBERS: number;
+  SYMBOLS: number;
+}
+
+const PASSWORD_COMPLEXITY: { [index: string]: Complexity } = {
+  '0': { LOWER: 0, UPPER: 0, NUMBERS: 0, SYMBOLS: 0 },
```
mshanemc, line 47: definitely feels like booleans

src/user.ts
```diff
@@ -29,14 +29,29 @@ import { SecureBuffer } from './secureBuffer';
...
-const PASSWORD_LENGTH = 13;
 const LOWER = 'abcdefghijklmnopqrstuvwxyz';
 const UPPER = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ';
 const NUMBERS = '1234567890';
 const SYMBOLS = ['!', '@', '#', '$', '%', '^', '&', '*', '(', ')', '_', '[', ']', '|', '-'];
 
 const rand = (len: Many<string>): number => Math.floor(Math.random() * len.length);
 
+interface Complexity {
+  LOWER: number;
```
mshanemc, line 40: you could also make these all optional unless true.  So `1` could be { LOWER: true } and 5 would be { LOWER:true, NUMBERS:true, SYMBOLS:true }

messages/user.json
```diff
@@ -1,9 +1,10 @@
...
-    "problemsDescribingTheUserObject": "Problems occurred while attempting to describe the user object",
-    "permsetNamesAreRequired": "The permission set names are missing but required."
+  "orgRequired": "An org instance is required.",
+  "userQueryFailed": "Failed to query for the user %s.",
+  "invalidHttpResponseCreatingUser": "An invalid http response code (%s) was received while trying to create a user.",
+  "missingFields": "The fields parameters is undefined.",
+  "missingId": "The Salesforce id for the user is required.",
+  "problemsDescribingTheUserObject": "Problems occurred while attempting to describe the user object",
+  "permsetNamesAreRequired": "The permission set names are missing but required.",
+  "complexityOutOffBound": "INVALID_COMPLEXITY: valid values are between 0 and 5"
```
WillieRuemmele, line 9: Do we need to include `INVALID_COMPLEXITY` can we make it less server-like, and more friendly? Have you worked with Juliet on this message?

src/user.ts
```diff
@@ -211,20 +234,30 @@ export class User extends AsyncCreatable<User.Options> {
...
+      throw SfdxError.create('@salesforce/core', 'user', 'complexityOutOffBound');
+    }
+    let password: string[] = [];
+    ['SYMBOLS', 'NUMBERS', 'UPPER', 'LOWER'].forEach((charSet) => {
+      if (PASSWORD_COMPLEXITY[passwordCondition.complexity][charSet]) {
+        password.push(CHARACTERS[charSet][rand(CHARACTERS[charSet])]);
+      }
+    });
 
+    // Concatinating remaining length randomly with all available characters
```
mshanemc, line 254: the comment is no longer true (it's just going to be lowers).

messages/user.json
mshanemc, line 9:
```suggestion
  "complexityOutOfBound": "INVALID_COMPLEXITY: valid values are between 0 and 5"
```

mshanemc: and also get Juliet to check the wording/format of this message.  The rest of the messages are complete sentences ending in a `.` so try to make it like that.

test/unit/userTest.ts
```diff
@@ -127,6 +127,32 @@ describe('User Tests', () => {
...
+    it('Should generate length 12 and complexity 1 password containing at least one number and one  lowercase letter ', () => {
+      const passwordCondition = { length: 12, complexity: 1 };
+      const password: SecureBuffer<void> = User.generatePasswordUtf8(passwordCondition);
+      password.value((buffer: Buffer): void => {
+        const passwordAsArrayOfCharacters = buffer.toString('utf8').split('');
+        expect(passwordAsArrayOfCharacters.length).to.be.equal(12);
+        expect(
+          passwordAsArrayOfCharacters.some((char) => ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9'].includes(char))
+        );
+        expect(passwordAsArrayOfCharacters.some((char) => char.toLowerCase() !== char.toLowerCase()));
```
WillieRuemmele, line 140: is `char.toLowerCase() !== char.toLowerCase()` correct? seems like you're comparing the same thing to itself

test/unit/userTest.ts
```diff
@@ -127,6 +127,32 @@ describe('User Tests', () => {
...
     });
+
+    it('Should generate length 12 and complexity 1 password containing at least one number and one  lowercase letter ', () => {
+      const passwordCondition = { length: 12, complexity: 1 };
+      const password: SecureBuffer<void> = User.generatePasswordUtf8(passwordCondition);
+      password.value((buffer: Buffer): void => {
+        const passwordAsArrayOfCharacters = buffer.toString('utf8').split('');
+        expect(passwordAsArrayOfCharacters.length).to.be.equal(12);
+        expect(
+          passwordAsArrayOfCharacters.some((char) => ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9'].includes(char))
```
mshanemc, line 138: regex `\d` ?

test/unit/userTest.ts
```diff
@@ -127,6 +127,32 @@ describe('User Tests', () => {
...
+      });
+    });
+    it('Should generate length 14 and complexity 3 password containing at least one number and one  lowercase letter ', () => {
+      const passwordCondition = { length: 12, complexity: 1 };
+      const password: SecureBuffer<void> = User.generatePasswordUtf8(passwordCondition);
+      password.value((buffer: Buffer): void => {
+        const passwordAsArrayOfCharacters = buffer.toString('utf8').split('');
+        expect(passwordAsArrayOfCharacters.length).to.be.equal(12);
+        expect(
+          passwordAsArrayOfCharacters.some((char) => ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9'].includes(char))
```
mshanemc, line 150: this could be cleaner with regex, too.

test/unit/userTest.ts
```diff
@@ -127,6 +127,32 @@ describe('User Tests', () => {
...
+        );
+        expect(passwordAsArrayOfCharacters.some((char) => char.toLowerCase() !== char.toLowerCase()));
+        expect(passwordAsArrayOfCharacters.some((char) => char.toLowerCase() !== char.toLowerCase()));
```
WillieRuemmele, line 153: The test says it is testing length 14, complexity 3. But the test is testing 12/1

test/unit/userTest.ts
```diff
@@ -127,6 +127,32 @@ describe('User Tests', () => {
...
+    it('Should generate length 14 and complexity 3 password containing at least one number and one  lowercase letter ', () => {
+      const passwordCondition = { length: 12, complexity: 1 };
+      const password: SecureBuffer<void> = User.generatePasswordUtf8(passwordCondition);
+      password.value((buffer: Buffer): void => {
+        const passwordAsArrayOfCharacters = buffer.toString('utf8').split('');
+        expect(passwordAsArrayOfCharacters.length).to.be.equal(12);
+        expect(
+          passwordAsArrayOfCharacters.some((char) => ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9'].includes(char))
+        );
+        expect(passwordAsArrayOfCharacters.some((char) => char.toLowerCase() !== char.toLowerCase()));
```
WillieRuemmele, line 152: same as above

src/user.ts
```diff
@@ -29,14 +29,32 @@ import { SecureBuffer } from './secureBuffer';
...
+  SYMBOLS: ['!', '@', '#', '$', '%', '^', '&', '*', '(', ')', '_', '[', ']', '|', '-'],
+};
+
+const PASSWORD_COMPLEXITY: { [index: number]: Complexity } = {
+  '0': { LOWER: true },
+  '1': { LOWER: true, NUMBERS: true },
+  '2': { LOWER: true, SYMBOLS: true },
+  '3': { LOWER: true, UPPER: true, NUMBERS: true },
+  '4': { LOWER: true, NUMBERS: true, SYMBOLS: true },
+  '5': { LOWER: true, UPPER: true, NUMBERS: true, SYMBOLS: true },
```
WillieRuemmele, line 55: is 4 or 5 the most complex? We need to standardize on the server options

messages/user.json
jshackell-sfdc, line 9:
```suggestion
  "complexityOutOffBound": "Invalid complexity value. Valid values are between 0 and 5. "
```

mshanemc (changes requested):
the function seems good, message needs help.

Thinking about the tests...definitely add one to test the complexity boundary error (ex: `9`).

what are the boundaries of the password length?  n >=8 ?  Writing a test with length=6 that you expect to fail would be a good.

WillieRuemmele (commented):
Functionality looks good, tests could use some work

messages/user.json
jshackell-sfdc: Also is there a reason this error message is in sfdx-core and not plugin-user, together with lengthOutOfBound?  If possible I think it's better to keep them all together, unless there's a good reason not to.

Finally, maybe name the variable complexityOutOfBound? (just one "f").

src/user.ts
jshackell-sfdc: Hmm, Vamsi originally had 4 and 5 switched (at least that's what he showed me when I worked on the description) and I asked the same question!  I think he switched it.  But did I steer him wrong?

WillieRuemmele: I think it makes more logical sense to have higher number = more security, but I thought the server had 4 as the most secure option... we need to make a note of what we're doing, and what the server is expecting

test/unit/userTest.ts
```diff
@@ -127,6 +127,36 @@ describe('User Tests', () => {
...
+        const complexity3Regex = new RegExp('^(?=.*[a-z])(?=.*[A-Z])(?=.*[0-9])(?=.{14})');
+
+        expect(complexity3Regex.test(passwordAsArrayOfCharacters));
+      });
+    });
+
+    it('Should throw an error beacuse of complexity', () => {
+      try {
+        const passwordCondition = { length: 14, complexity: 9 };
+        User.generatePasswordUtf8(passwordCondition);
```
WillieRuemmele, line 155: you need to add something like `assert.fail('the above should throw an error')` below `generatePasswordUtf8()` so incase the function doesn't throw an error, the test will fail

src/user.ts
jshackell-sfdc: @vamsimundra , please go with what the devs are saying!  Like I said earlier, to me the order you originally showed me looked wrong, which is why I flagged it, but I should really stay out of it.  But please make sure that whatever you decide is accurately reflected in the --help output.  Thanks

src/user.ts
```diff
@@ -211,20 +234,32 @@ export class User extends AsyncCreatable<User.Options> {
...
-          .map(() => ALLCHARS[rand(ALLCHARS)])
-      )
-      .sort(() => Math.random() - 0.5);
+  public static generatePasswordUtf8(
+    passwordCondition: PasswordConditions = { length: 13, complexity: 5 }
+  ): SecureBuffer<void> {
+    if (!PASSWORD_COMPLEXITY[passwordCondition.complexity]) {
+      throw SfdxError.create('@salesforce/core', 'user', 'complexityOutOfBound');
+    }
+    if (passwordCondition.length < 8 || passwordCondition.length > 1000)
```
WillieRuemmele, line 247: NIT: feel free to ignore this, but can we add {} for the second if statement? can we also add a new line between the error handling/throwing section, and the `let password`?

WillieRuemmele (approved):

mshanemc (approved):

PR #433, opened 2021-07-01
feat: use .sf/config.json @W-9546117@

amphro:
@W-9546117@

### DIFF
--- a/src/config/configAggregator.ts
+++ b/src/config/configAggregator.ts
@@ -6,7 +6,7 @@
  */
 
 import { AsyncOptionalCreatable, merge, snakeCase, sortBy } from '@salesforce/kit';
-import { AnyJson, definiteEntriesOf, Dictionary, isJsonMap, JsonMap, Optional } from '@salesforce/ts-types';
+import { AnyJson, definiteEntriesOf, Dictionary, isArray, isJsonMap, JsonMap, Optional } from '@salesforce/ts-types';
 import { Messages } from '../messages';
 import { Config, ConfigPropertyMeta } from './config';
 
@@ -161,9 +161,9 @@ export class ConfigAggregator extends AsyncOptionalCreatable<JsonMap> {
    *
    * @param key The key of the property.
    */
-  public getPropertyValue(key: string): Optional<AnyJson> {
+  public getPropertyValue<T extends AnyJson>(key: string): Optional<T> {
     if (this.getAllowedProperties().some((element) => key === element.key)) {
-      return this.getConfig()[key];
+      return this.getConfig()[key] as T;
     } else {
       throw messages.createError('unknownConfigKey', [key]);
     }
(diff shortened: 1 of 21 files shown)


src/config/configAggregator.ts
```diff
@@ -290,6 +290,17 @@ export class ConfigAggregator extends AsyncOptionalCreatable<JsonMap> {
     return this;
   }
 
+  /**
+   * Add an allowed config property.
+   */
+  public addAllowedProperties(configMetas: ConfigPropertyMeta | ConfigPropertyMeta[]): void {
+    if (isArray(configMetas)) {
```
RodEsp, line 297: Why do we have an imported `isArray` method when we can just use `Array.isArray()`?

src/org/authInfo.ts
```diff
@@ -640,13 +641,13 @@ export class AuthInfo extends AsyncOptionalCreatable<AuthInfo.Options> {
...
   /**
-   * Set the defaultusername or the defaultdevhubusername to the alias if
+   * Set the target-env (default) or the target-hub to the alias if
    * it exists otherwise to the username. Method will try to set the local
    * config first but will default to global config if that fails.
    *
    * @param options
    */
-  public async setAsDefault(options: { defaultUsername?: boolean; defaultDevhubUsername?: boolean }) {
+  public async setAsDefault(options: { org?: boolean; hub?: boolean } = { org: true }) {
```
RodEsp, line 650: Can we make this be `devhub` instead of just `hub` for context?

src/org/authInfo.ts
```diff
@@ -640,13 +641,13 @@ export class AuthInfo extends AsyncOptionalCreatable<AuthInfo.Options> {
   }
 
   /**
-   * Set the defaultusername or the defaultdevhubusername to the alias if
+   * Set the target-env (default) or the target-hub to the alias if
```
RodEsp, line 644: Can we make this be target-devhub instead of just target-hub for context?

src/org/org.ts
```diff
@@ -165,7 +164,9 @@ export class Org extends AsyncOptionalCreatable<Org.Options> {
   public async checkScratchOrg(devHubUsernameOrAlias?: string): Promise<Partial<AuthFields>> {
     let aliasOrUsername = devHubUsernameOrAlias;
     if (!aliasOrUsername) {
-      aliasOrUsername = asString(this.configAggregator.getPropertyValue(Config.DEFAULT_DEV_HUB_USERNAME));
+      aliasOrUsername =
+        this.configAggregator.getPropertyValue<string>(OrgConfigProperties.TARGET_HUB) ||
```
RodEsp, line 168: TARGET_DEV_HUB instead of just TARGET_HUB for context?

mdonnalley (changes requested):
Code changes look good to me but requesting changes because I think these changes need to be mentioned in the migration guide

amphro:
Added

mdonnalley (approved):

src/config/configAggregator.ts
amphro: I'm going to leave it for now since it is already brought in by ts-types, but your right we probably should be using that instead and just remove it from ts-types.

PR #437, opened 2021-07-14
refactor: use jsforce@v2

amphro:
`jsforce@v2` (typescript) has released a beta that includes removing the `request` lib.

**Note**: This is targeted to `td/config` but should be changed to `v3` when that PR is merged.

TODO:
* Some methods and types are removed. A section should be added to the migration guide.
* A lot of testing. Since the request lib was replaced with `node-fetch`, all usage of jsforce should be tested, ideally through nuts. Specifically, deploys and any requests that uses form data.

@W-9127326@


src/org/connection.ts
```diff
@@ -252,31 +248,12 @@ export class Connection extends JSForceConnection {
...
-    return super.request(requestInfo, options) as Promise<JsonCollection>;
-  }
-
-  /**
-   * Send REST API request with given HTTP request info, with connected session information
-   * and SFDX headers. This method returns a raw http response which includes a response body and statusCode.
-   *
-   * @param request HTTP request object or URL to GET request.
-   */
-  public async requestRaw(request: RequestInfo): Promise<JsonMap> {
```
amphro, line 269: No longer support requestRaw.

src/org/connection.ts
```diff
@@ -76,7 +71,7 @@ export interface Tooling extends JSForceTooling {
    * @param soql The SOQL string.
    * @param options The query options.  NOTE: the autoFetch option will always be true.
    */
-  autoFetchQuery<T>(soql: string, options?: ExecuteOptions): Promise<QueryResult<T>>;
+  autoFetchQuery<T>(soql: string, options?: QueryOptions): Promise<QueryResult<T>>;
```
amphro, line 74: It would be nice if `autoFetchQuery` was in jsforce.

mdonnalley:
Closed in favor of https://github.com/forcedotcom/sfdx-core/pull/457

PR #439, opened 2021-07-16
force:org:open force:org:open --urlonly display bug fix 

vamsimundra:
What does this PR do?
Modifies access token regex in logger which replaces the access token to hidden while logging the information. 
It fixes the following issue: https://github.com/forcedotcom/sfdx-core/issues/153

Testing Notes:
Try to reproduce the issue before. Test that functionality with the current branch. Test force:org:open --urlonly command with 
different tags by enabling and disabling SFDX_LOG_LEVEL

What issues does this PR fix or reference?
@W-6336711@

### DIFF
--- a/test/unit/loggerTest.ts
+++ b/test/unit/loggerTest.ts
@@ -217,13 +217,12 @@ describe('Logger', () => {
   });
 
   describe('filters', () => {
-    const sid = 'SIDHERE!';
+    const sid = '00D55000000M2qA!AQ0AQHg3LnYDOyobmH07';
     const simpleString = `sid=${sid}`;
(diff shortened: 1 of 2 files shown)


test/unit/loggerTest.ts
```diff
@@ -217,13 +217,12 @@ describe('Logger', () => {
   });
 
   describe('filters', () => {
-    const sid = 'SIDHERE!';
+    const sid = '00D55000000M2qA!AQ0AQHg3LnYDOyobmH07';
```
mshanemc, line 220: I like that you're testing with real-ish data

mshanemc (approved):
to the PR or the WI , please add some QA notes about how you think it should be tested (how to set up, things to try)

PR #447, opened 2021-07-28
Add getUser method to GlobalInfo

RodEsp:
@W-9608729@

### DIFF
--- a/src/config/globalInfoConfig.ts
+++ b/src/config/globalInfoConfig.ts
@@ -221,6 +221,12 @@ export class GlobalInfo extends ConfigFile<ConfigFile.Options, SfInfo> {
     aliases.forEach((alias) => this.unsetAlias(alias));
   }
 
+  public getUser(usernameOrAlias: string): string | null {
+    return (
+      this.getOrg(usernameOrAlias)?.username ?? this.getToken(usernameOrAlias)?.user ?? this.getAliasee(usernameOrAlias)
+    );
+  }
+
   public set(key: string, value: ConfigValue): void {
     if (isPlainObject(value)) {
       value = this.timestamp(value as JsonMap);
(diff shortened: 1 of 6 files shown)


src/config/globalInfoConfig.ts
```diff
@@ -221,6 +221,12 @@ export class GlobalInfo extends ConfigFile<ConfigFile.Options, SfInfo> {
     aliases.forEach((alias) => this.unsetAlias(alias));
   }
 
+  public getUser(usernameOrAlias: string): string | null {
+    return (
```
peternhale, line 225: @RodEsp are the `()` necessary around the expression?

test/unit/messagesTest.ts
```diff
@@ -171,7 +171,7 @@ describe('Messages', () => {
       expect(importMessageFileStub.secondCall.args[1]).to.equal(path.join(expectedMsgDirPath, msgFiles[1]));
     });
 
-    it('should not remove the "/lib" if dist is already removed', () => {
+    it.skip('should not remove the "/lib" if dist is already removed', () => {
```
peternhale, line 174: Why the skip?

peternhale (approved):
Just some questions

test/unit/messagesTest.ts
RodEsp: Oops... removed it.

I do it locally cuz this doesn't work on Windows for some reason... causes a JavaScript heap overload. Pretty sure it's something to do with the paths cuz there are another like 10 tests that fail because of path reasons on Windows.
image

src/config/globalInfoConfig.ts
RodEsp: VSCode auto-added them when it decided to put the expression on its own line.

RodEsp:
Closed in favor of #448

PR #448, opened 2021-07-29
Add Accessor classes to GlobalInfo

mdonnalley:
Adds `OrgAccessor`, `TokenAccessor`, and `AliasAccessor` classes to `GlobalInfo` so that we don't have a million methods stuffed into `GlobalInfo`

[skip-validate-pr]


peternhale (approved):

RodEsp:
Any reason not to put the accessors in their own files?

RodEsp (changes requested):

src/config/globalInfoConfig.ts
```diff
@@ -104,164 +111,219 @@ export class GlobalInfo extends ConfigFile<ConfigFile.Options, SfInfo> {
...
-   * @param aliasee - The entity whose aliases you want to find
+   * @param entity the aliasable entity that you want to get the aliases of
    */
-  public getAliases(aliasee: string | Partial<SfOrg> | Partial<SfToken>): string[] {
-    const aliaseeName = this.getAliaseeName(aliasee);
-    const matchedAliases = Object.entries(this.getAllAliases()).filter((entry) => entry[1] === aliaseeName);
-
-    // Only return the actual aliases
-    return matchedAliases.map((entry) => entry[0]);
+  public getAll(entity: Aliasable): string[] {
```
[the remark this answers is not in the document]
mdonnalley: Yeah I struggled with that one. I was trying to avoid making the naming too redundant, e.g. `globalInfo.aliases.getAliases`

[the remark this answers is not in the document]
RodEsp: Hmm, yeah... I suppose this is fine since we have comments and smart IDEs and whatnot.

RodEsp (approved):

src/globalInfo/accessors/tokenAccessor.ts
```diff
@@ -0,0 +1,38 @@
...
+
+import { Optional } from '@salesforce/ts-types';
+import { GlobalInfo } from '../globalInfoConfig';
+import { SfTokens, SfToken, Timestamp, SfInfoKeys } from '../types';
+
+export class TokenAccessor {
+  public constructor(private globalInfo: GlobalInfo) {}
+
+  public getAll(decrypt = false): SfTokens {
+    return this.globalInfo.get(SfInfoKeys.TOKENS, decrypt) || {};
```
peternhale, line 16: `||` vs `??`

src/globalInfo/types.ts
```diff
@@ -0,0 +1,54 @@
...
+ */
+import { JsonMap, Optional } from '@salesforce/ts-types';
+
+export enum SfInfoKeys {
+  ORGS = 'orgs',
+  TOKENS = 'tokens',
+  ALIASES = 'aliases',
+}
+
+export type Timestamp = { timestamp: string };
```
peternhale, line 15: @mdonnalley what is the relationship between Timestamp type and the timestamp property in a structured config value? Here is is defined as a string, but in [GlobalInfoConfig](https://www.npmjs.com/package/@salesforce/plugin-analytics) it is a Date.

test/unit/globalInfo/accessors/aliasAccessorTest.ts
```diff
@@ -0,0 +1,149 @@
...
+  const token = { token: '123', url: 'https://login.salesforce.com', user: username1 };
+  const $$ = testSetup();
+
+  beforeEach(async () => {
+    $$.SANDBOXES.CONFIG.restore();
+
+    stubMethod($$.SANDBOX, GlobalInfo.prototype, 'read').callsFake(async function (this: GlobalInfo) {
+      const contents = {
+        orgs: { [username1]: org },
+        aliases: {
```
peternhale, line 28: Way cool!

test/unit/globalInfo/accessors/aliasAccessorTest.ts
```diff
@@ -0,0 +1,149 @@
...
+
+    it('should return a username if an alias was provided', async () => {
+      const globalInfo = await GlobalInfo.create();
+      const username = globalInfo.aliases.getUsername(alias1);
+      expect(username).to.equal(username1);
+    });
+  });
+
+  describe('set', () => {
+    it('should set an alias for a username', async () => {
```
peternhale, line 96: @mdonnalley 

`should set alias to an alias` y=foobar@salesforce.com, a1=y, a2=a1 then getUsername('a2') = ?
`should not set an alias to itself` alias:set foobar=foobar (this is currently possible)

test/unit/globalInfo/accessors/orgAccessorTest.ts
```diff
@@ -0,0 +1,102 @@
...
+  describe('set', () => {
+    it('should set the org', async () => {
+      const globalInfo = await GlobalInfo.create();
+      const newOrg = { username: 'foobar@baz.com', token: '123' } as unknown as SfOrg;
+      globalInfo.orgs.set('foobar@baz.com', newOrg);
+      const result = globalInfo.orgs.get('foobar@baz.com');
+      expect(result).to.deep.equal(newOrg);
+    });
+
+    it('should add the username if does not exist on the object', async () => {
```
peternhale, line 73:
```suggestion
    it('should add the username if does not exist in global orgs'...
```

peternhale (changes requested):

src/globalInfo/types.ts
mdonnalley: The Date gets converted to a string when it gets written to the file. I think @RodEsp had some ideas around how to improve the types here

src/globalInfo/accessors/aliasAccessor.ts
```diff
@@ -0,0 +1,122 @@
...
+   */
+  public getAll(): SfAliases;
+  /**
+   * Returns all the aliases for a given entity
+   *
+   * @param entity the aliasable entity that you want to get the aliases of
+   */
+  public getAll(entity: Aliasable): string[];
+  public getAll(entity?: Aliasable): string[] | SfAliases {
+    const all = this.globalInfo.get(SfInfoKeys.ALIASES) || {};
```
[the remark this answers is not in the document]
mdonnalley: @peternhale No I think `||` is better in this scenario since `??` will only use the right hand-side operand if the left hand-side is `null` or `undefined`. While it's extremely unlikely, if `this.globalInfo.get(SfInfoKeys.ALIASES)` resolved to an empty string `''` and we used `??` then we'd end up with that empty string instead of the empty object - which would have bad downstream effects.

src/globalInfo/accessors/orgAccessor.ts
```diff
@@ -0,0 +1,45 @@
...
+    return !!this.getAll()[username];
+  }
+
+  public set(username: string, org: SfOrg): void {
+    // For legacy, and to keep things standard, some things wants the username in auth info.
+    if (!org.username) org.username = username;
+    this.globalInfo.set(`${SfInfoKeys.ORGS}["${username}"]`, org);
+  }
+
+  public update(username: string, authorization: Partial<SfOrg>): void {
```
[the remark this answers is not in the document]
mdonnalley: No reason. I'll update it to be consistent

test/unit/globalInfo/accessors/aliasAccessorTest.ts
mdonnalley: Good point but there's never been any safeguard against aliasing something that isn't an org username. I'm not sure that we even want to enforce that since there may be some valid use cases in the future where we do want to create aliases for things that are orgs.

src/globalInfo/accessors/aliasAccessor.ts
```diff
@@ -0,0 +1,119 @@
...
+    }
+  }
+
+  /**
+   * Returns the first alias found for a given entity
+   *
+   * @param entity the aliasable entity that you want to get the alias of
+   */
+  public get(entity: Aliasable): Nullable<string> {
+    return this.getAll(entity).find((alias) => alias) ?? null;
```
RodEsp, line 46: Why use `.find()` here instead of just an array index?


```suggestion
    return this.getAll(entity)[0] ?? null;
```

RodEsp, line 46: Why use `.find()` here instead of just an array index?

This is way more readable if you ask me.
```suggestion
    return this.getAll(entity)[0] ?? null;
```

src/globalInfo/globalInfoConfig.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  protected async init(): Promise<void> {
+    await this.initCrypto();
+    const contents = Global.SFDX_INTEROPERABILITY ? await this.mergeWithSfdxData() : await this.loadSfData();
+    this.setContents(contents);
+    await this.write(contents);
+  }
+
+  private timestamp<T extends JsonMap>(data: T): T {
+    return Object.assign(data, { timestamp: new Date() });
```
RodEsp, line 101: A little simpler, [a little faster](https://jsben.ch/uzw0r), and [doesn't trigger setters](https://thecodebarbarian.com/object-assign-vs-object-spread.html#differences-versus-objectassign).
```suggestion
    return {...data, timestamp: new Date()};
```

src/globalInfo/globalInfoConfig.ts
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  private async loadSfData(): Promise<SfInfo> {
+    const data = await this.read();
+    return Object.assign(GlobalInfo.emptyDataModel, data);
```
RodEsp, line 106:
```suggestion
    return { ...GlobalInfo.emptyDataModel, ...data };
```

test/unit/globalInfo/accessors/orgAccessorTest.ts
```diff
@@ -0,0 +1,102 @@
...
+      const result = globalInfo.orgs.get('foobar@baz.com');
+      expect(result.username).to.deep.equal('foobar@baz.com');
+    });
+  });
+
+  describe('update', () => {
+    const instanceUrl = 'https://login.salesforce.com';
+    it('should update the org', async () => {
+      const globalInfo = await GlobalInfo.create();
+      const newOrg = Object.assign(org, { instanceUrl });
```
RodEsp, line 86:
```suggestion
      const newOrg = { ...org, instanceUrl };
```

test/unit/globalInfo/accessors/orgAccessorTest.ts
```diff
@@ -0,0 +1,102 @@
...
+      globalInfo.orgs.update(username, newOrg);
+      const result = globalInfo.orgs.get(username);
+      expect(result.instanceUrl).to.deep.equal(instanceUrl);
+    });
+
+    it('should add the username if does not exist on the object', async () => {
+      const globalInfo = await GlobalInfo.create();
+      const newOrg = Object.assign(org, { instanceUrl });
```
RodEsp, line 94:
```suggestion
      const newOrg = { ...org, instanceUrl};
```

test/unit/globalInfo/accessors/tokenAccessorTest.ts
```diff
@@ -0,0 +1,93 @@
...
+      const result = globalInfo.tokens.get('foobar@baz.com');
+      expect(result).to.deep.equal(newToken);
+    });
+  });
+
+  describe('update', () => {
+    it('should update the token', async () => {
+      const globalInfo = await GlobalInfo.create();
+      const instanceUrl = 'https://login.salesforce.com';
+      const newToken = Object.assign(token, { instanceUrl });
```
RodEsp, line 78:
```suggestion
      const newToken = { ...token, instanceUrl };
```

src/globalInfo/types.ts
RodEsp: Why does it have to be a string here tho?

Can't whatever writes it `.toString()` it and whatever loads it `new Date(timestamp)` it? Or it that a much bigger change?


Def not required for the scope of this PR tho.

RodEsp (changes requested):

src/globalInfo/types.ts
RodEsp: Why does it have to be a string here tho?

Can't whatever writes it `.toString()` it and whatever loads it `new Date(timestamp)` it? Or it that a much bigger change?

Def not required for the scope of this PR tho.

src/globalInfo/accessors/aliasAccessor.ts
peternhale: What if getAll returns [null, 'myalias']?

peternhale (approved):

src/globalInfo/accessors/aliasAccessor.ts
RodEsp: It can't. It only returns an array of strings or an empty array.

RodEsp: I guess it could be a problem if someone manually edits the `sf.json` file to do something weird like
```
{
	"aliases": {
 		"": "user@sf.com"
	}
}
```

test/unit/globalInfo/accessors/tokenAccessorTest.ts
peternhale: I get a sense you don't like Object.assign ;)

RodEsp: Haha, I don't actively dislike it, I just think the spread operator serves us better here. [They're not _quite_ the same thing](https://github.com/forcedotcom/sfdx-core/pull/448#discussion_r679525368).

And the spread operator _is_ less verbose which I do like better.

peternhale (approved):

RodEsp (approved):

PR #452, opened 2021-08-03
fix: merge main(v2) into v3

peternhale:
@W-9670244@

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -281,20 +281,6 @@ export class Config extends ConfigFile<ConfigFile.Options, ConfigProperties> {
   }
 
   /**
-   * Gets default options.
-   *
-   * @param isGlobal Make the config global.
-   * @param filename Override the default file. {@link Config.getFileName}
-   */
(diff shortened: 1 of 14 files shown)


src/config/config.ts
```diff
@@ -279,6 +280,20 @@ export class Config extends ConfigFile<ConfigFile.Options, ConfigProperties> {
     });
   }
 
+  /**
+   * Gets default options.
+   *
+   * @param isGlobal Make the config global.
+   * @param filename Override the default file. {@link Config.getFileName}
+   */
+  public static getDefaultOptions(isGlobal = false, filename?: string): ConfigFile.Options {
```
mdonnalley, line 289: I don't think needs to be added. See the note in the `constructor` method

mdonnalley (changes requested):

src/config/config.ts
peternhale: Removing

mdonnalley (approved):

WillieRuemmele (approved):
I'm not familiar with these changes

PR #458, opened 2021-08-09
fix: add predicate function to list all auths

peternhale:
@W-9670267@ Modify listAllAuthorizations function that allows caller to filter results
Default for filter is to return all authorizations to protect backward compatibility.

### DIFF
--- a/src/org/authInfo.ts
+++ b/src/org/authInfo.ts
@@ -98,7 +98,7 @@ export type OrgAuthorization = {
   instanceUrl?: string;
   accessToken?: string;
   error?: string;
-  isExpired: boolean;
+  isExpired: boolean | unknown;
 };
 
 /**
@@ -332,7 +332,7 @@ export class AuthInfo extends AsyncOptionalCreatable<AuthInfo.Options> {
 
   /**
    * Get a list of all authorizations based on auth files stored in the global directory.
-   * One can supply a filter (see @param orgAuthFilter) and calling this function withoyt
+   * One can supply a filter (see @param orgAuthFilter) and calling this function without
    * a filter will return all authorizations.
    *
    * @param orgAuthFilter a boolean value or a predicate function that filters authorizations
@@ -373,7 +373,7 @@ export class AuthInfo extends AsyncOptionalCreatable<AuthInfo.Options> {
           isExpired:
             Boolean(devHubUsername) && expirationDate
               ? new Date(ensureString(expirationDate)).getTime() < new Date().getTime()
-              : false,
+              : undefined,
         });
       } catch (err) {
         final.push({
--- a/src/org/authInfo.ts
+++ b/src/org/authInfo.ts
@@ -94,9 +94,11 @@ export type OrgAuthorization = {
   aliases: Nullable<string[]>;
   configs: Nullable<string[]>;
   isScratchOrg?: boolean;
+  isDevHub?: boolean;
   instanceUrl?: string;
   accessToken?: string;
   error?: string;
+  isExpired: boolean;
 };
 
 /**
@@ -292,6 +294,10 @@ function base64UrlEscape(base64Encoded: string): string {
  * const connection = await Connection.create({ authInfo });
  * ```
  */
+
+const defaultOrgAuthFilter = (orgAuth: OrgAuthorization): boolean => !!orgAuth;
+export type OrgAuthorizationFilter = boolean | typeof defaultOrgAuthFilter;
+
 export class AuthInfo extends AsyncOptionalCreatable<AuthInfo.Options> {
   // Possibly overridden in create
   private usingAccessToken = false;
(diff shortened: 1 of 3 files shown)


src/org/authInfo.ts
```diff
@@ -326,16 +332,25 @@ export class AuthInfo extends AsyncOptionalCreatable<AuthInfo.Options> {
 
   /**
    * Get a list of all authorizations based on auth files stored in the global directory.
+   * One can supply a filter (see @param orgAuthFilter) and calling this function withoyt
```
mdonnalley, line 335: withoyt => without

src/org/authInfo.ts
```diff
@@ -344,32 +359,50 @@ export class AuthInfo extends AsyncOptionalCreatable<AuthInfo.Options> {
...
         const authInfo = await AuthInfo.create({ username });
-        const { orgId, instanceUrl, devHubUsername } = authInfo.getFields();
+        const { orgId, instanceUrl, devHubUsername, expirationDate, isDevHub } = authInfo.getFields();
         final.push({
           aliases,
           configs,
           username,
           instanceUrl,
-          isScratchOrg: Boolean(devHubUsername),
+          isScratchOrg: Boolean(devHubUsername) ?? false,
```
mdonnalley, line 368: I don't think the `?? false` is necessary here since `Boolean` will always return a boolean

```javascript
> Boolean('string')
true
> Boolean('')
false
> Boolean(null)
false
> Boolean(undefined)
false
> Boolean(NaN)
false
> Boolean([])
true
> Boolean({})
true
```

src/org/authInfo.ts
```diff
@@ -344,32 +359,50 @@ export class AuthInfo extends AsyncOptionalCreatable<AuthInfo.Options> {
...
           configs,
           username,
-          orgId: auth.orgId,
-          instanceUrl: auth.instanceUrl,
+          orgId: org.orgId,
+          instanceUrl: org.instanceUrl,
           accessToken: undefined,
           oauthMethod: 'unknown',
           error: err.message,
+          isExpired: false,
```
mdonnalley, line 388: Is it accurate to say that these orgs aren't expired? Maybe `isExpired` should be `boolean | 'unknown'`?

mdonnalley (changes requested):

src/org/authInfo.ts
peternhale: Since the only orgs where is expired is well-known is scratch orgs, I thought that defaulting to false was simplest thing to do, but I can see unknown as a alternative to a default of false.

src/org/authInfo.ts
```diff
@@ -98,7 +98,7 @@ export type OrgAuthorization = {
   instanceUrl?: string;
   accessToken?: string;
   error?: string;
-  isExpired: boolean;
+  isExpired: boolean | unknown;
```
mdonnalley, line 101: In my previous comment I meant that `isExpired` should be the literal string `'unknown'` similar to the `oauthMethod` property on line 93

src/org/authInfo.ts
```diff
@@ -94,9 +94,11 @@ export type OrgAuthorization = {
   aliases: Nullable<string[]>;
   configs: Nullable<string[]>;
   isScratchOrg?: boolean;
+  isDevHub?: boolean;
   instanceUrl?: string;
   accessToken?: string;
   error?: string;
+  isExpired: boolean | 'unknown';
```
peternhale, line 101: @mdonnalley I think this is what you are looking for.

src/org/authInfo.ts
```diff
@@ -326,16 +332,25 @@ export class AuthInfo extends AsyncOptionalCreatable<AuthInfo.Options> {
 
   /**
    * Get a list of all authorizations based on auth files stored in the global directory.
+   * One can supply a filter (see @param orgAuthFilter) and calling this function without
+   * a filter will return all authorizations.
+   *
+   * @param orgAuthFilter a boolean value or a predicate function that filters authorizations
+   *        For boolean value of true, return active org authorizations, false returns inactive
+   *        org authorizations. For a predicate function as (orgAuthorization: {OrgAuthorization}): boolean return
+   *        true for those org authorizations that are to be retained.
```
RodEsp, line 341: This is an extremely complex parameter that radically changes what this function returns in completely unintuitive ways. The logic around this parameter in `listAllAuthorizations()` is not easy to read or follow.

What is the reasoning behind adding it?
For W-9670267 or W-9670141 just having the `isExpired` property be returned as part of the authorization or org should be more than enough.

peternhale: How many times do we call listAllAuthorizations and then run some sot of filter across the collection.? I thought it would be useful to allow the dev to apply a filter as part of the call.

peternhale: Does not change the result returned and in practice it is not that hard to use.

peternhale: Here is how I plan on using this 
```
      const authorizations = await AuthInfo.listAllAuthorizations(
        (orgAuth) => !orgAuth.error && !orgAuth.isDevHub && !orgAuth.isExpired
      );
```

RodEsp: Why not just allow the user to pass in a function to do the filtering and then just do something like
```JavaScript
if (typeof orgAuthFilter === 'function') {
  final = final.filter(orgAuthFilter);
}

return final;
```

The logic used to account for both booleans and functions being passed in makes trying to understand what exactly will be returned and when kinda convoluted.

RodEsp (approved):

PR #460, opened 2021-08-09
sm/external-nuts-from-orb

mshanemc:
runs plugin nuts from sfdx-core branches @W-9330701@
I decided not to put this in the orb because fixing the core dependencies (deleting from command node module) got very specific to sfdx-core.

known gaps:
this will fix the plugin-config error: https://github.com/salesforcecli/plugin-config/pull/181

this will fix the plugin-auth compile error: https://github.com/salesforcecli/cli-plugins-testkit/pull/172

toolbelt is omitted becuase it would require ssh key setup since that's a private repo

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,10 +24,6 @@ jobs:
         description: "Path to sfdx executable to be used by NUTs, defaults to ''"
         default: ''
         type: string
-      dependency_to_replace:
-        description: 'The fully qualified npm module name, i.e. @salesforce/core, that will be replaced in the external NUTs.  Should match this project'
-        type: string
-        default: ''
       external_project_git_url:
         description: 'The url that will be cloned.  This contains the NUTs you want to run.  Ex: https://github.com/salesforcecli/plugin-user'
         type: string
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -2,6 +2,103 @@ version: 2.1
...
+        type: string
+      sfdx_executable_path:
+        description: "Path to sfdx executable to be used by NUTs, defaults to ''"
+        default: ''
+        type: string
+      external_project_git_url:
+        description: 'The url that will be cloned.  This contains the NUTs you want to run.  Ex: https://github.com/salesforcecli/plugin-user'
+        type: string
+        default: ''
+      # repo_tag:
```
peternhale, line 31: Remove if no longer needed

.circleci/config.yml
```diff
@@ -19,12 +116,34 @@ workflows:
                 - latest
                 - lts
                 - maintenance
+      - external-nut:
+          requires:
+            - release-management/test-package
+          sfdx_version: latest
+          matrix:
+            parameters:
+              os: [linux]
```
peternhale, line 125: @mshanemc is Windows going to be added once external-nuts is stable on Linux?

peternhale (changes requested):

.circleci/config.yml
mshanemc: I'd like to do that separately.  We have a ticket open with Circle to get newer windows images because
1) they're node 12 by default
2) we've run into several weird errors installing other versions of node

peternhale (approved):

shetzel (approved):

PR #461, opened 2021-08-10
feat: config interoperability

mdonnalley:
Supports complete interoperability of sfdx and sf configs

@W-9735962@

Closes #465

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -145,7 +145,7 @@ export const SFDX_ALLOWED_PROPERTIES = [
       // If a value is provided validate it otherwise no value is unset.
       validator: (value: ConfigValue) => {
         if (value == null) return true;
-        // validate if the value is a stirng and is a valid url and is either a salesforce domain
+        // validate if the value is a string and is a valid url and is either a salesforce domain
         // or an internal url.
         return (
           isString(value) &&
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -143,8 +143,16 @@ export const SFDX_ALLOWED_PROPERTIES = [
     description: '',
     input: {
       // If a value is provided validate it otherwise no value is unset.
-      validator: (value: ConfigValue) =>
-        value == null || (isString(value) && SfdcUrl.isValidUrl(value) && new SfdcUrl(value).isSalesforceDomain()),
+      validator: (value: ConfigValue) => {
+        if (value == null) return true;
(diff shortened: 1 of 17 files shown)


peternhale (approved):

peternhale (approved):

src/config/config.ts
```diff
@@ -130,7 +143,16 @@ export const SFDX_ALLOWED_PROPERTIES = [
     description: '',
     input: {
       // If a value is provided validate it otherwise no value is unset.
-      validator: (value: ConfigValue) => value == null || (isString(value) && new SfdcUrl(value).isSalesforceDomain()),
+      validator: (value: ConfigValue) => {
+        if (value == null) return true;
+        // validate if the value is a stirng and is a valid url and is either a salesforce domain
```
RodEsp, line 148:
```suggestion
        // validate if the value is a string and is a valid url and is either a salesforce domain
```

src/util/sfdcUrl.ts
```diff
@@ -28,6 +28,15 @@ export class SfdcUrl extends URL {
...
 
+  public static isValidUrl(input: string | URL): boolean {
+    try {
+      new URL(input.toString());
+      return true;
+    } catch {
+      return false;
+    }
+  }
+
```
RodEsp, line 39: Is this method actually necessary?
SfdcUrl extends URL and [its constructor just calls `super(input.toString(), base)`](https://github.com/forcedotcom/sfdx-core/pull/461/files?w=1#diff-da0e75f131dacdb7d5ded55cd7cce7c477aabe742f203f73b40df22134d3ee2dR24) which will error out if you give it an invalid URL. Makes me think that using this [up here](https://github.com/forcedotcom/sfdx-core/pull/461/files?w=1#diff-615dacd7000fa06d7afef0f22b7b6d9267c6219fbfc9fa0315e51a5f5d963755R152) is redundant.

RodEsp (approved):

src/util/sfdcUrl.ts
mdonnalley: It's not absolutely necessary but it's helpful. The problem was that the validator needs to return a boolean but `new SfdcUrl('invalidUrl')` would throw an error. So I could either but a try/catch in the validator function or I could add a static method that would always return a boolean. My personal opinion is that the `isValidUrl` method is a cleaner solution

PR #465, opened 2021-08-11
chore: update references to defaultusername

RodEsp:
Just updating a few more references to `defaultusername`


RodEsp:
No need to change these for #461.

PR #471, opened 2021-08-31
@W-9790306@ Custom templates config

xyc:
See [Support Custom Templates in CLI](https://salesforce.quip.com/GA1rASFbBX30)

### DIFF
--- a/messages/config.json
+++ b/messages/config.json
@@ -3,6 +3,7 @@
   "InvalidConfigValue": "Invalid config value. %s",
   "InvalidInstanceUrl": "Specify a valid Salesforce instance URL",
   "InvalidApiVersion": "Specify a valid Salesforce API version, for example, 42.0",
+  "InvalidCustomTemplates": "Specify a valid Custom Templates repository URL or location",
   "InvalidIsvDebuggerSid": "Specify a valid Debugger SID",
   "InvalidIsvDebuggerUrl": "Specify a valid Debugger URL",
   "InvalidNumberConfigValue": "Specify a valid number, for example, 150000",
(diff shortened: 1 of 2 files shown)


messages/config.json
```diff
@@ -3,6 +3,7 @@
   "InvalidConfigValue": "Invalid config value. %s",
   "InvalidInstanceUrl": "Specify a valid Salesforce instance URL",
   "InvalidApiVersion": "Specify a valid Salesforce API version, for example, 42.0",
+  "InvalidCustomOrgMetadataTemplates": "Specify a valid Custom Org Metadata Templates repository URL or location",
```
jshackell-sfdc, line 6: As I understand it, this config value will simply point to a directory that contains the templates, right?  i.e. it's NOT going to automatically clone/download a repo?  If so, I suggest this wording:

"Specify a valid directory that contains the custom project templates for working with org metadata."

Okay? @mshanemc , I'd love your opinion on the wording.  For example, is it weird say "for working with org metadata" when there's currently only one type of templates?  This is part of the "thinking aheadd when the CLI works with all clouds", but is it confusing to mention it now?

jshackell-sfdc: Also, is there an equivalent environment variable?  If so, what is it? Thx!

xyc: > this config value will simply point to a directory that contains the templates, right? i.e. it's NOT going to automatically clone/download a repo
It's not going to automatically clone/download a repo when `config:set` command is run, but you can specify a repo URL as the config value and the client (CLI / VS Code) will download the repo when creating with the custom templates.

xyc: The env variable would be automatically applied like other env variables: `SFDX_CUSTOM_ORG_METADATA_TEMPLATES`

jshackell-sfdc: Gotcha.  So now I have just a small edit to that error message:

"Specify a valid repository URL or directory for the custom org metadata templates."

xyc: Thanks, updated the message!

mshanemc (approved):
approved.  Adding Mike to the approvals, too.

jshackell-sfdc (approved):

sbudhirajadoc (approved):

mdonnalley (approved):

PR #474, opened 2021-09-09
feat: add help section helper

peternhale:
@W-9791511@


src/util/helpSectionUtils.ts
```diff
@@ -0,0 +1,38 @@
...
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import { HelpSection } from '@oclif/core';
+import { ORG_CONFIG_ALLOWED_PROPERTIES, OrgConfigProperties } from '../org/orgConfigProperties';
+import { SFDX_ALLOWED_PROPERTIES, SfdxPropertyKeys } from '../config/config';
+import { EnvironmentVariable, SUPPORTED_ENV_VARS } from '../config/envVars';
+
+export const toHelpSection = (
```
peternhale, line 12: produces a help section from the provided array of org/sfdx/env var enums

mdonnalley (changes requested):
I think this is too specific to `sf` to belong in sfdx-core - is there a reason we wouldn't just put this in the custom help class?

peternhale:
> I think this is too specific to `sf` to belong in sfdx-core - is there a reason we wouldn't just put this in the custom help class?

@mdonnalley placing this in help class would require that each command reference that class. I could make this available in @salesforce/command.

mdonnalley:
> > I think this is too specific to `sf` to belong in sfdx-core - is there a reason we wouldn't just put this in the custom help class?
> 
> @mdonnalley placing this in help class would require that each command reference that class. I could make this available in @salesforce/command.

@peternhale yeah, I think that would make more sense

peternhale:
> > > I think this is too specific to `sf` to belong in sfdx-core - is there a reason we wouldn't just put this in the custom help class?
> > 
> > 
> > @mdonnalley placing this in help class would require that each command reference that class. I could make this available in @salesforce/command.
> 
> @peternhale yeah, I think that would make more sense

@mdonnalley moving this to @salesforce/command is complicated, since this code needs access to core v3 and command is dependent on v2. I can alias in v3, but seem pretty heavy at the moment.

package.json
```diff
@@ -39,6 +39,7 @@
     }
   },
   "dependencies": {
+    "@oclif/core": "^0.5.35",
```
mdonnalley, line 42: this should be removed

mdonnalley (changes requested):

mdonnalley (approved):

PR #478, opened 2021-09-17
fix: improve time to build auths for many orgs @W-9914839@

peternhale:
I have been looking at perf around keychain access and found that the keychain program is invoked many times during commands like org list/env list. Each org has a new auth info constructed, which requires the keychain be accessed to fetch the crypto key used for encrypt/decrypt. I found this inefficient and this PR proposes a solution for this.

My experiments were on my local box where I have 32 orgs. When running sf env list w/o this change I see times in the ~3 second range and with this change the time is reduced to ~1.2 seconds.

### DIFF
--- a/src/crypto/crypto.ts
+++ b/src/crypto/crypto.ts
@@ -41,6 +41,12 @@ interface CredType {
   password: string;
 }
 
+const makeSecureBuffer = (password: string | undefined): SecureBuffer<string> => {
+  const newSb = new SecureBuffer<string>();
+  newSb.consume(Buffer.from(ensure(password), 'utf8'));
+  return newSb;
+};
+
 /**
  * osxKeyChain promise wrapper.
  */
@@ -58,17 +64,13 @@ const keychainPromises = {
       return new Promise((resolve, reject): {} => {
         return _keychain.getPassword({ service, account }, (err: Nullable<Error>, password?: string) => {
           if (err) return reject(err);
-          const newSb = new SecureBuffer<string>();
-          newSb.consume(Buffer.from(ensure(password), 'utf8'));
-          Cache.set(`${service}:${account}`, newSb);
+          Cache.set(`${service}:${account}`, makeSecureBuffer(password));
           return resolve({ username: account, password: ensure(password) });
         });
       });
     } else {
       const pw = sb.value((buffer) => buffer.toString('utf8'));
-      const newSb = new SecureBuffer<string>();
-      newSb.consume(Buffer.from(ensure(pw), 'utf8'));
-      Cache.set(`${service}:${account}`, newSb);
+      Cache.set(`${service}:${account}`, makeSecureBuffer(pw));
       return new Promise((resolve): void => {
         return resolve({ username: account, password: ensure(pw) });
       });
--- a/src/crypto/crypto.ts
+++ b/src/crypto/crypto.ts
@@ -13,6 +13,7 @@ import { ensure, Nullable, Optional } from '@salesforce/ts-types';
 import { AsyncOptionalCreatable, env } from '@salesforce/kit';
 import { Logger } from '../logger';
 import { Messages } from '../messages';
+import { Cache } from '../util/cache';
 import { retrieveKeychain } from './keyChain';
 import { KeyChain } from './keyChainImpl';
 import { SecureBuffer } from './secureBuffer';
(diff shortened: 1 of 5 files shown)


src/crypto/crypto.ts
```diff
@@ -52,12 +53,26 @@ const keychainPromises = {
...
-        return resolve({ username: account, password: ensure(password) });
-      })
-    );
+    const sb = Cache.get<SecureBuffer<string>>(`${service}:${account}`);
+    if (!sb) {
+      return new Promise((resolve, reject): {} => {
+        return _keychain.getPassword({ service, account }, (err: Nullable<Error>, password?: string) => {
+          if (err) return reject(err);
+          const newSb = new SecureBuffer<string>();
+          newSb.consume(Buffer.from(ensure(password), 'utf8'));
```
peternhale, line 62: At this moment, password has been properly fetch from the SecureBuffer used as part of the keychain access. The password will need to be stored in the cache in a SecureBuffer, so create SecureBuffer and have buffer consume the password, then store it in cache.

src/crypto/crypto.ts
```diff
@@ -52,12 +53,26 @@ const keychainPromises = {
...
+          Cache.set(`${service}:${account}`, newSb);
+          return resolve({ username: account, password: ensure(password) });
+        });
+      });
+    } else {
+      const pw = sb.value((buffer) => buffer.toString('utf8'));
```
peternhale, line 68: Extract password from SecureBuffer to pass back to caller
Store the password in cache in a new SecureBuffer

src/crypto/crypto.ts
```diff
@@ -52,12 +53,26 @@ const keychainPromises = {
    * @param account The keychain account name.
    */
   getPassword(_keychain: KeyChain, service: string, account: string): Promise<CredType> {
```
mshanemc, line 55: "password"...would keyChainEntry or keychainItem be a better description of what we're actually getting here?  This is getting back the key we use for decrypting stuff?

src/util/cache.ts
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+export class Cache extends Map {
+  private static _instance: Cache;
+  private _hits = 0;
```
mshanemc, line 9: wanna use the new stuff? `#hits = 0`

src/util/cache.ts
```diff
@@ -0,0 +1,49 @@
...
+  private _enabled = true;
+
+  public static instance(): Cache {
+    if (!Cache._instance) {
+      Cache._instance = new Cache();
+    }
+    return Cache._instance;
+  }
+
+  public static set<V>(key: string, value: V) {
```
mshanemc, line 20: what's the `V` for?

cache.set<string>('key', 4) // compile error?

src/crypto/crypto.ts
```diff
@@ -13,6 +13,7 @@ import { ensure, Nullable, Optional } from '@salesforce/ts-types';
 import { AsyncOptionalCreatable, env } from '@salesforce/kit';
 import { Logger } from '../logger';
 import { Messages } from '../messages';
+import { Cache } from '../util/cache';
```
mshanemc, line 16: checking my understanding: By putting the cache inside Crypto, each consumer of Crypto (typically configfile or one of its descendants), the cache is scoped to that instance of the configFile(etc) class?

mdonnalley (approved):

src/util/cache.ts
RodEsp: Yeah, totally should. Typescript's `private` only treats methods as private during compile time, using `#` will do it both at compile and runtime.
https://www.amitmerchant.com/typescript-private-modifier-vs-ecmascript-hash-private-fields/

src/util/cache.ts
peternhale: @mshanemc `V` is type of the value. So in your example, `cache.set('key', 4)` will compile properly. I believe TS will infer the type from the type of the value provided in the set call.

src/crypto/crypto.ts
peternhale: @mshanemc No Cache is a singleton, so same instance is used for all references across the app.

RodEsp (approved):

src/util/cache.ts
mshanemc: sorry, type-o.  I meant to ask about the `<T>` outside of the method, saying that

`cache.set<'string'>('key', 4)` would fail.  Trying to figure out what the purpose of that is

mshanemc (approved):
approved--my questions are non-blocking 
that #private stuff is so pretty.

PR #480, opened 2021-09-18
fix: edit the config var blurbs

jshackell-sfdc:
@peternhale , I left the two isvXX config vars as-is.  I couldn't find any more info about them.  Can we hide them from the --help output for now?  Thx!

### DIFF
--- a/messages/config.md
+++ b/messages/config.md
@@ -73,19 +73,19 @@ The sfdx-project.json file must include one, and only one, default package direc
 
 # targetOrg
 
-The target to be used for any command communicating with an org.
+Username or alias of the org that all commands run against by default. (sf only)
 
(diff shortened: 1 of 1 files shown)


peternhale (approved):

jshackell-sfdc:
@mdonnalley , I need a second reviewer :). Can you merge if you approve? Thx!

jshackell-sfdc:
Another thing: can someone remove the isvDebuggerSid and isvDebuggerURL config vars from being printed in the --help?  We never figured out what these config vars are, so I'd prefer not to include them.  I couldn't find any official docs on the equivalent env vars.  FYI @peternhale , @mdonnalley  Thx!

PR #483, opened 2021-10-04
Migrate scratchOrgCreateCommand from toolbelt

maggiben:
### What does this PR do?
Migrate `scratchOrgCreateCommand` from toolbelt
@W-9805905@


src/scratchOrgInfoApi.ts
```diff
@@ -0,0 +1,415 @@
...
+
+/**
+ * This retrieves the ScratchOrgInfo, polling until the status is Active or Error
+ *
+ * @param hubOrg
+ * @param scratchOrgInfoId - the id of the scratchOrgInfo that we are retrieving
+ * @param timeout - A Duration object
+ * @returns {BBPromise}
+ */
+export const pollForScratchOrgInfo = async (
```
peternhale, line 310: @mshanemc Any plans on improving user feedback during org create? Right now there is absolutely no feedback during org create, which is frustrating to the user and to us when trying to diagnose org create issues. It sure would be nice if org create had better feedback.

src/scratchOrgInfoGenerator.ts
```diff
@@ -0,0 +1,186 @@
...
+    }
+    scratchOrgInfoPayload.features = scratchOrgInfoPayload.features.map((feature: string) => feature.trim());
+
+    const scratchOrgFeatureDeprecation = new ScratchOrgFeatureDeprecation();
+
+    scratchOrgInfoPayload.features = scratchOrgFeatureDeprecation.filterDeprecatedFeatures(
+      scratchOrgInfoPayload.features
+    );
+
+    scratchOrgInfoPayload.features = scratchOrgInfoPayload.features.join(';');
```
peternhale, line 170:
``` suggestion
    scratchOrgInfoPayload.features = scratchOrgFeatureDeprecation.filterDeprecatedFeatures(
      scratchOrgInfoPayload.features
    ).join(';');
```

src/scratchOrgSettingsGenerator.ts
```diff
@@ -0,0 +1,421 @@
...
+    if (parseFloat(apiVersion) >= 47.0 && this.orgPreferenceSettingsMigrationRequired(scratchDef)) {
+      this.extractAndMigrateSettings(scratchDef);
+    } else {
+      this.settingData = getObject(scratchDef, 'settings');
+      this.objectSettingsData = getObject(scratchDef, 'objectSettings');
+    }
+
+    this.logger.debug('settings are', this.settingData);
+    // TODO, this is where we will validate the settings.
+    // See W-5068155
```
peternhale, line 94: @mshanemc The work item referenced here has been 'nevered'. Is that something you agree with? If so, then these comments can be removed.

src/scratchOrgSettingsGenerator.ts
```diff
@@ -0,0 +1,421 @@
...
+    }
+
+    if (!apiVersion) {
+      apiVersion = this.currentApiVersion;
+    }
+
+    // First, let's map the old style tooling preferences into MD-API preferences
+    this.settingData = {};
+
+    function lhccmdt(mdt: string): string {
```
peternhale, line 129: Would prefer the use of npm module [change-case](https://www.npmjs.com/package/change-case) for these types of operations.

peternhale (changes requested):

src/scratchOrgInfoApi.ts
```diff
@@ -0,0 +1,415 @@
...
+    delay,
+    retries,
+  });
+
+  const authInfo = retries
+    ? await retryAuthorize({
+        username: scratchOrgInfoComplete.SignupUsername,
+        parentUsername: hubOrg.getUsername(),
+        oauth2Options,
+      }).catch((reason) => {
```
WillieRuemmele, line 192: we try to use `try{} catch{}` instead of `.catch`

test/unit/util/mapKeysTest.ts
```diff
@@ -0,0 +1,174 @@
+/*
+ * Copyright (c) 2021, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { expect } from 'chai';
```
WillieRuemmele, line 8: Are there no other unit tests for the other scratch org creation logic? There's a lot of untested code

WillieRuemmele (changes requested):
I started this before Pete finished his, so we have some overlapping comments

src/scratchOrgCreateCommand.ts
```diff
@@ -0,0 +1,300 @@
...
+  /**
+   * executes the command. this is a protocol style function intended to be represented by other commands.
+   *
+   * @param cliContext - the cli context
+   * @param stdinValues - param values obtained from stdin
+   * @returns {Promise}
+   */
+  public async execute(
+    clientSecret: string
+  ): Promise<{ orgId: Optional<string>; username: Optional<string>; warnings: Optional<string[]> }> {
```
mshanemc, line 129: this is our 1 chance to redefine the return structure--imagine all the things that you could get back out of this.

For example, the settings deployment produces and SDR deployResult.

If you had that, the command could show success/failure of what it deployed, or didn't and why!
Wanna show what API was used to deploy those settings, as a sanity check?

What would a timeout want to return?  At least a page you could open to check the ScratchOrgInfo to see it's final outcome or delete it?

src/scratchOrgFeatureDeprecation.ts
```diff
@@ -0,0 +1,123 @@
...
+  ],
+};
+
+interface FeatureTypes {
+  simpleFeatureMapping: { [key: string]: string[] };
+  quantifiedFeatureMapping: Record<string, string>;
+  deprecatedFeatures: string[];
+}
+
+export class ScratchOrgFeatureDeprecation {
```
mshanemc, line 47: this class is going to be pretty easy to test (pass various featureTypes, expect values and warnings).

I'd like to see some other very testable functions that could, for example,

1. build the scratchOrgInfo request (give inputs, get object that you'd send OR errors/warnings)
2. build the scratchOrg settings (give inputs, get deployable settings OR errors/warnings)

Then you can avoid having to mock all the server stuff since most of the complexity lies in all the possible inputs, and the nuts can handle the e2e testing.

changing this stuff to a more input/output approach can simplify the UT work

src/scratchOrgInfoApi.ts
```diff
@@ -0,0 +1,415 @@
...
+import mapKeys from './util/mapKeys';
+import { AuthInfo } from './authInfo';
+import { Messages } from './messages';
+import { SfdxError } from './sfdxError';
+import { SfdcUrl } from './util/sfdcUrl';
+import { MyDomainResolver } from './status/myDomainResolver';
+
+import SettingsGenerator, { ScratchDefinition } from './scratchOrgSettingsGenerator';
+
+export interface ScratchOrgInfo {
```
mshanemc, line 32: There's really 2 flavors of this type: a request (some things won't exist, others are required, some can't be set) and what exists on the server after creation (more things are guaranteed to exist, all are readonly) that's in the response.

Think of the other code that's using it (org:display|list, org:delete) that could benefit from having pretty fantastic typings in sfdx-core.

src/scratchOrgSettingsGenerator.ts
```diff
@@ -0,0 +1,421 @@
...
+    return destRoot;
+  }
+
+  /**
+   * Deploys the settings to the org.
+   */
+  public async deploySettingsViaFolder(username: string, sourceFolder: string): Promise<FileResponse[]> {
+    this.logger.debug(`deploying settings from ${sourceFolder}`);
+
+    const componentSet = ComponentSet.fromSource(sourceFolder);
```
mshanemc, line 284: of this entire PR, this is the one that's really causing me to think about options.  We're about to create a circular dependency SDR <=> sfdx-core.

It's not a bad one (SDR doesn't use this part of core, so it's not like 2 classes pointed at each other) but it might be a eventual operations pain (think version bumps and ts conflicts).  I'd love some others' thoughts about this.

The alternative would be returning a componentSet to the command(s), letting the command call deploy, and then having them maintain an instance or pass back the deploy results into sfdx-core.  All of which sounds nasty.

I think this might be the best option for sharing code with `sf`.

src/scratchOrgSettingsGenerator.ts
```diff
@@ -0,0 +1,421 @@
...
+    const deploy = await componentSet.deploy({ usernameOrConnection: username });
+
+    // Wait for polling to finish and get the DeployResult object
+    const result = await deploy.pollStatus();
+    // display errors if any
+    const errors = result.getFileResponses().filter((fileResponse) => fileResponse.state === ComponentStatus.Failed);
+
+    if (result.response.status === 'Failed') {
```
mshanemc, line 292: corresponding to that top-level comment about output...this probably returns the SDR result.

mshanemc (changes requested):

test/unit/util/mapKeysTest.ts
maggiben: Yes, that's true, tons of untested code in fact, there are portions that can be `ported` from toolbelt but my best guess is that 80% of them will have to be re-written, I created this W-9991956 to tackle this problem, though I will probably put all changes here in this PR

src/scratchOrgInfoApi.ts
peternhale: I was thinking of letting the user know how the org create is progressing by using scratch org info status as a source and possibly combine that set with the post signup steps, auth and lightning domain verification.

src/scratchOrgCreateCommand.ts
```diff
@@ -0,0 +1,271 @@
...
+  private logger!: Logger;
+  private settingsGenerator: SettingsGenerator;
+
+  public constructor(private options: ScratchOrgCreateCommand.Options) {
+    super(options);
+    this.settingsGenerator = new SettingsGenerator();
+  }
+
+  /**
+   * executes the command. this is a protocol style function intended to be represented by other commands.
```
shetzel, line 108: This comment doesn't apply

src/scratchOrgCreateCommand.ts
```diff
@@ -0,0 +1,271 @@
...
+   *
+   * @param cliContext - the cli context
+   * @param stdinValues - param values obtained from stdin
+   * @returns {Promise}
+   */
+  protected async init(): Promise<void> {
+    this.logger = await Logger.child('scratchOrgCreateCommand');
+    this.logger.debug('scratchOrgCreateCommand: execute');
```
shetzel, line 116: The logger should have a different name and the debug output is wrong.

src/scratchOrgCreateCommand.ts
```diff
@@ -0,0 +1,271 @@
...
+        .update(queryResult.map((record) => ({ Id: record.Id, RevisionCounter: 0 })));
+    } catch (err) {
+      const message = messages.getMessage('SourceStatusResetFailure', [
+        this.scratchOrg.getOrgId(),
+        this.scratchOrg.getUsername(),
+      ]);
+      throw new SfdxError(message, 'SourceStatusResetFailure');
+    }
+  }
+}
```
shetzel, line 257: We shouldn't do this as part of scratch org creation.  This is for source tracking and the caller may never want/use it.  Source tracking initialization should handle this.

shetzel (changes requested):

src/scratchOrgFeatureDeprecation.ts
```diff
@@ -0,0 +1,123 @@
...
+      });
+      /* If a simply mapped feature is defined, log a warning.*/
+      Object.keys(this.featureTypes.simpleFeatureMapping).forEach((key) => {
+        if (new RegExp(`${key};|${key},|${key}$`, 'i').test(requestedFeatures)) {
+          const tokens = '[' + this.featureTypes.simpleFeatureMapping[key].map((v) => "'" + v + "'").join(',') + ']';
+          featureWarningMessages.push(messages.getMessage('mappedFeatureWarning', [key, tokens]));
+        }
+      });
+      /* If a deprecated feature is identified as deprecated, throw a warning.*/
+      this.featureTypes.deprecatedFeatures.forEach((deprecatedFeature) => {
```
mshanemc, line 94: that looks like a `filter`/`map`
`featureWarningMessages.push(
    ...this.featureTypes.deprecatedFeatures
      .filter(deprecated => requestedFeatures.includes(deprecated))
      .map(deprecated => messages.getMessage('deprecatedFeatureWarning', [deprecatedFeature]))
)`

src/scratchOrgInfoGenerator.ts
```diff
@@ -0,0 +1,298 @@
...
+
+  if (options.connectedAppConsumerKey) {
+    scratchOrgInfoPayload.connectedAppConsumerKey = options.connectedAppConsumerKey;
+  }
+
+  scratchOrgInfoPayload.durationDays = options.durationDays;
+
+  // Throw warnings for deprecated scratch org features.
+  const scratchOrgFeatureDeprecation = new ScratchOrgFeatureDeprecation();
+  scratchOrgFeatureDeprecation.getFeatureWarnings(scratchOrgInfoPayload.features).forEach((warning) => {
```
mshanemc, line 276: I think this belongs after the `if (typeof scratchOrgInfoPayload.features === 'string') {` code (which ensures the features are an array).

That would simplify what has to happen in `getFeatureWarnings` (it could always expect an array)

src/scratchOrgFeatureDeprecation.ts
```diff
@@ -0,0 +1,123 @@
...
+              key,
+              this.featureTypes.quantifiedFeatureMapping[key],
+            ])
+          );
+        }
+      });
+      /* If a simply mapped feature is defined, log a warning.*/
+      Object.keys(this.featureTypes.simpleFeatureMapping).forEach((key) => {
+        if (new RegExp(`${key};|${key},|${key}$`, 'i').test(requestedFeatures)) {
+          const tokens = '[' + this.featureTypes.simpleFeatureMapping[key].map((v) => "'" + v + "'").join(',') + ']';
```
mshanemc, line 89: yeah, this part would be a lot simpler if the function were working with a `string[]` instead of turning it into a giant string

src/scratchOrgFeatureDeprecation.ts
```diff
@@ -0,0 +1,123 @@
...
+    return featureWarningMessages;
+  }
+
+  /**
+   * Removes all deprecated features for the organization.
+   *
+   * @param features List of features to filter
+   * @returns feature array with proper mapping.
+   */
+  public filterDeprecatedFeatures(features: string[]): string[] {
```
mshanemc, line 109: this would be simpler as a filter (to remove deprecated) and map (to replace the `simpleFeatureMapping`) than using `.reduce`

src/scratchOrgCreate.ts
```diff
@@ -0,0 +1,171 @@
...
+
+Messages.importMessagesDirectory(__dirname);
+const messages: Messages = Messages.loadMessages('@salesforce/core', 'scratchOrgCreate');
+
+export interface ScratchOrgCreateResult {
+  username: Optional<string>;
+  authInfo: Optional<AuthInfo>;
+  warnings: string[];
+}
+export interface ScratchOrgCreateOptions {
```
mshanemc, line 39: so they're all required?

src/scratchOrgCreate.ts
```diff
@@ -0,0 +1,171 @@
...
+  configAggregator: ConfigAggregator;
+  test: boolean;
+  clientSecret: string;
+}
+
+export const scratchOrgCreate = async (options: ScratchOrgCreateOptions): Promise<ScratchOrgCreateResult> => {
+  const logger = await Logger.child('scratchOrgCreate');
+  logger.debug('scratchOrgCreate');
+
+  const {
```
mshanemc, line 62: and none of them have any defaults?

src/scratchOrgCreate.ts
```diff
@@ -0,0 +1,171 @@
...
+    settingsGenerator,
+    scratchOrg
+  );
+
+  logger.trace('Settings deployed to org');
+  /** updating the revision num to zero during org:creation if source members are created during org:create.This only happens for some specific scratch org definition file.*/
+  await updateRevisionCounterToZero(scratchOrg);
+
+  return {
+    username,
```
mshanemc, line 144: is there anything else a command/consumer might want?  ex: orgId, scratchOrgInfo id/link for the dev hub record, the mdapi output of what got deployed to the org?

mshanemc (changes requested):

src/scratchOrgInfoApi.ts
```diff
@@ -0,0 +1,495 @@
...
+
+Messages.importMessagesDirectory(__dirname);
+const messages = Messages.loadMessages('@salesforce/core', 'scratchOrgInfoApi');
+
+const checkScratchOrgInfoForErrors = (orgInfo: ScratchOrgInfo, hubUsername: Optional<string>, logger: Logger) => {
+  if (orgInfo.Status === 'Active') {
+    return orgInfo;
+  }
+  if (orgInfo.Status === 'Error' && orgInfo.ErrorCode) {
+    throw new SfdxError(messages.getMessage('signupFailed', [orgInfo.ErrorCode]), 'RemoteOrgSignupFailed');
```
mshanemc, line 79: this should have all the error codes.  See the current version in toolbelt.

They're missing from the messages.json on this PR

src/scratchOrgCreate.ts
```diff
@@ -0,0 +1,171 @@
...
+  wait: Duration;
+  setdefaultusername: boolean;
+  setalias: string;
+  retry: number;
+  apiversion: string;
+  definitionjson: string;
+  definitionfile: string;
+  orgConfig: Record<string, unknown>;
+  configAggregator: ConfigAggregator;
+  test: boolean;
```
mshanemc, line 54: what does this do?

src/scratchOrgCreate.ts
```diff
@@ -0,0 +1,180 @@
...
+
+const DEFAULT_STREAM_TIMEOUT_MINUTES = 6;
+
+export interface ScratchOrgCreateResult {
+  username?: string;
+  authInfo?: AuthInfo;
+  warnings: string[];
+}
+export interface ScratchOrgCreateOptions {
+  hubOrg: Org;
```
mshanemc, line 42: I think you should jsdoc this interface.  

Examples--what does retry do?  
What happens if you don't specify an apiVersion?
Should I pass in a `definitionjson` or a `definitionfile` or both ?  What happens if I do both?
what's an orgConfig?  I wouldn't know from the type itself (Record<string, unknown>) what to put in there.  

Look at this through the lens of someone who wants to create scratch orgs from a node process outside of our commands, who doesn't know our flags.  How would they figure out what to send?

src/scratchOrgCreate.ts
```diff
@@ -0,0 +1,180 @@
...
+    setalias,
+    retry = 0,
+    apiversion,
+    definitionjson,
+    definitionfile,
+    orgConfig,
+    configAggregator,
+    clientSecret,
+  } = options;
+
```
mshanemc, line 80: I think we need to validate the inputs.  The command flags does a lot for us (ex: min/max on days) but this API is going to accept the inputs and won't have errors until they fail on the server.

src/scratchOrgInfoApi.ts
mshanemc: I mean these: https://github.com/salesforcecli/toolbelt/blob/main/messages/scratchorg_error_codes.json

this was fixed here after you started this work because Judy noticed the error codes were missing.  https://github.com/salesforcecli/toolbelt/blob/main/src/lib/org/scratchOrgErrorCodes.ts

mshanemc (changes requested):
big items
1. missing messages
2. input validation
3. error details for deploy fails
4. response to or address the remaining comments.

small things I still want done
1. jsdoc fixes and additions
2. non-required properties and defaults on the input

src/scratchOrgCreate.ts
```diff
@@ -0,0 +1,239 @@
...
+const validateRetry = (retry: number) => {
+  const max = 10;
+  if (Number.isInteger(retry)) {
+    if (retry > max) {
+      throw new SfdxError(`Expected 'retry' less than or equal to ${max} but received ${retry}`, 'BoundsError');
+    }
+    return;
+  }
+  throw new SfdxError("Expected 'durationDays' to be an integer number", 'TypeError');
```
WillieRuemmele, line 107: this error mentions `durationDays` for `validateRetry`

src/scratchOrgInfoApi.ts
```diff
@@ -117,9 +117,19 @@ const getOrgInstanceAuthority = function (
...
+ * @param clientSecret The OAuth client secret. May be null for JWT OAuth flow.
+ * @param scratchOrgInfoComplete The completed ScratchOrgInfo which should contain an access token.
+ * @param retry auth retry attempts
+ * @param signupTargetLoginUrlConfig the login url
+ * @returns {OAuth2Options, number, number, number} options, retries, timeout, delay
+ */
 const buildOAuth2Options = async (options: {
   hubOrg: Org;
-  clientSecret: string;
+  clientSecret?: string;
```
mshanemc, line 132: style suggestion: required properties first, optional properties last.

src/scratchOrgSettingsGenerator.ts
```diff
@@ -130,18 +130,19 @@ export default class SettingsGenerator {
     logger.debug(`settings deployment status ${result.status}`);
 
     if (result.status !== RequestStatus.Succeeded) {
-      throw new SfdxError(
+      const error = new SfdxError(
         `A scratch org was created with username ${username}, but the settings failed to deploy`,
         'ProblemDeployingSettings'
       );
+      error.setData(result);
```
mshanemc, line 137: nice

src/scratchOrgCreate.ts
```diff
@@ -0,0 +1,171 @@
...
+  });
+
+  // gets the scratch org settings (will use in both signup paths AND to deploy the settings)
+  await settingsGenerator.extract(scratchOrgInfo);
+  logger.debug(`the scratch org def file has settings: ${settingsGenerator.hasSettings()}`);
+
+  // creates the scratch org info in the devhub
+  const scratchOrgInfoRequestResult = await requestScratchOrgCreation(hubOrg, scratchOrgInfo, settingsGenerator);
+
+  const scratchOrgInfoId = ensureString(getString(scratchOrgInfoRequestResult, 'id'));
```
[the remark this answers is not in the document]
mshanemc: suggestion: avoid getString and use the return typing

src/scratchOrgCreate.ts
```diff
@@ -31,54 +31,113 @@ import { generateScratchOrgInfo, getScratchOrgInfoPayload } from './scratchOrgIn
...
+        'BoundsError'
+      );
+    }
+    return;
+  }
+  throw new SfdxError("Expected 'durationDays' to be an integer number", 'TypeError');
+};
+
+const validateRetry = (retry: number) => {
+  const max = 10;
```
mshanemc, line 100: do we need a max?  The integer check is good.

src/scratchOrgCreate.ts
```diff
@@ -31,54 +31,113 @@ import { generateScratchOrgInfo, getScratchOrgInfoPayload } from './scratchOrgIn
...
   username?: string;
   authInfo?: AuthInfo;
   warnings: string[];
 }
+
+/**
+ * interface ScratchOrgCreateOptions
+ *
+ * @param hubOrg the environment hub org
+ * @param connectedAppConsumerKey The connected app consumer key. May be null for JWT OAuth flow.
```
mshanemc, line 46: jwt would require a consumer key, right?

src/scratchOrgInfoGenerator.ts
```diff
@@ -208,32 +208,32 @@ export const generateScratchOrgInfo = async ({
...
+ * @param orgConfig overrides definitionjson
  * @returns scratchOrgInfoPayload: ScratchOrgInfoPayload;
             ignoreAncestorIds: boolean;
             warnings: string[]; 
  */
 export const getScratchOrgInfoPayload = async (options: {
-  definitionjson: string;
-  definitionfile: string;
-  connectedAppConsumerKey: string;
+  definitionjson?: string;
```
mshanemc, line 223: suggestion: consider using the TS Pick to avoid having to repeat the types for all these properties, since this is really a subset of `ScratchOrgCreateOptions`

mshanemc (changes requested):
changes requested: 

willie's comment about the validation message being wrong
consumer key is required ( I think...)
I unresolved this comment because I think it's still missing: https://github.com/forcedotcom/sfdx-core/pull/483#discussion_r749445672


All good with everything else.

src/scratchOrgInfoApi.ts
```diff
@@ -0,0 +1,528 @@
...
+      return await retryAuthorize({
+        username: options.username,
+        parentUsername: options.hubOrg.getUsername(),
+        oauth2Options: options.oauth2Options,
+      });
+    } catch (err) {
+      const error = err as RetryError;
+      logger.error(error);
+      if (error.message.startsWith('Timeout after')) {
+        throw SfdxError.create('salesforce-alm', 'org_create', 'jwtAuthRetryTimedOut', [
```
shetzel, line 220: This is not correct.  It should be getting an appropriate message from this library.

src/scratchOrgInfoApi.ts
```diff
@@ -0,0 +1,528 @@
...
+  } else {
+    try {
+      return await AuthInfo.create({
+        username: options.username,
+        parentUsername: options.hubOrg.getUsername(),
+        oauth2Options: options.oauth2Options,
+      });
+    } catch (error) {
+      throw SfdxError.wrap(error as Error);
+    }
```
shetzel, line 237: `AuthInfo.create` should already throw SfdxErrors so I don't think you have to wrap and rethrow.

src/scratchOrgInfoApi.ts
```diff
@@ -0,0 +1,528 @@
...
+    logger.debug(`_authorize - setting alias to ${options.alias}`);
+    await options.authInfo.setAlias(options.alias);
+    logger.debug(`_authorize - AuthInfo has alias to ${options.authInfo.getFields().alias}`);
+  }
+  if (options.setAsDefault) {
+    await options.authInfo.setAsDefault({ defaultUsername: true });
+  }
+
+  logger.debug(`_authorize - orgConfig.loginUrl: ${options.authInfo.getFields().loginUrl}`);
+  logger.debug(`_authorize - orgConfig.instanceUrl: ${options.authInfo.getFields().instanceUrl}`);
```
shetzel, line 280: All of these debugs should not have "_authorize - " in them.  That is left over from toolbelt.

You probably only need 1 debug of the alias (line 273) as well.

src/scratchOrgInfoApi.ts
```diff
@@ -0,0 +1,528 @@
...
+  hubOrg: Org;
+  clientSecret?: string;
+  setAsDefault: boolean;
+  signupTargetLoginUrlConfig?: string;
+  alias?: string;
+  retry?: number;
+}): Promise<AuthInfo> => {
+  const { scratchOrgInfoComplete, hubOrg, clientSecret, setAsDefault, signupTargetLoginUrlConfig, alias } = options;
+  const logger = await Logger.child('authorizeScratchOrg');
+  logger.debug(`_authorize - scratchOrgInfoComplete: ${JSON.stringify(scratchOrgInfoComplete, null, 4)}`);
```
shetzel, line 306: can remove the "_authorize - " part of the debug

src/scratchOrgInfoApi.ts
```diff
@@ -0,0 +1,528 @@
...
+ */
+export const pollForScratchOrgInfo = async (
+  hubOrg: Org,
+  scratchOrgInfoId: string,
+  // org:create specifies a default timeout of 6.  This longer default is for other consumers
+  timeout: Duration = Duration.minutes(15)
+): Promise<ScratchOrgInfo> => {
+  const logger = await Logger.child('scratchOrgInfoApi-pollForScratchOrgInfo');
+  logger.debug(`PollingTimeout in minutes: ${timeout.minutes}`);
+  logger.debug(`pollForScratchOrgInfo this.scratchOrgInfoId: ${scratchOrgInfoId} from devHub ${hubOrg.getUsername()}`);
```
shetzel, line 435: "pollForScratchOrgInfo" is redundant in this debug.

src/scratchOrgCreate.ts
```diff
@@ -97,14 +97,10 @@ const validateDuration = (durationDays: number) => {
 };
 
 const validateRetry = (retry: number) => {
```
mshanemc, line 99: set return type void.

mshanemc (approved):
nice!

src/scratchOrgErrorCodes.ts
```diff
@@ -0,0 +1,62 @@
...
+import { Messages } from './messages';
+import { SfdxError } from './sfdxError';
+import { ScratchOrgInfo } from './scratchOrgInfoApi';
+
+const WORKSPACE_CONFIG_FILENAME = 'sfdx-project.json';
+
+Messages.importMessagesDirectory(__dirname);
+const messages: Messages = Messages.loadMessages('@salesforce/core', 'scratchOrgErrorCodes');
+
+// Messages.importMessagesDirectory(__dirname);
```
WillieRuemmele, line 19: this can be deleted L19-20

src/scratchOrgInfoApi.ts
```diff
@@ -0,0 +1,504 @@
...
+    throw new SfdxError(messages.getMessage('deprecatedPrefFormat'));
+  }
+
+  const scratchOrgInfo = mapKeys(scratchOrgRequest, upperFirst, true);
+
+  await checkOrgDoesntExist(scratchOrgInfo); // throw if it does exists.
+  try {
+    return await hubOrg.getConnection().sobject('ScratchOrgInfo').create(scratchOrgInfo);
+  } catch (error) {
+    const jsForceError = error as JsForceError;
```
WillieRuemmele, line 388: Can this be a `SfdxError` - I don't think we should be using `JsForceError` here, or if we need to, can you add a comment explaining why

src/scratchOrgInfoApi.ts
```diff
@@ -0,0 +1,504 @@
...
+    } catch (error) {
+      const sfdxError = SfdxError.wrap(error as Error);
+      logger.debug(`processScratchOrgInfoResult - err: ${JSON.stringify(error, null, 4)}`);
+      if (sfdxError.name === 'MyDomainResolverTimeoutError') {
+        sfdxError.setData({
+          orgId: scratchOrgAuthInfo.getFields().orgId,
+          username: scratchOrgAuthInfo.getFields().username,
+          instanceUrl: scratchOrgAuthInfo.getFields().instanceUrl,
+        });
+        logger.debug(`processScratchOrgInfoResult - err data: ${JSON.stringify(sfdxError.data, null, 4)}`);
```
WillieRuemmele, line 497: fyi: you can change this to be `'- err data: %s', sfdxError.data)` and it will handle formatting an object for you

WillieRuemmele (commented):
mostly smaller NITs/formatting

package.json
```diff
@@ -34,16 +34,20 @@
     "!lib/**/*.map"
   ],
   "dependencies": {
+    "@oclif/config": "^1.18.1",
```
mshanemc, line 37: why oclif config?

test/unit/scratchOrgCreateTest.ts
```diff
@@ -0,0 +1,124 @@
+/*
```
peternhale, line 1: I would have expected more tests for org create. Are there plans to increase unit testing for this?

peternhale (changes requested):

peternhale (approved):

package.json
```diff
@@ -35,15 +35,18 @@
   ],
   "dependencies": {
     "@salesforce/bunyan": "^2.0.0",
-    "@salesforce/kit": "^1.5.0",
+    "@salesforce/kit": "^1.5.17",
     "@salesforce/schemas": "^1.0.1",
-    "@salesforce/ts-types": "^1.5.13",
+    "@salesforce/source-deploy-retrieve": "^4.4.7",
```
shetzel, line 40: @maggiben - does the SDR dependency need to be here?

src/scratchOrgSettingsGenerator.ts
```diff
@@ -114,16 +113,13 @@ export default class SettingsGenerator {
     logger.debug(`deploying settings id ${id}`);
 
     let result = await connection.metadata.checkDeployStatus(id);
-    await Lifecycle.getInstance().emit('deploySettingsViaFolder', {
-      status: result.status,
-    });
 
     const timer = (timeout?: number) => {
```
mshanemc, line 117: I'd prefer that we consistently use the same library that's in PollingClient.

src/scratchOrgSettingsGenerator.ts
```diff
@@ -134,16 +130,15 @@ export default class SettingsGenerator {
       });
     };
 
-    const timeout = timer(10 * 60 * 1000); // timeout of 10 minutes
+    const { timeout, timerId } = timer(10 * 60 * 1000); // timeout of 10 minutes
 
     while (!breakPooling.includes(result.status) && !timeout()) {
```
mshanemc, line 135: breakPolling?

src/scratchOrgSettingsGenerator.ts
```diff
@@ -119,11 +118,30 @@ export default class SettingsGenerator {
...
-    while (!breakPooling.includes(result.status)) {
+    const timer = (timeout?: number) => {
+      let expired = false;
+      setTimeout(() => {
+        expired = true;
+      }, timeout);
+      return (): boolean => expired;
+    };
+
+    const sleep = async (delay?: number): Promise<void> => {
```
mshanemc, line 125: if you decide to do it this way, there's already a `sleep` in `kit` that you can use instead of writing another one.

src/scratchOrgSettingsGenerator.ts
```diff
@@ -114,34 +116,38 @@ export default class SettingsGenerator {
...
+        } catch (error) {
+          logger.debug(`An error occurred trying to check deploy id: ${id}`);
+          logger.debug(`Error: ${(error as Error).message}`);
+          logger.debug('Re-trying deploy check again....');
+          return {
+            completed: false,
+          };
+        }
+      },
+      timeout: Duration.seconds(10 * 60), // 10 minutes
```
mshanemc, line 142: `Duration.minutes(10)`

PR #488, opened 2021-10-11
fix: support for .mil sandboxes

mshanemc:
@W-10016942@ 

.mil sandboxes are yet another weird url structure we didn't know about.  The previous fix fixed their prod domains but not their sandboxes


maggiben (approved):

WillieRuemmele (approved):
let's hope (once again) this is all of the .mil

PR #491, opened 2021-11-08
Deletes Scratch org + Sandboxes

WillieRuemmele:
adds logic/tests to delete a scratch org from the DevHub and removes local auth files

adds logic/tests to delete a sandbox from a production org

@W-10144418@
@W-10144443@

### DIFF
--- a/messages/org.json
+++ b/messages/org.json
@@ -1,6 +1,7 @@
 {
   "NotFoundOnDevHub": "The scratch org does not belong to the dev hub username %s.",
   "NotADevHub": "The provided dev hub username %s is not a valid dev hub.",
+  "NoDevHubFound": "Unable to associate this scratch org with a DevHub",
   "deleteOrgHubError": "The Dev Hub org cannot be deleted.",
   "insufficientAccessToDelete": "You do not have the appropriate permissions to delete a scratch org. Please contact your Salesforce admin.",
   "attemptingToDeleteExpiredOrDeleted": "Attempting to delete an expired or deleted org",
--- a/messages/org.json
+++ b/messages/org.json
@@ -2,9 +2,9 @@
   "NotFoundOnDevHub": "The scratch org does not belong to the dev hub username %s.",
   "NotADevHub": "The provided dev hub username %s is not a valid dev hub.",
   "NoDevHubFound": "Unable to associate this scratch org with a DevHub",
-  "deleteOrgHubError": "The Dev Hub org cannot be deleted.",
-  "insufficientAccessToDelete": "You do not have the appropriate permissions to delete a scratch org. Please contact your Salesforce admin.",
-  "attemptingToDeleteExpiredOrDeleted": "Attempting to delete an expired or deleted org",
-  "sandboxDeleteFailed": "The sandbox org deletion failed with a result of %s.",
-  "sandboxProcessNotFoundByOrgId": "We can't find a SandboxProcess for the sandbox org %s."
+  "DeleteOrgHubError": "The Dev Hub org cannot be deleted.",
+  "InsufficientAccessToDelete": "You do not have the appropriate permissions to delete a scratch org. Please contact your Salesforce admin.",
+  "ScratchOrgNotFound": "Attempting to delete an expired or deleted org",
+  "SandboxDeleteFailed": "The sandbox org deletion failed with a result of %s.",
+  "SandboxNotFound": "We can't find a SandboxProcess for the sandbox org %s."
 }
--- a/messages/org.json
+++ b/messages/org.json
@@ -1,4 +1,8 @@
 {
-    "NotFoundOnDevHub": "The scratch org does not belong to the dev hub username %s.",
-    "NotADevHub": "The provided dev hub username %s is not a valid dev hub."
+  "NotFoundOnDevHub": "The scratch org does not belong to the dev hub username %s.",
+  "NotADevHub": "The provided dev hub username %s is not a valid dev hub.",
+  "deleteOrgHubError": "The Dev Hub org cannot be deleted.",
+  "insufficientAccessToDelete": "You do not have the appropriate permissions to delete a scratch org. Please contact your Salesforce admin.",
+  "attemptingToDeleteExpiredOrDeleted": "Attempting to delete an expired or deleted org",
+  "sandboxDeleteFailed": "The sandbox org deletion failed with a result of %s."
 }
--- a/messages/org.json
+++ b/messages/org.json
@@ -5,5 +5,6 @@
   "deleteOrgHubError": "The Dev Hub org cannot be deleted.",
   "insufficientAccessToDelete": "You do not have the appropriate permissions to delete a scratch org. Please contact your Salesforce admin.",
   "attemptingToDeleteExpiredOrDeleted": "Attempting to delete an expired or deleted org",
-  "sandboxDeleteFailed": "The sandbox org deletion failed with a result of %s."
+  "sandboxDeleteFailed": "The sandbox org deletion failed with a result of %s.",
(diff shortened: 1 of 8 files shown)


src/org.ts
```diff
@@ -211,6 +211,49 @@ export class Org extends AsyncCreatable<Org.Options> {
...
+      const hubOrgConn = hubOrg.getConnection();
+
+      try {
+        const activeScratchOrgRecordId = (
+          await hubOrgConn.singleRecordQuery<{ Id: string }>(
+            `SELECT Id FROM ActiveScratchOrg WHERE SignupUsername='${this.getUsername()}'`
+          )
+        ).Id;
+        logger.trace(`found matching ActiveScratchOrg with SignupUsername: ${this.getUsername()}.  Deleting...`);
+        await hubOrgConn.delete('ActiveScratchOrg', activeScratchOrgRecordId);
```
mshanemc, line 238: can these happen in parallel?

mshanemc (approved):
approved with 1 small perf suggestion

src/org.ts
```diff
@@ -211,6 +211,49 @@ export class Org extends AsyncCreatable<Org.Options> {
...
+  public async deleteScratchOrg(hubOrg?: Org): Promise<void> {
+    const logger = await Logger.child('deleteScratchOrg');
+
+    // if we didn't get a hubOrg, we'll get it from the this org
+    hubOrg = hubOrg ?? (await this.getDevHubOrg());
+    if (hubOrg?.getOrgId() === this.getOrgId()) {
+      // we're attempting to delete a DevHub
+      throw SfdxError.create('@salesforce/core', 'org', 'deleteOrgHubError');
+    }
+    if (hubOrg) {
```
mshanemc, line 228: what's the scenario where there's not a hubOrg here?  Trying to delete something that doesn't have a hub (like another non-scratch org?)

Would it be better to throw something instead of returning?

src/org.ts
```diff
@@ -211,6 +211,49 @@ export class Org extends AsyncCreatable<Org.Options> {
     }
   }
 
+  /**
+   * If this Org is a scratch org, calling this method will delete the scratch org from the DevHub and clean up any local files
+   *
+   * @param hubOrg - optional DevHub Org of the to-be-deleted scratch org
+   */
+  public async deleteScratchOrg(hubOrg?: Org): Promise<void> {
```
mshanemc, line 219: is it up to the command to figure out if this is scratchOrg or not and then call the right method?

mshanemc (approved):

src/org.ts
WillieRuemmele: that's what's currently happening. I was thinking about creating a general `delete` method that would figure it out and delegate

src/org.ts
WillieRuemmele: I don't think so because the `hubOrgConn.delete` could fail, and the local `this.remove` would clear auth files so your scratch org would be in a remote only state without an easy way (from the CLI) to access it right?

thinking about this some more, that above scenario might be fine. It might be a question of behavior :hmm: 🤔

messages/org.json
```diff
@@ -1,4 +1,10 @@
...
-}
\ No newline at end of file
+  "NotFoundOnDevHub": "The scratch org does not belong to the dev hub username %s.",
+  "NotADevHub": "The provided dev hub username %s is not a valid dev hub.",
+  "NoDevHubFound": "Unable to associate this scratch org with a DevHub",
+  "deleteOrgHubError": "The Dev Hub org cannot be deleted.",
+  "insufficientAccessToDelete": "You do not have the appropriate permissions to delete a scratch org. Please contact your Salesforce admin.",
+  "attemptingToDeleteExpiredOrDeleted": "Attempting to delete an expired or deleted org",
+  "sandboxDeleteFailed": "The sandbox org deletion failed with a result of %s.",
+  "sandboxProcessNotFoundByOrgId": "We can't find a SandboxProcess for the sandbox org %s."
```
shetzel, line 9: If these are error names then they should start with a capital letter and be more generically named.  Leave the details to the message.

src/org.ts
```diff
@@ -486,6 +523,111 @@ export class Org extends AsyncCreatable<Org.Options> {
     throw new SfdxError('Not Supported');
   }
 
+  private async queryProduction(org: Org, field: string, value: string): Promise<{ SandboxInfoId: string }> {
+    return await org.connection.singleRecordQuery<{ SandboxInfoId: string }>(
```
shetzel, line 527: don't need to `return await` here.  Just `return org.connection...`

src/org.ts
```diff
@@ -486,6 +523,111 @@ export class Org extends AsyncCreatable<Org.Options> {
...
+  }
+
+  /**
+   * this method will delete the sandbox org from the production org and clean up any local files
+   *
+   * @param prodOrg - Production org associated with this sandbox
+   * @private
+   */
+  private async deleteSandbox(prodOrg?: Org): Promise<void> {
+    prodOrg =
```
shetzel, line 540: This can be just `prodOrg ??= await Org.create(...);`

src/org.ts
```diff
@@ -486,6 +523,111 @@ export class Org extends AsyncCreatable<Org.Options> {
...
+      // try to calculate sandbox name from the production org
+      // production org: admin@integrationtesthub.org
+      // this.getUsername: admin@integrationtesthub.org.dev1
+      // sandboxName in Production: dev1
+      const sandboxName = (this.getUsername() || '').split(`${prodOrg.getUsername()}.`)[1];
+      if (!sandboxName) {
+        this.logger.debug('Could not construct a sandbox name');
+        throw new Error();
+      }
+      this.logger.debug(`attempting to locating sandbox with username ${sandboxName}`);
```
shetzel, line 559: nit: "attempting to locate sandbox ..."

src/org.ts
```diff
@@ -486,6 +523,111 @@ export class Org extends AsyncCreatable<Org.Options> {
...
+
+  /**
+   * If this Org is a scratch org, calling this method will delete the scratch org from the DevHub and clean up any local files
+   *
+   * @param devHub - optional DevHub Org of the to-be-deleted scratch org
+   * @private
+   */
+  private async deleteScratchOrg(devHub?: Org): Promise<void> {
+    // if we didn't get a devHub, we'll get it from the this org
+    devHub = devHub ?? (await this.getDevHubOrg());
```
shetzel, line 593: Could be just `devHub ??= await this.getDevHubOrg();`

test/unit/orgTest.ts
```diff
@@ -221,6 +221,218 @@ describe('Org Tests', () => {
...
     });
 
+    describe('delete', () => {
+      describe('scratch org', () => {
+        it('should throw error when attempting to delete devhub org', async () => {
+          const dev: Org = await Org.create({
+            connection: await Connection.create({
+              authInfo: await AuthInfo.create({ username: testData.username }),
+            }),
+          });
```
shetzel, line 231: DRY - use a `createOrg` function.

shetzel (approved):

shetzel (approved):

cristiand391:
QA notes:

```ts
await org.delete()
await org.deleteFrom()
```
✅ deleted scratch org, can see its status as `Status: Deleted` in devhub web ui.
✅ auth info json file was deleted from `~/.sfdx`
✅ trying to delete a devhub org throwed error `The Dev Hub org cannot be deleted.`.

TODO: sandbox

Question ❓  : should `sfdx-core` also remove the alias from `~/.sfdx/alias.json`? I can see deleted scratch org usernames still there. Note: I can also see old aliases created some weeks ago in that file, that were deleted with `force:org:delete`.

PR #493, opened 2021-11-09
Sm/library-events

mshanemc:
part 1 of @W-9660828@

- modify lifecycle singleton to replace with newer version <-- please validate this logic!
- adds telemetry/warning events
- move process.emitWarning to the lifecycle.emitWarning in connection/sfdcUrl

### DIFF
--- a/src/lifecycleEvents.ts
+++ b/src/lifecycleEvents.ts
@@ -50,7 +50,7 @@ export class Lifecycle {
 
   /**
    * return the package.json version of the sfdx-core library.
-  */
+   */
   public static staticVersion(): string {
     return pjson.version;
   }
--- a/src/lifecycleEvents.ts
+++ b/src/lifecycleEvents.ts
@@ -77,8 +77,14 @@ export class Lifecycle {
       // an older version was loaded that should be replaced
       compare(global.salesforceCoreLifecycle.version(), Lifecycle.staticVersion()) === -1
     ) {
+      const oldInstance = global.salesforceCoreLifecycle;
       // use the newer version and transfer any listeners from the old version
-      global.salesforceCoreLifecycle = new Lifecycle(global.salesforceCoreLifecycle.listeners);
+      // object spread keeps them from being references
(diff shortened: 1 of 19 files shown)


src/lifecycleEvents.ts
```diff
@@ -32,14 +36,20 @@ declare const global: {
...
+  public static readonly warningEventName = 'warning';
   private debug = Debug(`sfdx:${this.constructor.name}`);
-  private readonly listeners: Dictionary<callback[]>;
 
-  private constructor() {
-    this.listeners = {};
+  private constructor(private readonly listeners: Dictionary<callback[]> = {}) {}
+
+  public static staticVersion(): string {
+    return pjson.version;
```
shetzel, line 52: Should this be private?

src/lifecycleEvents.ts
```diff
@@ -63,11 +73,28 @@ export class Lifecycle {
...
+      global.salesforceCoreLifecycle = new Lifecycle({ ...oldInstance.listeners });
+      // clean up any listeners on the old version
+      Object.keys(oldInstance.listeners).map((eventName) => {
+        oldInstance.removeAllListeners(eventName);
+      });
     }
     return global.salesforceCoreLifecycle;
   }
 
+  public version(): string {
```
shetzel, line 94: Should we js doc this?

shetzel (approved):
a few comments but nothing blocking

src/lifecycleEvents.ts
mshanemc: I don't see the harm in public, and it makes the tests much simpler.

src/lifecycleEvents.ts
mshanemc: yes.

mshanemc: yes, done.

src/lifecycleEvents.ts
mshanemc: I'm not concerned about that, but I did use it in the tests.

PR #495, opened 2021-11-16
feat: missing props for pkgDir per schema

mshanemc:
missing props for 
https://github.com/forcedotcom/sfdx-core/issues/489

### DIFF
--- a/src/sfdxProject.ts
+++ b/src/sfdxProject.ts
@@ -29,6 +29,10 @@ export type PackageDirDependency = {
 export type PackageDir = {
   ancestorId?: string;
   ancestorVersion?: string;
+  apexTestAccess?: {
+    permissionSets?: string | string[];
+    permissionSetLicenses?: string | string[];
+  };
   default?: boolean;
   definitionFile?: string;
   dependencies?: PackageDirDependency[];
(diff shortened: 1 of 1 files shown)


cristiand391 (approved):
For reference:
[`apexTestAccess` docs](https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_unlocked_pkg_unpackaged_md.htm)
[`unpackagedMetadata` docs](https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_dev2gp_unpackaged_md.htm)

src/sfdxProject.ts
```diff
@@ -39,6 +43,9 @@ export type PackageDir = {
   postInstallUrl?: string;
   releaseNotesUrl?: string;
   uninstallScript?: string;
+  unpackagedMetadata?: {
+    path: string;
+  };
```
shetzel, line 48: I wonder if it would be better to have this type live in the schema repo.  I think it should be generated from the schema as well so that any changes to it are automatically reflected in the type.

mshanemc: for clarity: you'd want to have some script in `schemas` that rewrites exported TS types based on their json-schema, then sfdx-core would import (and would have to re-export to avoid breaking changes in coreV2) for those types?

PR #502, opened 2021-12-10
fix: no maxquery warning when records.length is 0

mshanemc:
avoid silly SFDX_MAX_QUERY_LIMIT warning
 
@W-10294697@
https://github.com/forcedotcom/cli/issues/1320

### DIFF
--- a/test/unit/connectionTest.ts
+++ b/test/unit/connectionTest.ts
@@ -270,7 +270,7 @@ describe('Connection', () => {
 
     const records = [{ id: 7 }, { id: 8 }, { id: 10 }, { id: 11 }, { id: 12 }];
     const queryResponse = { totalSize: 50000, done: true, records };
-    requestMock.returns(Promise.resolve(queryResponse));
+    requestMock.resolves(queryResponse);
 
     const conn = await Connection.create({ authInfo: fromStub(testAuthInfo) });
     const toolingQuerySpy = $$.SANDBOX.spy(conn.tooling, 'query');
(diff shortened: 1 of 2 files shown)


test/unit/connectionTest.ts
```diff
@@ -282,6 +282,24 @@ describe('Connection', () => {
     expect(toolingQuerySpy.firstCall.args[1]).to.have.property('maxFetch', 50000);
   });
 
+  it('tooling.autoFetchQuery() should not throw a warning when records is empty', async () => {
+    const soql = 'TEST_SOQL';
+
+    const queryResponse = { totalSize: 5, done: true, records: [] };
+    requestMock.returns(Promise.resolve(queryResponse));
```
WillieRuemmele, line 289: is there a reason you can't do `requestMock.resolves(queryResponse)`?

WillieRuemmele (approved):

test/unit/connectionTest.ts
mshanemc: not that I know of.  I kept it consistent with the rest of the tests. 

gonna change them all.

WillieRuemmele:
✅ : a query resulting in 0 rows like `"SELECT count() FROM PermissionSetAssignment"` doesn't emit the `maxQueryLimit` warning
✅ : the max query limit is still respected
❓ : unable to test default 50,000 query limit, but when setting the limit to 3 I was able to trigger the warning. it should work

PR #503, opened 2021-12-13
feat: switch sfdx-faye with faye

shetzel:
Swaps `sfdx-faye` with `faye`.  The `faye` lib has the fix that `sfdx-faye` was applying.
@W-10288331@


src/status/streamingClient.ts
```diff
@@ -10,7 +10,7 @@ import { resolve as resolveUrl } from 'url';
 import { AsyncOptionalCreatable, Duration, Env, env, set } from '@salesforce/kit/lib';
 import { AnyFunction, AnyJson, ensure, ensureString, JsonMap } from '@salesforce/ts-types/lib';
 // @ts-ignore
```
WillieRuemmele, line 12: could the ignore be removed now?

WillieRuemmele (approved):

src/status/streamingClient.ts
shetzel: Not without providing more typing that I'd rather do in v3.  Good eyes!

PR #504, opened 2021-12-15
Merge main into v3

mdonnalley:
[skip-validate-pr]


src/org/authInfo.ts
```diff
@@ -372,14 +372,16 @@ export class AuthInfo extends AsyncOptionalCreatable<AuthInfo.Options> {
...
+    sfdxAuthUrl: string
+  ): Pick<AuthFields, 'clientId' | 'clientSecret' | 'refreshToken' | 'loginUrl'> {
     const match = sfdxAuthUrl.match(
       /^force:\/\/([a-zA-Z0-9._-]+):([a-zA-Z0-9._-]*):([a-zA-Z0-9._-]+={0,2})@([a-zA-Z0-9._-]+)/
     );
 
     if (!match) {
       throw new SfdxError(
-        'Invalid sfdx auth url. Must be in the format `force://<clientId>:<clientSecret>:<refreshToken>@<loginUrl>`. The instanceUrl must not have the protocol set.',
+        'Invalid SFDX auth URL. Must be in the format "force://<clientId>:<clientSecret>:<refreshToken>@<instanceUrl>".  Note that the SFDX auth URL uses the "force" protocol, and not "http" or "https".  Also note that the "instanceUrl" inside the SFDX auth URL doesn\'t include the protocol ("https://").',
```
peternhale, line 384: @mdonnalley should we start considering blurring the line for messages considering these changes will also be used in sf?

peternhale (changes requested):

.circleci/config.yml
```diff
@@ -118,10 +118,16 @@ workflows:
...
+                - windows
               node_version:
                 - latest
                 - lts
                 - maintenance
+            exclude:
+              - os: windows
+                node_version: lts
+              - os: windows
+                node_version: maintenance
```
RodEsp, line 130: We don't need these exclusions anymore.

package.json
```diff
@@ -31,13 +33,6 @@
     "messages",
     "!lib/**/*.map"
   ],
-  "husky": {
-    "hooks": {
-      "commit-msg": "sf-husky-commit-msg",
-      "pre-commit": "sf-husky-pre-commit",
-      "pre-push": "sf-husky-pre-push"
-    }
-  },
```
RodEsp, line 40: Do we really wanna get rid of these?

mdonnalley: Those are no longer need with husky 7

src/org/connection.ts
```diff
@@ -167,7 +168,8 @@ export class Connection<S extends Schema = Schema> extends JSForceConnection<S>
           }`
         );
       }
-    } catch (e) {
+    } catch (err) {
```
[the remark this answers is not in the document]
mdonnalley: Screen Shot 2021-12-15 at 12 47 29 PM

src/org/authInfo.ts
mdonnalley: This particular method wouldn't surface in sf since it's only for authing with an sfdx auth url

But, yes, we should probably give more thought to making messages applicable to both executables. I think that's outside the scope of this PR though

RodEsp (approved):

peternhale (approved):

PR #507, opened 2021-12-27
UT for scratch org creation

maggiben:
Unit Tests for scratch org creation

@W-9991956@


test/unit/scratchOrgInfoApiTest.ts
```diff
@@ -0,0 +1,539 @@
...
+      create: sinon.stub().resolves(true),
+    } as unknown as SObject<unknown>);
+    stubMethod(sandbox, Org.prototype, 'getConnection').returns(connectionStub);
+  });
+
+  afterEach(() => {
+    sandbox.restore();
+  });
+
+  it('requestScratchOrgCreation', async () => {
```
WillieRuemmele, line 60: NIT: I'm confused on what this is exactly testing, could you expand the test title

test/unit/scratchOrgInfoApiTest.ts
```diff
@@ -0,0 +1,539 @@
...
+import { shouldThrow } from '../../src/testSetup';
+import { Org } from '../../src/org';
+import { Connection } from '../../src/connection';
+import { AuthInfo } from '../../src/authInfo';
+import { MyDomainResolver } from '../../src/status/myDomainResolver';
+import SettingsGenerator from '../../src/scratchOrgSettingsGenerator';
+import {
+  requestScratchOrgCreation,
+  ScratchOrgInfo,
+  JsForceError,
```
WillieRuemmele, line 22: can we convert this to a `SfdxError`? `sfdx-core` shouldn't have a type called `JsForceError`. If it's thrown from `jsforce` we should wrap that Error as a `SfdxError`

test/unit/scratchOrgInfoApiTest.ts
```diff
@@ -0,0 +1,539 @@
...
+    stubMethod(sandbox, AuthInfo, 'create').rejects(err);
+    connectionStub.sobject.withArgs('ScratchOrgInfo').returns({
+      create: sinon.stub().rejects(),
+    } as unknown as SObject<unknown>);
+    const hubOrg = new Org({});
+    const settings = new SettingsGenerator();
+    try {
+      await shouldThrow(requestScratchOrgCreation(hubOrg, TEMPLATE_SCRATCH_ORG_INFO, settings));
+    } catch (error) {
+      expect(error).to.exist;
```
WillieRuemmele, line 112: NIT: do we need to assert any values on the error?

test/unit/util/zipWriterTest.ts
```diff
@@ -0,0 +1,192 @@
...
+    expect(createWriteStreamStub.firstCall.firstArg).to.equal(rootDestination);
+    expect(createReadStreamStub.firstCall.firstArg).to.equal(rootDestination);
+  });
+});
+
+describe('ZipWriter write to file', () => {
+  const sandbox: sinon.SinonSandbox = sinon.createSandbox();
+  const rootDestination = path.join(os.tmpdir(), 'my-zip.zip');
+  const bufferString = 'DEADBEEF';
+  // const rootDestination = path.join('/Users', 'bmaggi', 'tmp', 'my-zip.zip');
```
WillieRuemmele, line 145: NIT: minor cleanup

src/scratchOrgFeatureDeprecation.ts
```diff
@@ -64,25 +66,38 @@ export class ScratchOrgFeatureDeprecation {
...
-    });
-
-    /* If a deprecated feature is identified as deprecated, throw a warning.*/
-    this.featureTypes.deprecatedFeatures.forEach((deprecatedFeature) => {
-      if (requestedFeatures.includes(deprecatedFeature)) {
-        featureWarningMessages.push(messages.getMessage('deprecatedFeatureWarning', [deprecatedFeature]));
-      }
-    });
+      /* If a public quantified feature is defined without a quantity, throw a warning.*/
+      Object.keys(this.featureTypes.quantifiedFeatureMapping).forEach((key) => {
```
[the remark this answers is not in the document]
maggiben: If `features` is a `string[]` then is converted back to string by `join(';')` can't be used as an iterator, it is ultimately  used to test agains every key in `this.featureTypes.*`

src/scratchOrgInfoApi.ts
```diff
@@ -138,7 +138,7 @@ const buildOAuth2Options = async (options: {
 
   if (isJwtFlow && !process.env.SFDX_CLIENT_SECRET) {
     oauth2Options.privateKeyFile = options.hubOrg.getConnection().getAuthInfoFields().privateKey;
-    const retries = options?.retry || env.getNumber('SFDX_JWT_AUTH_RETRY_ATTEMPTS') || 0;
+    const retries = (options && options.retry) || env.getNumber('SFDX_JWT_AUTH_RETRY_ATTEMPTS') || 0;
```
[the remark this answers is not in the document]
maggiben: optional chaining is harder to test (it branches twice vs the `and` operator with does it once), but I agree it's easier to read.

test/unit/scratchOrgFeatureDeprecationTest.ts
```diff
@@ -0,0 +1,98 @@
...
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { expect } from 'chai';
+import { ScratchOrgFeatureDeprecation } from '../../src/scratchOrgFeatureDeprecation';
+
+describe('scratchOrgFeatureDeprecation', () => {
+  const FEATURE_TYPES = {
```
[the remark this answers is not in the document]
maggiben: To be honest these are the only tests I did not write myself I stole them from here: https://github.com/salesforcecli/toolbelt/blob/6b9f3ac97e48139c5f6ca4aea47e418941e62353/src/test/unit/org/scratchOrgFeatureDeprecationTest.ts#L16 any suggestion on where to pick real features to put in here ?

test/unit/scratchOrgInfoApiTest.ts
maggiben: It is only used here to simulate a `jsforce` I will however wrap the `jsforce` error in  an `SfdxError` in function `requestScratchOrgCreation` L367 where the code re-throw

test/unit/scratchOrgInfoApiTest.ts
maggiben: YES! this is interesting because right below there are two a very similar but completely different type of errors thrown by `jsforce` and we need to test them all so it makes sense to test for the `error.message`

test/unit/scratchOrgInfoApiTest.ts
maggiben: It's a test for a basic test on a fully populated `ScratchOrgInfo` object, I added extra comments

src/orgHooks.ts
```diff
@@ -1,48 +0,0 @@
-/*
```
mshanemc, line 1: org hook stuff is moving to `plugin-org`, right?

src/scratchOrgSettingsGenerator.ts
```diff
@@ -185,32 +185,12 @@ export default class SettingsGenerator {
...
         const typeName = upperFirst(item);
         const fname = typeName.replace('Settings', '');
-        const fileContent = this.createSettingsFileContent(typeName, value as Record<string, unknown>);
+        const fileContent = js2xmlparser.parse(typeName, value);
         this.writer.addToZip(fileContent, path.join(settingsDir, fname + '.settings'));
       }
     }
   }
 
-  private createSettingsFileContent(name: string, json: Record<string, unknown>) {
```
mshanemc, line 194: this isn't necessarily related to the tests, but to this pair of PR in general:

Can we throw a nice error if the customer tries to use `OrgPreferenceSettings` BEFORE trying to create the scratch org?

test/unit/scratchOrgCreateTest.ts
```diff
@@ -0,0 +1,128 @@
...
+import { SObject } from 'jsforce'; // RecordResult
+import { Org } from '../../src/org';
+import { Connection } from '../../src/connection';
+import { AuthInfo } from '../../src/authInfo';
+import { scratchOrgCreate, ScratchOrgCreateOptions } from '../../src/scratchOrgCreate';
+import { SfdxProjectJson, SfdxProject } from '../../src/sfdxProject';
+import { Messages } from '../../src/messages';
+
+Messages.importMessagesDirectory(__dirname);
+// const messages = Messages.loadMessages('@salesforce/core', 'scratchOrgCreate');
```
mshanemc, line 18: take it out if you don't need it

test/unit/scratchOrgFeatureDeprecationTest.ts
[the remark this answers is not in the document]
mshanemc: you could test using the ones currently in `ScratchOrgFeatureDeprecation.ts`

test/unit/scratchOrgInfoApiTest.ts
```diff
@@ -0,0 +1,540 @@
...
+    error.name = 'NamedOrgNotFound';
+    stubMethod(sandbox, AuthInfo, 'create').rejects(error);
+    const hubOrg = new Org({});
+    const settings = new SettingsGenerator();
+    const result = await requestScratchOrgCreation(hubOrg, TEMPLATE_SCRATCH_ORG_INFO, settings);
+    expect(result).to.be.true;
+  });
+
+  it('requestScratchOrgCreation no username field in scratchOrgRequest', async () => {
+    const scratchOrgRequest = Object.assign({}, TEMPLATE_SCRATCH_ORG_INFO);
```
mshanemc, line 71: style comment--I really like object spread vs. [clone via assign, then delete props]
`const { Username, ...scratchOrgRequest } = { ...TEMPLATE_SCRATCH_ORG_INFO };`

test/unit/scratchOrgInfoApiTest.ts
```diff
@@ -0,0 +1,540 @@
...
+    delete scratchOrgRequest.Username;
+    const hubOrg = new Org({});
+    const settings = new SettingsGenerator();
+    const result = await requestScratchOrgCreation(hubOrg, scratchOrgRequest, settings);
+    expect(result).to.be.true;
+  });
+
+  it('requestScratchOrgCreation no username field in scratchOrgRequest is empty', async () => {
+    const scratchOrgRequest = Object.assign({}, TEMPLATE_SCRATCH_ORG_INFO);
```
mshanemc, line 80: with spread, this would be 
```
    const scratchOrgRequest = {...TEMPLATE_SCRATCH_ORG_INFO, Username: undefined};
```

mshanemc (approved):
UT look great to me--way better than what we had previously.

Made a few optinoal code style comments.

As I'm looking at this, I'm wondering (and this is for both Ben and Willie) whether all his "org" stuff belongs in a folder under `src` (and under `test`).  I'm ok either way

src/orgHooks.ts
maggiben: yes!

WillieRuemmele (approved):

PR #509, opened 2022-01-13
fix: pin ts to 4.4.4 and bump other deps

peternhale:
@W-0@


package.json
```diff
@@ -35,55 +35,54 @@
...
-    "pretty-quick": "^3.1.0",
-    "shelljs": "0.8.1",
+    "prettier": "^2.5.1",
+    "pretty-quick": "^3.1.3",
+    "shelljs": "0.8.5",
     "sinon": "10.0.0",
-    "ts-node": "^10.0.0",
-    "typescript": "^4.1.3"
+    "ts-node": "^10.4.0",
+    "typescript": "4.4.4"
```
RodEsp, line 85: I don't think pinning `typescript` to `4.4.4` is the right move here. It leaves us open to issues and/or security vulnerabilities since latest is `4.5.4` right now.

RodEsp (changes requested):

package.json
peternhale: I understand Rob, but without the pin we encounter a compile error for which there is no fix. I don't know of an alternative and I am open for suggestions.

RodEsp: Approving this now with a work item to repro and create an issue on TypeScript - W-10437470

RodEsp (approved):
Approving this now but with a WI for follow up work - W-10437470

mdonnalley (approved):

PR #517, opened 2022-01-31
fix: network error tolerance

maggiben:
### What does this PR do?
tolerate ETIMEDOUT and ENOTFOUND on core org create for settings deployment

### PollingClient changes
With the changes in this PR the class `PollingClient` now tolerates network errors and is available to every consumer, no modification to the source needed.
Error will `await Lifecycle.getInstance().emitWarning`

### Functionality Before
Org create throws on first connection problem

### Functionality After

it'll keep trying (within the retry boundaries provided to the poller).
It does emit warnings and debugs about the problem so they're not hidden from the user.

@W-10304836@

Testing/QA
- link `@salesforce/core` in `plugin-org`
- link `plugin-org` in `sfdx` (you can also use ./bin/run)
- try to create an org with large settings then turn off your wifi (or pull a cable if you've got one) might be tricky since settings usually deploy really fast.


src/scratchOrgSettingsGenerator.ts
```diff
@@ -134,9 +135,16 @@ export default class SettingsGenerator {
           logger.debug(`An error occurred trying to check deploy id: ${id}`);
           logger.debug(`Error: ${(error as Error).message}`);
           logger.debug('Re-trying deploy check again....');
-          return {
-            completed: false,
-          };
+          if (
+            ['ETIMEDOUT', 'ENOTFOUND', 'ECONNRESET', 'socket hang up'].some((retryableNetworkError) =>
+              (error as Error).message.includes(retryableNetworkError)
```
WillieRuemmele, line 140: we're casting `error as Error` three times here, should we just set it as a `const err = error as Error`?

test/unit/scratchOrgSettingsGeneratorTest.ts
```diff
@@ -428,6 +434,48 @@ describe('scratchOrgSettingsGenerator', () => {
...
+        .and.to.include('<LightningExperienceSettings>')
+        .and.to.include('<enableS1DesktopEnabled>true</enableS1DesktopEnabled>');
+      expect(addToZipStub.firstCall.args[1]).to.include(path.join('settings', 'LightningExperience.settings'));
+      expect(addToZipStub.secondCall.firstArg)
+        .to.be.a('string')
+        .and.length.to.be.greaterThan(0)
+        .and.to.include('<MobileSettings>')
+        .and.to.include('<enableS1EncryptedStoragePref2>false</enableS1EncryptedStoragePref2>');
+      expect(finalizeStub.callCount).to.equal(1);
+      expect(addToZipStub.secondCall.args[1]).to.include(path.join('settings', 'Mobile.settings'));
```
WillieRuemmele, line 477: can we assert that the `Lifecycle` emit was called ?

test/unit/status/pollingClientTest.ts
```diff
@@ -124,4 +125,56 @@ describe('clientTest', () => {
...
+      },
+      frequency: Duration.milliseconds(100),
+      timeout: Duration.seconds(5),
+    };
+    const client = await PollingClient.create(options);
+    const pollResult = ensureJsonMap(await client.subscribe());
+
+    expect(pollResult.name).equals('foo');
+    expect(emitWarning.callCount).to.be.equal(9 /* errors */);
+    expect(callCount).to.be.equal(9 /* errors */ + 1 /* initial poll */ + 11 /* iterations */);
```
maggiben, line 155: I tried to be descriptive here as to where the value comes from, if it's too verbose I can just put the plain number here.

src/scratchOrgInfoApi.ts
```diff
@@ -353,33 +355,50 @@ export const pollForScratchOrgInfo = async (
...
-    async () => {
-      const resultInProgress = await hubOrg
-        .getConnection()
-        .sobject<ScratchOrgInfo>('ScratchOrgInfo')
-        .retrieve(scratchOrgInfoId);
-      logger.debug(`polling client result: ${JSON.stringify(resultInProgress, null, 4)}`);
-      // Once it's "done" we can return it
-      if (resultInProgress.Status === 'Active' || resultInProgress.Status === 'Error') {
-        return resultInProgress;
+  let resultInProgress!: ScratchOrgInfo;
```
mshanemc, line 358: I avoid `!` where practical.  It's saying, "I 100% guarantee this will be defined" and prevents TS from reminding you to check for that to be the case.

mshanemc (approved):
approvable as is, but see if my ts-error typing makes sense.

src/status/pollingClient.ts
```diff
@@ -62,6 +63,15 @@ export class PollingClient extends AsyncOptionalCreatable<PollingClient.Options>
         result = await this.options.poll();
       } catch (error) {
         errorInPollingFunction = error;
+        if (
+          ['ETIMEDOUT', 'ENOTFOUND', 'ECONNRESET', 'socket hang up'].some((retryableNetworkError) =>
+            error.message.includes(retryableNetworkError)
```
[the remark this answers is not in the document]
WillieRuemmele: I didn't see any errors around this :hmm:

WillieRuemmele (approved):

src/status/pollingClient.ts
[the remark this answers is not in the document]
maggiben: Yes `error` is of type `unknown` and should be casted to `Error` to access property `message`, though I see red wavy lines beneath the variable the linter passes ok.

src/status/pollingClient.ts
```diff
@@ -54,19 +55,28 @@ export class PollingClient extends AsyncOptionalCreatable<PollingClient.Options>
...
+          await Lifecycle.getInstance().emitWarning('Network error occurred. Continuing to poll.');
+          throw SfdxError.wrap(err);
+        }
         // there was an actual error thrown, so we don't want to keep retrying
-        throw new NotRetryableError(error.name);
+        throw new NotRetryableError(err.name);
       }
       if (result.completed) {
-        return result.payload;
+        return result.payload as unknown as T;
```
mshanemc, line 79: needs a comment to explain why we did this, so that it doesn't carry into v3.

src/status/pollingClient.ts
```diff
@@ -54,19 +55,28 @@ export class PollingClient extends AsyncOptionalCreatable<PollingClient.Options>
    * Returns a promise to call the specified polling function using the interval and timeout specified
    * in the polling options.
    */
-  public async subscribe(): Promise<AnyJson | undefined> {
+  public async subscribe<T>(): Promise<T> {
```
mshanemc, line 58:
```suggestion
  public async subscribe<T = AnyJson>(): Promise<T | undefined> {
```
Then people don't have to supply a type (it's always been AnyJson) unless they want to.

I suggest keeping the `undefined` to avoid breaking the contract, even if we're not using it.
And leave a comment explaining why it's there, with a TODO: remove for v3

mshanemc (changes requested):

mshanemc (approved):

cristiand391 (approved):
**QA notes**

Linked `core` into `plugin-org`.
Tried a few times `DEBUG=* sfdx force:org:beta:create -s -f config/project-scratch-def.json -a dreamhous3 --durationdays=1` and turned off wifi

network-error
network-error2

PR #525, opened 2022-02-21
Sm/delete deletes alias

mshanemc:
removes what seems to be an unnecessary check for config.Exists() around removing auth infos.  This was affecting both sandboxes and scratch orgs from what I can see.

That prevents both the defaultusername (config) and aliases from being deleted.

QA notes: 
I would definitely create a sandbox and `org:delete` it.  All of the deletes of config files seemed to be protected with their own `exists` check.

I think there are org:delete sandbox NUTs so you should run those, too.

[@W-10717340@](https://gus.lightning.force.com/a07EE00000pSWdtYAG)

### DIFF
--- a/src/org.ts
+++ b/src/org.ts
@@ -807,30 +807,32 @@ export class Org extends AsyncCreatable<Org.Options> {
     this.logger.debug(`using path for org users: ${config.getPath()}`);
     this.logger.info(`Cleaning up usernames in org: ${this.getOrgId()}`);
 
-    for (const auth of authInfos) {
-      const username = auth.getFields().username;
-      const aliasKeys = (username && aliases.getKeysByValue(username)) || [];
-      aliases.unsetAll(aliasKeys);
+    await Promise.all(
+      authInfos
+        .map((auth) => auth.getFields().username)
+        .map(async (username) => {
+          const aliasKeys = (username && aliases.getKeysByValue(username)) || [];
+          aliases.unsetAll(aliasKeys);
 
-      const orgForUser =
-        username === this.getUsername()
-          ? this
-          : await Org.create({
-              connection: await Connection.create({ authInfo: await AuthInfo.create({ username }) }),
-            });
+          const orgForUser =
+            username === this.getUsername()
+              ? this
+              : await Org.create({
+                  connection: await Connection.create({ authInfo: await AuthInfo.create({ username }) }),
+                });
 
-      const orgType = this.isDevHubOrg() ? Config.DEFAULT_DEV_HUB_USERNAME : Config.DEFAULT_USERNAME;
+          const orgType = this.isDevHubOrg() ? Config.DEFAULT_DEV_HUB_USERNAME : Config.DEFAULT_USERNAME;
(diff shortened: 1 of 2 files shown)


src/org.ts
```diff
@@ -799,43 +799,42 @@ export class Org extends AsyncCreatable<Org.Options> {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   private async removeUsers(throwWhenRemoveFails: boolean): Promise<void> {
```
WillieRuemmele, line 800: we could remove the parameter since it's not used and private

src/org.ts
```diff
@@ -799,43 +799,42 @@ export class Org extends AsyncCreatable<Org.Options> {
...
     this.logger.debug(`Removing users associate with org: ${this.getOrgId()}`);
-    const config = await this.retrieveOrgUsersConfig();
+    const [config, authInfos, aliases] = await Promise.all([
+      this.retrieveOrgUsersConfig(),
+      this.readUserAuthFiles(),
+      Aliases.create(Aliases.getDefaultOptions()),
+    ]);
     this.logger.debug(`using path for org users: ${config.getPath()}`);
-    if (await config.exists()) {
```
WillieRuemmele, line 804: why was `config.exists()` returning (I'm assuming) `false` when the config actually existed?

mshanemc: Because it's based on <orgId>.json (true for sandboxes) but not true for scratch orgs, which use <username>.json.

There's several layers of `remove` that are redundant (see some of the comments) as it tries to handle both scratch org and sandboxes and non-deletable orgs that you can still delete auths/configs/aliases for.

Some of my wilder ideas were a single, idempotent "ensureRemoved" method that looks in all the places for all the things and removes anything it finds but handles not finding things.

src/org.ts
mshanemc: I almost did, and then left it for fear of that thing TS does where it says Org !== Org if private methods have different signatures.

v3 might be a better opportunity.

src/org.ts
mshanemc:
```ts
 public async retrieveOrgUsersConfig(): Promise<OrgUsersConfig> {
    return await OrgUsersConfig.create(OrgUsersConfig.getOptions(this.getOrgId()));
}
```

In the scenario for this bug, there is no OrgUsers file (the one scratch orgs have with the orgId only after you've run `user:create`) so the alias and defaultusername removal gets skipped

WillieRuemmele (approved):

maggiben:
Tested my case where defaultusername wasn't being unset and with this change it is:

image

iowillhoit:
🔎 QA Notes 🔎 
✅ Alias removing before using branch
✅ Alias deletes when using force:org:delete -a the_alias
✅ Alias deletes when using default username force:org:delete
✅ Alias deletes when using username instead of alias force:org:delete -u test-user@example.com
✅ Deletes by alias when alias is not the default user

iowillhoit (approved):

PR #526, opened 2022-02-22
fix: add create warning and info convenience functions

peternhale:
@W-10310042@
add createWarning and createInfo functions


src/messages.ts
```diff
@@ -543,20 +543,90 @@ export class Messages<T extends string> {
...
+   * `info.name` will be the upper-cased key, remove prefixed `info.` and will always end in Info.
+   * `info.actions` will be loaded using `${key}.actions` if available.
+   *
+   * @param key The key of the warning message.
+   * @param tokens The warning message tokens.
+   * @param actionTokens The action messages tokens.
+   * @param codeOrCause The error that caused this warning.
+   * @param cause The underlying error that caused this warning to be created.
+   */
+  public createInfo(
```
mdonnalley, line 583: Why do `createInfo` and `createWarning` create `SfdxErrors`?

peternhale: SfdxError provides the data model needed to express actions. This seemed like the simplest way to accomplish this. I can refactor SfdxError a bit to extract bits needed by all three, but that would probably just be actions.

mdonnalley: Yeah, I think it might be worth putting in the time to abstract out the shared functionality - it feels slightly off to rely on an Error class for info and warning messages

peternhale: The sf-plugins-core change I amde to support info and warn is doing [something similar](https://github.com/salesforcecli/sf-plugins-core/blob/phale/custom-error-messages/src/sfCommand.ts#L225). Do I need to make a similar change there?

peternhale: The real benefit is getting more that a simple string message. The message.createWarning allows for the injection of message and actions, which is one of the components Matt wants for warning and info messages. And a warning could actually come from a caught error, so being able to use an error class as the basis for info/warning seemed reasonable to me.

peternhale: @mdonnalley your thoughts?

peternhale: @mdonnalley I went ahead and created a new type `StructuredMessage` and modified createInfo and createWarning to return that structure.

src/messages.ts
```diff
@@ -543,20 +549,73 @@ export class Messages<T extends string> {
...
+    const message = this.getMessage(key, tokens);
+    let actions;
     try {
-      errActions = this.getMessageWithMap(`${key}.actions`, actionTokens, this.messages);
+      actions = this.getMessageWithMap(`${key}.actions`, actionTokens, this.messages);
     } catch (e) {
       /* just ignore if actions aren't found */
     }
-    return new SfdxError(errMessage, errName, errActions, exitCodeOrCause, cause);
+    return [message, name, actions];
```
mdonnalley, line 618: I think it'd be more idiomatic to return an object instead, `{ message, name, actions }`

That way you can simplify the `createInfo` and `createWarning` methods:

```typescript
createInfo(key, ...) {
  return this.formatMessageContents(key, ...)
}
```

mdonnalley (changes requested):

mdonnalley (approved):

src/messages.ts
```diff
@@ -543,20 +549,77 @@ export class Messages<T extends string> {
...
     // Convert key to name:
-    //     'myMessage' -> `MyMessageError`
-    //     'myMessageError' -> `MyMessageError`
-    //     'error.myMessage' -> `MyMessageError`
-    //     'errors.myMessage' -> `MyMessageError`
-    const errName = `${upperFirst(key.replace(/^errors*\./, ''))}${/Error$/.exec(key) ? '' : 'Error'}`;
-    const errMessage = this.getMessage(key, tokens);
-    let errActions;
+    //     'myMessage' -> `MyMessageWarning`
+    //     'myMessageError' -> `MyMessageWarning`
```
RodEsp, line 612: Is this comment accurate or should it be this instead?
```suggestion
    //     'myMessageError' -> `MyMessageError`
```

RodEsp (approved):

PR #532, opened 2022-02-28
Wr/deprecate

WillieRuemmele:
@W-8776370@

removes the `@deprecated` methods/classes/code


src/config/config.ts
```diff
@@ -104,20 +105,6 @@ export interface ConfigPropertyMetaInput {
 }
 
 export enum SfdxPropertyKeys {
-  /**
```
mdonnalley, line 107: We need to keep these for the foreseeable future so that sfdx-core knows how to map between the username configs used by sfdx and the ones used by sf

src/config/config.ts
```diff
@@ -184,18 +171,6 @@ export const SFDX_ALLOWED_PROPERTIES = [
       failedMessage: messages.getMessage('invalidApiVersion'),
     },
   },
-  {
```
mdonnalley, line 187: same for this

mdonnalley (changes requested):

README.md
```diff
@@ -117,13 +117,13 @@ describe('Using the built in Sinon sandbox.', () => {
 It's important to have negative tests that ensure proper error handling. With `shouldThrow` it's easy to test for expected async rejections.
 
 ```typescript
-import { SfdxError } from '@salesforce/core';
+import { SfError } from '@salesforce/core';
```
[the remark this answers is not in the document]
WillieRuemmele: must've been an issue with the rebase. It looks clean now

mdonnalley (approved):

PR #538, opened 2022-03-14
feat: add TTLConfig

mdonnalley:
Adds `TTLConfig` class for use by `sf deploy metadata`, `sf env create scratch`, `sf env create sandbox`

Example usage:
```typescript
import { Global, TTLConfig } from '@salesforce/core';
import { Duration } from '@salesforce/kit';

type CachedOptions = {
  api: API;
  'target-org': Org | string;
  'test-level': TestLevel;
  'api-version'?: string;
  'dry-run'?: boolean;
  'ignore-errors'?: boolean;
  'ignore-warnings'?: boolean;
  manifest?: string;
  metadata?: string[];
  'source-dir'?: string[];
  tests?: string[];
  wait?: Duration;
  verbose?: boolean;
};

export class DeployCache extends TTLConfig<TTLConfig.Options, CachedOptions> {
  public static getFileName(): string {
    return 'deploy-cache.json';
  }

  public static getDefaultOptions(): TTLConfig.Options {
    return {
      isGlobal: false,
      isState: true,
      filename: DeployCache.getFileName(),
      stateFolder: Global.SF_STATE_FOLDER,
      ttl: Duration.days(1),
    };
  }
}
```

[skip-validate-pr]

### DIFF
--- a/src/config/ttlConfig.ts
+++ b/src/config/ttlConfig.ts
@@ -1,0 +1,50 @@
+/*
+ * Copyright (c) 2022, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { Duration } from '@salesforce/kit';
+import { JsonMap } from '@salesforce/ts-types';
+import { ConfigFile } from './configFile';
+
+/**
+ * A Time To Live configuration file where each entry is timestamped and removed once the TTL has expired.
+ *
+ * @example
+ * import { Duration } from '@salesforce/kit';
+ * const config = await TTLConfig.create({
+ *   isGlobal: false,
+ *   ttl: Duration.days(1)
+ * });
+ */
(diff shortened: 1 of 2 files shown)


peternhale (changes requested):
@mdonnalley Tests?

peternhale (approved):
@mdonnalley thanks for writing the tests!

peternhale (approved):

src/config/ttlConfig.ts
```diff
@@ -0,0 +1,63 @@
...
+  public set(key: string, value: Partial<TTLConfig.Entry<P>>): void {
+    super.set(key, this.timestamp(value));
+  }
+
+  public getLatestEntry(): [string, TTLConfig.Entry<P>] {
+    const entries = this.entries() as Array<[string, TTLConfig.Entry<P>]>;
+    const sorted = entries.sort(([, valueA], [, valueB]) => {
+      return new Date(valueB.timestamp).getTime() - new Date(valueA.timestamp).getTime();
+    });
+    return sorted[0];
```
mshanemc, line 32: empty cache would throw, but it'd be an index out of range?

mshanemc: I think I'd put a test case on the expected empty cache behavior

mshanemc (changes requested):

mshanemc (approved):

PR #540, opened 2022-03-16
fix: move plugin-auth commons funcs to auth info

peternhale:
@W-9711767@

### DIFF
--- a/src/authInfo.ts
+++ b/src/authInfo.ts
@@ -26,7 +26,7 @@ import {
   Nullable,
   Optional,
 } from '@salesforce/ts-types';
-import { OAuth2, OAuth2Options, TokenResponse } from 'jsforce';
+import { OAuth2, OAuth2Options, QueryResult, TokenResponse } from 'jsforce';
 // No typings directly available for jsforce/lib/transport
 // @ts-ignore
 import * as Transport from 'jsforce/lib/transport';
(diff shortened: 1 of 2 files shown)


src/authInfo.ts
```diff
@@ -500,6 +501,79 @@ export class AuthInfo extends AsyncCreatable<AuthInfo.Options> {
...
+
+    const hubAuthInfos = await this.getDevHubAuthInfos();
+    logger.debug(`found ${hubAuthInfos.length} DevHubs`);
+    if (hubAuthInfos.length === 0) return;
+
+    // ask all those orgs if they know this orgId
+    await Promise.all(
+      hubAuthInfos.map(async (hubAuthInfo) => {
+        try {
+          const data = await AuthInfo.queryScratchOrg(hubAuthInfo.getUsername(), fields.orgId as string);
```
WillieRuemmele, line 534: could save _some_ iterations by setting `const hubAuthUsername = hubAuthInfo.getUsername()` instead of calling it 3+ times in a row. Especially since this is in a `Promise.all`

peternhale: @WillieRuemmele `hubAuthInfo.getUsername()` is a simple getter. I could see running it once if it were async or or function was very complex.

WillieRuemmele (approved):

peternhale:
@mshanemc any input on this review?

PR #546, opened 2022-03-28
feat: new cloneSandbox method

maggiben:
### What does this PR do ?
Moves sandbox clone login into sfdx-core

### WI:
@W-10774438@


src/org.ts
```diff
@@ -193,6 +193,25 @@ export class Org extends AsyncCreatable<Org.Options> {
     });
   }
 
+  /**
+   *
+   * @param sandboxReq SandboxRequest options to create the sandbox with
+   * @param sandboxName
+   * @param options Wait: The amount of time to wait before timing out, Interval: The time interval between polling
```
shetzel, line 200: You might want to augment this a bit by adding the return object jsDoc and defining what the default options are.  Maybe even create a type/interface for the options?

src/org.ts
```diff
@@ -193,6 +193,25 @@ export class Org extends AsyncCreatable<Org.Options> {
...
+   */
+  public async cloneSandbox(
+    sandboxReq: SandboxRequest,
+    sandboxName: string,
+    options: { wait?: Duration; interval?: Duration }
+  ): Promise<SandboxProcessObject> {
+    if (sandboxName) {
+      sandboxReq.SourceId = await this.querySandboxInfoIdBySandboxName(sandboxName);
+      this.logger.debug('Clone sandbox sourceId %s', sandboxReq.SourceId);
+    }
```
shetzel, line 210: Is it still a clone if the `sandboxName` was not provided?  I.e., should we throw an error if it's not provided?

src/org.ts
```diff
@@ -987,6 +1006,19 @@ export class Org extends AsyncCreatable<Org.Options> {
...
+   *
+   * @param sandboxNameIn
+   * @returns sandboxInfoId
+   */
+  private async querySandboxInfoIdBySandboxName(sandboxNameIn: string): Promise<string> {
+    this.logger.debug('QuerySandboxInfoIdBySandboxName called with SandboxName: %s ', sandboxNameIn);
+    const queryStr = `SELECT Id, SandboxName FROM SandboxInfo WHERE SandboxName='${sandboxNameIn}'`;
+    const record = await this.connection.singleRecordQuery<{ Id: string }>(queryStr);
+    this.logger.debug('Return from calling queryToolingApi: %s ', record);
+    return record.Id;
```
shetzel, line 1019: Is it possible that `record` is undefined from the query?  If so, this will throw a TypeError.

test/unit/orgTest.ts
```diff
@@ -507,6 +507,37 @@ describe('Org Tests', () => {
...
+            await prod.createSandbox({ SandboxName: 'testSandbox' }, { wait: Duration.seconds(30) });
+          });
+
+          it('will clone the sandbox given a SandBoxName', async () => {
+            await prod.cloneSandbox({ SandboxName: 'testSandbox' }, 'testSandbox', { wait: Duration.seconds(30) });
+            expect(createStub.calledTwice).to.be.true;
+            expect(querySandboxProcessStub.calledTwice).to.be.true;
+            expect(pollStatusAndAuthStub.calledTwice).to.be.true;
+            expect(devHubQueryStub.calledOnce).to.be.true;
+          });
```
shetzel, line 539: We should add a test or 2 for when the clone doesn't go as planned.  I.e., error conditions.

shetzel (changes requested):

src/org.ts
mshanemc: singleRecordQuery throws unless there is exactly 1 record.

src/org.ts
maggiben: Not sure, originally it was written like that in toolbelt, which means if clone can't find a matching sandbox to clone from it would create one for you nonetheless. by throwing instead we might break compatibility, though I've been told this are low traffic commands if you know what I mean, so might be a good idea to throw when we cannot find a parent sandbox to clone from ? what you think @mshanemc ?

src/org.ts
maggiben: @WillieRuemmele you used the same options and return values for `createSandbox` you think is worth making an interface for just for the two options ? what are their default values ? If you think is worth making an interface we should add it to both `cloneSandbox` & `createSandbox` class members, let me know what you think.

src/org.ts
maggiben: If `sandboxName` query returns less than 1 result (`querySandboxInfoIdBySandboxName `) it will throw.
No sandbox will be created if there are no matches

shetzel (approved):

shetzel:
Tested with the new force:org:clone command

PR #547, opened 2022-03-30
fix: metadata rest deploys with jsforce2

mshanemc:
[@W-10743680@](https://gus.lightning.force.com/a07EE00000rPGopYAG)

### DIFF
--- a/src/org/connection.ts
+++ b/src/org/connection.ts
@@ -220,7 +220,11 @@ export class Connection<S extends Schema = Schema> extends JSForceConnection<S> 
       const request: HttpRequest = {
         url: '/metadata/deployRequest',
         method: 'POST',
-        headers: { ...form.getHeaders() },
+        headers: {
+          ...form.getHeaders(),
+          'Sforce-Call-Options': 'client=sfdx-core',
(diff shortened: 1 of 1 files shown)


src/org/connection.ts
```diff
@@ -231,10 +217,17 @@ export class Connection<S extends Schema = Schema> extends JSForceConnection<S>
         contentType: 'application/json',
       });
 
-      const url = `${this.baseUrl()}/metadata/deployRequest`;
-
-      const httpRequest: HttpRequest = { method: 'POST', url, headers, body: form };
-      return this.request(httpRequest);
+      const request: HttpRequest = {
+        url: '/metadata/deployRequest',
```
WillieRuemmele, line 221: `jsforce` resolves the url to include `baseUrl` and any other things correct?

mshanemc: yes, see jsdoc on https://github.com/jsforce/jsforce/blob/7d254241799b8433572cbb435ef4127d2add90b9/src/connection.ts#L645

WillieRuemmele (approved):

mshanemc:
closing...fix is in jsforce2 now so replacement PR will use that

PR #548, opened 2022-03-31
fix: more FormData with buffers and headers

WillieRuemmele:
uses `FormData` and `toBuffer` and `getHeaders` in more places. Especially `auth:device:login`

@W-10743791@


mshanemc (approved):
yep, this needs to be fixed in jsforce but it's better to get all our stuff done on our side than wait.

iowillhoit (approved):
Confirmed this change fixes the auth:device:login

PR #549, opened 2022-04-01
feat: move common authinfo utils to core

peternhale:
@W-9711767@

Moves function identifyPossibleScratchOrgs (and its supporting functions) from plugin-auth to core so they can be used as part of authentication functions.

### DIFF
--- a/src/org/authInfo.ts
+++ b/src/org/authInfo.ts
@@ -23,7 +23,7 @@ import {
   Nullable,
   Optional,
 } from '@salesforce/ts-types';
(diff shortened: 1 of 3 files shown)


WillieRuemmele:
QA:
✅ : dropped in as a replacement for `Common.IdentifyPossibleScratchOrgs` from the parking-orbit branch of `plugin-auth` and all UTs passed
✅ : verified code is essentially copy/paste
✅ : quick manual test of commands that use it

PR #560, opened 2022-04-18
feat: detect if org is a devhub at auth time

cristiand391:
### What does this PR do?

Updates `AuthInfo` class to always set`isDevHub` field based on a server-check.

**Before**:
It was up to consumers to set the field based on a server-check or anything else like [passing a flag in a plugin](https://github.com/salesforcecli/plugin-auth/pull/375)

**After**:
sfdx-core will do a server-side check and set the field to either `true` or `false`.

**QA steps**:

1. Clone [plugin-auth](https://github.com/salesforcecli/plugin-auth/)
2. `yarn link @salesforce/core` into ⬆️ 
3. test all auth commands both with/without `--setdefaultdevhubusername`
4. Even without passing `--setdefaultdevhubusername `, you should see the `isDevHub` field in the authfile.

@[W-10999290](https://gus.my.salesforce.com/apex/ADM_WorkLocator?bugorworknumber=W-10999290)@

### DIFF
--- a/src/org/authInfo.ts
+++ b/src/org/authInfo.ts
@@ -813,6 +813,11 @@ export class AuthInfo extends AsyncOptionalCreatable<AuthInfo.Options> {
         }
       }
 
+      authConfig.isDevHub = await this.determineIfDevHub(
+        ensureString(authConfig.instanceUrl),
+        ensureString(authConfig.accessToken)
+      );
+
       // Update the auth fields WITH encryption
       this.update(authConfig);
     }
@@ -1064,6 +1069,30 @@ export class AuthInfo extends AsyncOptionalCreatable<AuthInfo.Options> {
       errorMsg = `${bodyAsString}`;
     }
     throw new SfError(errorMsg);
+  }
+
+  /**
+   * Returns `true` if the org is a Dev Hub.
+   *
+   * Check access to the ScratchOrgInfo object to determine if the org is a dev hub.
+   */
+  private async determineIfDevHub(instanceUrl: string, accessToken: string): Promise<boolean> {
+    // Make a REST call for the SratchOrgInfo obj directly.  Normally this is done via a connection
+    // but we don't want to create circular dependencies or lots of snowflakes
+    // within this file to support it.
+    const apiVersion = 'v51.0'; // hardcoding to v51.0 just for this call is okay.
+    const instance = ensure(instanceUrl);
+    const baseUrl = new SfdcUrl(instance);
+    const scratchOrgInfoUrl = `${baseUrl}/services/data/${apiVersion}/query?q=SELECT%20Id%20FROM%20ScratchOrgInfo%20limit%201`;
+    const headers = Object.assign({ Authorization: `Bearer ${accessToken}` }, SFDX_HTTP_HEADERS);
+
+    try {
+      await new Transport().httpRequest({ url: scratchOrgInfoUrl, method: 'GET', headers });
+      return true;
+    } catch (err) {
+      /* Not a dev hub */
+      return false;
+    }
   }
 }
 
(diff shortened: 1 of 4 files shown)


src/org/authInfo.ts
```diff
@@ -1065,6 +1070,30 @@ export class AuthInfo extends AsyncOptionalCreatable<AuthInfo.Options> {
...
+    // but we don't want to create circular dependencies or lots of snowflakes
+    // within this file to support it.
+    const apiVersion = 'v51.0'; // hardcoding to v51.0 just for this call is okay.
+    const instance = ensure(instanceUrl);
+    const baseUrl = new SfdcUrl(instance);
+    const scratchOrgInfoUrl = `${baseUrl}/services/data/${apiVersion}/query?q=SELECT%20Id%20FROM%20ScratchOrgInfo%20limit%201`;
+    const headers = Object.assign({ Authorization: `Bearer ${accessToken}` }, SFDX_HTTP_HEADERS);
+
+    try {
+      await new Transport().httpRequest({ url: scratchOrgInfoUrl, method: 'GET', headers });
```
cristiand391, line 1090: The `Connection` class writes the authFile when grabing the latest api version:
https://github.com/forcedotcom/sfdx-core/blob/fa6e4e25255df57f2b80d05bfc62c38d192aa924/src/org/connection.ts#L502-L507

It's up to the consumer to save it or not with `.save()`, so I added this method that only hits the REST endpoint, it doesn't mutate/save the in-memory authConfig.

src/org/authInfo.ts
```diff
@@ -1065,6 +1070,30 @@ export class AuthInfo extends AsyncOptionalCreatable<AuthInfo.Options> {
     }
     throw new SfError(errorMsg);
   }
+
+  /**
+   * Returns `true` if the org is a Dev Hub.
+   *
+   * Check access to the ScratchOrgInfo object to determine if the org is a dev hub.
+   */
+  private async determineIfDevHub(instanceUrl: string, accessToken: string): Promise<boolean> {
```
peternhale, line 1079: @cristiand391 The [Org class already has a function to determine if an org is a devhub](https://github.com/forcedotcom/sfdx-core/blob/cd/detect-devhub/src/org/org.ts#:~:text=public%20async-,determineIfDevHubOrg,-(forceServerCheck%20%3D%20false)%3A%20Promise) Is there a reason it is not being used here?

peternhale (changes requested):

src/org/authInfo.ts
cristiand391: Yes, we can't use it here because it writes the authFile: https://github.com/forcedotcom/sfdx-core/blob/d0c02854ba1e80184589b1eb4b42ceb3a3ef0ee0/src/org/org.ts#L425

For example, let's say a consumer creates a new instance of `AuthInfo`:

```ts
import { Connection,AuthInfo } from '@salesforce/core'

(async() => {
  const user = await AuthInfo.create({
    username: 'cris@test.org',
    oauth2Options: {
      clientId: '1234',
      privateKeyFile: '/home/cris.key'
    }
  })

  user.getUsername()
})()
```

after that code is executed, you should not see an auth file in `~/.sfdx | ~/.sf/sf.json` because you never called `user.save()`.

Both `Org` and `Connection` classes write the auth instance to a file.

peternhale: Is there any code in common between the two that can be extracted into a distinct function?

cristiand391: I don't think so, Steve about this in Slack too, to sum up:

1) This new method in `AuthInfo` is private, Org exposes `determineIfDevHub` as a public method.
2) While we could remove the server check code from [`Org.determineIfDevHub()`](https://github.com/forcedotcom/sfdx-core/blob/d0c02854ba1e80184589b1eb4b42ceb3a3ef0ee0/src/org/org.ts#L414), already created auth files may not have `isDevHub` set and not doing the server check in `Org` could break code like this: https://github.com/forcedotcom/sfdx-core/blob/fa6e4e25255df57f2b80d05bfc62c38d192aa924/src/org/scratchOrgInfoApi.ts#L197

So we want to keep the code in `Org` to handle old auth files, for newly created auth files Org will still check the cached value first.

peternhale (approved):

PR #563, opened 2022-04-19
feat: add SfdxConfigAggregator

mdonnalley:
@W-11025387@

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -8,7 +8,7 @@
 import { dirname as pathDirname, join as pathJoin } from 'path';
 import * as fs from 'fs';
 import { keyBy, parseJsonMap, set } from '@salesforce/kit';
-import { Dictionary, ensure, isBoolean, isString, JsonPrimitive } from '@salesforce/ts-types';
+import { Dictionary, ensure, isBoolean, isString, JsonPrimitive, Nullable } from '@salesforce/ts-types';
 import * as mkdirp from 'mkdirp';
 import { Global } from '../global';
 import { Logger } from '../logger';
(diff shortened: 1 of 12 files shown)


src/config/config.ts
```diff
@@ -457,7 +466,7 @@ export class Config extends ConfigFile<ConfigFile.Options, ConfigProperties> {
     await this.cryptProperties(true);
 
     await super.write();
-    if (Global.SFDX_INTEROPERABILITY) await this.sfdxConfig.writeSync();
+    if (Global.SFDX_INTEROPERABILITY) await this.sfdxConfig.write();
```
peternhale, line 469:
``` suggestion
    if (!this.interoperabilityDisabled()) await this.sfdxConfig.write();
```

peternhale (changes requested):

src/config/config.ts
```diff
@@ -614,27 +622,35 @@ class SfdxConfig {
...
+      const translated = this.translate(config as ConfigProperties, 'toOld');
+      await this._write(translated);
     } catch (error) {
       /* Do nothing */
-      return {};
     }
   }
 
-  public async writeSync(config = this.config.toObject()) {
+  private readSync(): ConfigProperties {
```
WillieRuemmele, line 634: why is there `_readSync` and `readSync`? Same for write?

mdonnalley: I originally did it for some unit tests that I had written for my initial solution. However, those tests are no longer needed since I scraped that solution. But yeah, I can revert this change

src/config/configAggregator.ts
```diff
@@ -408,3 +406,47 @@ export namespace ConfigAggregator {
     ENVIRONMENT = 'Environment',
   }
 }
+
+export class SfdxConfigAggregator extends ConfigAggregator {
+  public getPropertyMeta(key: string): ConfigPropertyMeta {
+    const match = this.getAllowedProperties().find((element) => key === element.key);
+    if (match?.deprecated && match?.newKey) {
```
WillieRuemmele, line 413: do we need all of this deprecation logic, if all `sfdx` config variables aren't deprecated?

PR #569, opened 2022-04-20
WIP - Remove jsen

iowillhoit:


src/schema/validator.ts
```diff
@@ -162,52 +154,22 @@ export class SchemaValidator {
...
-            return obj[name] || obj[prop];
-          };
-          return error.path.split('.').reduce(reducer, schema);
-        };
-
-        const getEnumValues = (): string => {
-          const enumSchema = asJsonMap(getPropValue('enum'));
-          return (enumSchema && getJsonArray(enumSchema, 'enum', []).join(', ')) || '';
-        };
+        const msg = `${error.schemaPath}: ${error.message}`;
```
smithgp, line 164: Sorry to jump in randomly, but, if you haven't already, you could look at using [better-ajv-errors](https://www.npmjs.com/package/better-ajv-errors) to get nicely formatted errors messages out of ajv.

test/unit/schema/validatorTest.ts
```diff
@@ -42,17 +42,83 @@ describe('schemaValidator', () => {
       }
     };
 
-    it('should show additional property', async () => {
+    // We want all validation errors, not just the first error encountered (default)
+    // JSEN used the option `greedy: true`
+    // AJV uses { allErrors: true }
+    it('should display multiple errors', async () => {
       const schema = {
```
iowillhoit, line 49: JSEN used the option `greedy: true`, AJV uses { allErrors: true }

test/unit/schema/validatorTest.ts
```diff
@@ -154,5 +260,31 @@ describe('schemaValidator', () => {
         });
       });
     });
+
+    it('response should include default values', async () => {
```
iowillhoit, line 264: JSEN used the `build()` function, AJV uses { useDefaults: true } option

src/schema/validator.ts
iowillhoit: Hey @smithgp, I had not seen that package and it does look like it does its job well. Although, we are swapping out JSEN for AJV because of [a vulnerability](https://github.com/bugventure/jsen/issues/85), because of that we are trying to reduce any additional dependencies if we can. I think what we've got here is good enough for our use case. Thanks for sharing, always feel free to jump up! 😄

shetzel (approved):
LGTM!

PR #574, opened 2022-05-02
fix(error): no schema with key or ref error #1493

thvd:
Fixes [#1493](https://github.com/forcedotcom/cli/issues/1493)


thvd:
Just signed the CLA, but didn't update this PR yet?

PR #589, opened 2022-05-26
fix(conn): only return emitted records from jsforce

cristiand391:
This PR makes `autoFetchQuery` return records emitted from jsforce instead of the full `response`.

jsforce: https://github.com/jsforce/jsforce/blob/78c03336626e5264f4f5a4f84a9d74822ea274fb/src/query.ts#L636

https://github.com/forcedotcom/cli/issues/1543
@[W-11204187](https://gus.lightning.force.com/a07EE00000xoHkAYAU)@


cristiand391:
Closed in favor of https://github.com/salesforcecli/plugin-data/pull/306

PR #593, opened 2022-06-08

maggiben:

@W-11163191@

### DIFF
--- a/src/scratchOrgSettingsGenerator.ts
+++ b/src/scratchOrgSettingsGenerator.ts
@@ -6,13 +6,14 @@
  */
 
 import * as path from 'path';
+import * as util from 'util';
 import { isEmpty, env, upperFirst, Duration } from '@salesforce/kit';
-import { ensureObject, getObject, JsonMap, Optional } from '@salesforce/ts-types';
+import { ensureObject, getObject, JsonMap /* , Optional  */ } from '@salesforce/ts-types';
 import * as js2xmlparser from 'js2xmlparser';
 import { Org } from './org';
 import { Logger } from './logger';
 import { SfdxError } from './sfdxError';
-import { JsonAsXml } from './util/jsonXmlTools';
+// import { JsonAsXml } from './util/jsonXmlTools';
 import { ZipWriter } from './util/zipWriter';
 import { ScratchOrgInfo } from './scratchOrgInfoApi';
 import { StatusResult } from './status/client';
@@ -35,12 +36,12 @@ export interface ObjectSetting extends JsonMap {
   defaultRecordType?: string;
 }
 
-interface ObjectToBusinessProcessPicklist {
-  [key: string]: {
-    fullName: string;
-    default?: boolean;
-  };
-}
+// interface ObjectToBusinessProcessPicklist {
+//   [key: string]: {
+//     fullName: string;
+//     default?: boolean;
+//   };
+// }
 
 export interface BusinessProcessFileContent extends JsonMap {
   fullName: string;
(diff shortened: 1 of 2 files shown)


src/scratchOrgSettingsGenerator.ts
```diff
@@ -62,7 +78,10 @@ export default class SettingsGenerator {
   private objectSettingsData?: { [objectName: string]: ObjectSetting };
   private logger: Logger;
   private writer: ZipWriter;
+  private allRecTypes: string[] = [];
+  private allBps: string[] = [];
```
RodEsp, line 82: Could you expand these out to the full words instead of using abbreviations please?

It's hard to understand what `RecTypes` and `Bps` means if you aren't super familiar with the codebase.

src/scratchOrgSettingsGenerator.ts
```diff
@@ -210,62 +246,172 @@ export default class SettingsGenerator {
...
-      active: true,
-    };
-    // all the edge cases
-    if (['Case', 'Lead', 'Opportunity', 'Solution'].includes(upperFirst(objectName))) {
-      return { ...output, businessProcess: `${defaultRecordType}Process` };
+  private createObjectFileContent(
+    name: string,
+    json: Record<string, unknown>,
+    allRecTypes: string[],
+    allBps: string[]
```
RodEsp, line 285: Please expand these out to the full word too.

src/scratchOrgSettingsGenerator.ts
```diff
@@ -210,62 +246,172 @@ export default class SettingsGenerator {
...
-      Case: { fullName: 'New', default: true },
-      Lead: { fullName: 'New - Not Contacted', default: true },
-      Solution: { fullName: 'Draft', default: true },
-    };
+    const defaultRecordType = json['defaultRecordType'];
+    if (defaultRecordType) {
+      // We need to keep track of these globally for when we generate the package XML.
+      allRecTypes.push(name + '.' + upperFirst(defaultRecordType as string));
+      let bpName = null;
+      let bpPicklistVal = null;
```
RodEsp, line 301: Please expand `bp` in these variable names to the full words.

src/scratchOrgSettingsGenerator.ts
```diff
@@ -35,12 +36,12 @@ export interface ObjectSetting extends JsonMap {
...
-  };
-}
+// interface ObjectToBusinessProcessPicklist {
+//   [key: string]: {
+//     fullName: string;
+//     default?: boolean;
+//   };
+// }
 
 export interface BusinessProcessFileContent extends JsonMap {
```
mshanemc, line 46: remove the interface if you're not gonna use it

src/scratchOrgSettingsGenerator.ts
```diff
@@ -53,6 +54,21 @@ export interface BusinessProcessFileContent extends JsonMap {
   ];
 }
 
+/** This is the contents of the package.xml that we will use when we deploy settings to a scratch org. */
+
+const packageFileContents = `<?xml version="1.0" encoding="UTF-8"?>
```
mshanemc, line 59: the package.xml structure is the same between source and mdapi format.  

Overall feedback: I strongly recommend using the more recent approach "build a js object that represents the file, then convert it to xml"

the file names will be different (source vs mdapi)
and what were separate files inside the object folder (record types) would now be nodes inside the object.

So, the approach would be:
create an interface that describes the parts of customObject that we care about: https://developer.salesforce.com/docs/atlas.en-us.236.0.api_meta.meta/api_meta/customobject.htmbps
write logic to populate those properties where needed for each CustomObject file
json->xml
write files

PR #594, opened 2022-06-08
Sm/nuts-support

mshanemc:
v3 is serious about caching, which has caused problems with nuts.

StateAggregator now maintains sharded instances of itself by Global.Dir (which references homedir) so that as tests move homedir, StateAggregator will respect that.
- solves the error ex: `     NamedOrgNotFoundError: No authorization information found for test-pqreumf0s1mv@example.com.` where StateAggregator had cached the information from another test

Crypto was caching the key.json value for utils/Cache and later requests to AuthInfo would find the correct file (due to the StateAggregator fix) but try to decrypt it with the key.json value from another test.  It now includes the Global.Dir as part of its cache key so each test gets the correct decrypt key.

### DIFF
--- a/src/global.ts
+++ b/src/global.ts
@@ -80,7 +80,7 @@ export class Global {
   public static readonly LOG_FILE_PATH: string = path.join(Global.SF_DIR, 'sf.log');
 
   /**
-   * Gets the curr ent mode environment variable as a {@link Mode} instance.
+   * Gets the current mode environment variable as a {@link Mode} instance.
    *
    * ```
    * console.log(Global.getEnvironmentMode() === Mode.PRODUCTION);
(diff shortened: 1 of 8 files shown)


src/global.ts
```diff
@@ -61,14 +61,14 @@ export class Global {
...
-   * **See** {@link Global.SF_STATE_FOLDER}
+   * **See**  {@link Global.SF_STATE_FOLDER}
    */
   public static get SF_DIR() {
     return path.join(os.homedir(), Global.SF_STATE_FOLDER);
   }
 
   /**
-   * The full system path to the peferred global state folder
+   * The  full system path to the preferred global state folder
```
peternhale, line 71:
``` suggestion
   * The full system path to the preferred global state folder
```

peternhale (approved):

mdonnalley (approved):

PR #599, opened 2022-06-15
fix: deprecate maxQueryLimit config var

cristiand391:
### What does this PR do?

Marks `maxQueryLimit` as deprecated,  right now it shows as a 'normal' config var so this method:
https://github.com/forcedotcom/sfdx-core/blob/95cfdd0269dbf84b436b26a3d5f885415e526336/src/config/configAggregator.ts#L457
returns the same key.

This should fix SfdxConfigAgg not respecting `SFDX_MAX_QUERY_LIMIT`
https://github.com/salesforcecli/plugin-data/pull/316#issuecomment-1155603365

@W-11204187@
https://github.com/forcedotcom/cli/issues/1543


mshanemc:
QA'd as part of https://github.com/salesforcecli/plugin-data/pull/316

PR #601, opened 2022-06-21

maggiben:

Testing notes:

Link: core in `plugin-org` main branch is ok, then test the beta commands, besides the regular org, customer has provided a custom [org](https://github.com/jonathankretzmer/forcedotcom-issue-scratchOrg-objectSettings) that was failing before you can look it up in this issue: https://github.com/forcedotcom/cli/issues/1526 

@W-11163191@

### DIFF
--- a/src/scratchOrgSettingsGenerator.ts
+++ b/src/scratchOrgSettingsGenerator.ts
@@ -7,12 +7,11 @@
 
 import * as path from 'path';
 import { isEmpty, env, upperFirst, Duration } from '@salesforce/kit';
-import { ensureObject, getObject, JsonMap, Optional } from '@salesforce/ts-types';
+import { ensureObject, getObject, JsonMap, JsonArray } from '@salesforce/ts-types';
 import * as js2xmlparser from 'js2xmlparser';
 import { Org } from './org';
 import { Logger } from './logger';
 import { SfdxError } from './sfdxError';
-import { JsonAsXml } from './util/jsonXmlTools';
 import { ZipWriter } from './util/zipWriter';
 import { ScratchOrgInfo } from './scratchOrgInfoApi';
 import { StatusResult } from './status/client';
@@ -35,24 +34,6 @@ export interface ObjectSetting extends JsonMap {
   defaultRecordType?: string;
 }
 
-interface ObjectToBusinessProcessPicklist {
-  [key: string]: {
-    fullName: string;
-    default?: boolean;
-  };
-}
-
-export interface BusinessProcessFileContent extends JsonMap {
-  fullName: string;
-  isActive: boolean;
-  values: [
-    {
-      fullName: string;
-      default?: boolean;
-    }
-  ];
-}
-
 /**
  * Helper class for dealing with the settings that are defined in a scratch definition file.  This class knows how to extract the
  * settings from the definition, how to expand them into a MD directory and how to generate a package.xml.
@@ -62,7 +43,10 @@ export default class SettingsGenerator {
   private objectSettingsData?: { [objectName: string]: ObjectSetting };
   private logger: Logger;
   private writer: ZipWriter;
+  private allRecordTypes: string[] = [];
+  private allbusinessProcesses: string[] = [];
   private shapeDirName = `shape_${Date.now()}`;
+  private packageFilePath = path.join(this.shapeDirName, 'package.xml');
 
   public constructor() {
     this.logger = Logger.childFromRoot('SettingsGenerator');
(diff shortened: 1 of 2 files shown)


src/scratchOrgSettingsGenerator.ts
```diff
@@ -99,7 +86,8 @@ export default class SettingsGenerator {
   public async deploySettingsViaFolder(scratchOrg: Org, apiVersion: string): Promise<void> {
     const username = scratchOrg.getUsername();
     const logger = await Logger.child('deploySettingsViaFolder');
-    this.createPackageXml(apiVersion);
+    // this.createPackageXml(apiVersion);
+    await this.writePackageFile(this.allRecordTypes, this.allbusinessProcesses, this.packageFilePath, apiVersion);
```
mshanemc, line 90: since `writePackageFile` is a method on the class, you could use the `this.` instance props instead of passing them in.

It might be cleaner to put those in an exported function (could be a different file) that would take inputs and return an object.

Then, your tests could pass a few permutations (with/without record types, BP, objectSettings, etc) and verify that the js objects for `CustomObject` and `package.xml` comes back as expected.

That'd be a really clean input/output function test that would cover most of the complexity since the test changes lost all the assertions about the xml contents

src/scratchOrgSettingsGenerator.ts
```diff
@@ -210,62 +172,168 @@ export default class SettingsGenerator {
...
-    businessProcessName: string
-  ): BusinessProcessFileContent {
-    const objectToBusinessProcessPicklist: ObjectToBusinessProcessPicklist = {
-      Opportunity: { fullName: 'Prospecting' },
-      Case: { fullName: 'New', default: true },
-      Lead: { fullName: 'New - Not Contacted', default: true },
-      Solution: { fullName: 'Draft', default: true },
-    };
+    const defaultRecordType = json['defaultRecordType'];
+    if (defaultRecordType) {
```
mshanemc, line 271: if you check for type instead of truthiness, you can omit all the `defaultRecordType as string` (same with sharing model above)
```suggestion
    if (typeof defaultRecordType === 'string') {
```

src/scratchOrgSettingsGenerator.ts
```diff
@@ -210,62 +172,168 @@ export default class SettingsGenerator {
...
+              },
+            ],
+          },
+        };
+      }
     }
-    return output;
+    output = { ...output, ...{ version: apiVersion } };
+    const xml = js2xmlparser.parse('Package', output);
+    this.writer.addToZip(xml, packageFilePath);
```
mshanemc, line 248: again, imagine how nice the tests for `writePackageFile` could be if it were a pure function that took the inputs it needed, didn't use `this` and returned an object

and let something else convert/write the xml (keeping stuff w/ side effects super simple)

test/unit/scratchOrgSettingsGeneratorTest.ts
```diff
@@ -567,22 +556,9 @@ describe('scratchOrgSettingsGenerator', () => {
...
-        .and.to.include('<RecordType>')
-        .and.to.include('<fullName>My-record-type</fullName>')
-        .and.to.include('<label>My-record-type</label>')
-        .and.to.include('<active>true</active>');
-      expect(addToZipStub.secondCall.args[1]).to.include(
-        path.join('objects', 'MyObject', 'recordTypes', 'My-record-type.recordType')
-      );
+      expect(addToZipStub.callCount).to.equal(1);
+      expect(addToZipStub.firstCall.firstArg).to.be.a('string').and.length.to.be.greaterThan(0);
+      expect(addToZipStub.firstCall.args[1]).to.include(path.join('objects', 'MyObject.object'));
```
mshanemc, line 561: it's writing a file, but not verifying that the file is good.

If you tried to do all these permutations in NUTs, that would be a lot of scratch orgs and you'd have to find a way to check the object settings/record types etc.

that's why the "pure function" approach is good for stuff like this where there's a lot of input/output permutations.

mshanemc (changes requested):
blocking change
- `only` on tests
- tests should assert the object/file are being constructed correctly in the various situations (see comments about testable functions that return objects instead of building an object, converting AND writing to the fs)

src/scratchOrgSettingsGenerator.ts
```diff
@@ -210,62 +172,168 @@ export default class SettingsGenerator {
...
+    apiVersion: string
+  ) {
+    let output = {
+      '@': {
+        xmlns: 'http://soap.sforce.com/2006/04/metadata',
+      },
+      types: [] as JsonArray,
+    };
+    if (this.settingData) {
+      const settingsMemberReferences = Object.keys(this.settingData).reduce((acc, item) => {
```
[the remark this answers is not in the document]
maggiben: I only need the keys

maggiben:
This PR has been closed as a consequence of needing to move this code to V3 [here](https://github.com/forcedotcom/sfdx-core/pull/608) please make your comments on the new PR.
Thank you.

PR #607, opened 2022-06-27
fix: encrypted key regex tests will no longer be stateful

RodEsp:
## What does this PR do?
Removes the `/g` flag from the `encryptedKeys` regexes so that when they are tested against the properties in an auth file they will match regardless of whether they have matched a property beforehand (as is the case when `'accessToken'` is tested before `'refreshToken'` in the same auth file)

Fixes https://github.com/forcedotcom/cli/issues/1314
@W-11349782@

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
   },
   "devDependencies": {
     "@salesforce/dev-config": "^3.0.1",
-    "@salesforce/dev-scripts": "^2.0.1",
+    "@salesforce/dev-scripts": "^2.0.3",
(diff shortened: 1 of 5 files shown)


shetzel:
I tested without the change using the `force:org:list` command and some unencrypted refreshTokens within auth files.  I saw it fail (InvalidEncryptedFormatError and failed to refresh the token).  I tested with the change and it succeeded.  I invalidated all sessions to an org and ran the command again and it succeeded.

PR #608, opened 2022-06-28

maggiben:

Testing notes:

Link: core in `plugin-org` main branch is ok, then test the beta commands, besides the regular org, customer has provided a custom [org](https://github.com/jonathankretzmer/forcedotcom-issue-scratchOrg-objectSettings) that was failing before you can look it up in this issue: https://github.com/forcedotcom/cli/issues/1526 

This PR: was previously created for V1 if you're interested in seeing the comments there please check [here](https://github.com/forcedotcom/sfdx-core/pull/601)

@W-11163191@

### DIFF
--- a/src/org/scratchOrgSettingsGenerator.ts
+++ b/src/org/scratchOrgSettingsGenerator.ts
@@ -159,11 +159,11 @@ export const createRecordTypeAndBusinessProcessFileContent = (
         recordTypes: {
           ...recordTypes,
           businessProcess: businessProcessName,
-          businessProcesses: {
-            fullName: businessProcessName,
-            isActive: true,
-            values,
-          },
+        },
+        businessProcesses: {
+          fullName: businessProcessName,
+          isActive: true,
+          values,
         },
       };
     }
--- a/src/org/scratchOrgSettingsGenerator.ts
+++ b/src/org/scratchOrgSettingsGenerator.ts
@@ -104,6 +104,73 @@ const calculateBusinessProcess = (objectName: string, defaultRecordType: string)
   return [businessProcessName, businessProcessPicklistVal];
 };
 
+export const createRecordTypeAndBusinessProcessFileContent = (
+  objectName: string,
+  json: Record<string, unknown>,
+  allRecordTypes: string[],
+  allbusinessProcesses: string[]
+) => {
+  let output = {
+    '@': {
+      xmlns: 'http://soap.sforce.com/2006/04/metadata',
+    },
+  } as JsonMap;
+  const name = upperFirst(objectName);
+  const sharingModel = json.sharingModel;
+  if (sharingModel) {
+    output = {
+      ...output,
+      sharingModel: upperFirst(sharingModel as string),
+    };
+  }
+
+  const defaultRecordType = json.defaultRecordType;
+  if (typeof defaultRecordType === 'string') {
+    // We need to keep track of these globally for when we generate the package XML.
+    allRecordTypes.push(`${name}.${upperFirst(defaultRecordType)}`);
+    const [businessProcessName, businessProcessPicklistVal] = calculateBusinessProcess(name, defaultRecordType);
+    // Create the record type
+    const recordTypes = {
+      fullName: upperFirst(defaultRecordType),
+      label: upperFirst(defaultRecordType),
+      active: true,
+    };
+
+    output = {
+      ...output,
+      recordTypes: {
+        ...recordTypes,
+      },
+    };
+
+    if (businessProcessName) {
+      // We need to keep track of these globally for the package.xml
+      const values: { fullName: string | null; default?: boolean } = {
+        fullName: businessProcessPicklistVal,
+      };
+
+      if (name !== 'Opportunity') {
+        values.default = true;
+      }
+
+      allbusinessProcesses.push(`${name}.${businessProcessName}`);
+      output = {
+        ...output,
+        recordTypes: {
+          ...recordTypes,
+          businessProcess: businessProcessName,
+          businessProcesses: {
+            fullName: businessProcessName,
+            isActive: true,
+            values,
+          },
+        },
+      };
+    }
+  }
+  return output;
+};
+
 /**
  * Helper class for dealing with the settings that are defined in a scratch definition file.  This class knows how to extract the
  * settings from the definition, how to expand them into a MD directory and how to generate a package.xml.
@@ -234,13 +301,14 @@ export default class SettingsGenerator {
   ) {
     if (this.objectSettingsData) {
       for (const [item, value] of Object.entries(this.objectSettingsData)) {
-        const fileContent = this.createRecordTypeAndBusinessProcessFileContent(
-          upperFirst(item),
+        const fileContent = createRecordTypeAndBusinessProcessFileContent(
+          item,
           value,
           allRecordTypes,
           allbusinessProcesses
         );
-        this.writer.addToZip(fileContent, path.join(objectsDir, upperFirst(item) + '.object'));
+        const xml = js2xmlparser.parse('CustomObject', fileContent);
+        this.writer.addToZip(xml, path.join(objectsDir, upperFirst(item) + '.object'));
       }
     }
   }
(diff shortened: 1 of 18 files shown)


src/org/scratchOrgSettingsGenerator.ts
```diff
@@ -209,62 +229,92 @@ export default class SettingsGenerator {
...
-    // all the edge cases
-    if (['Case', 'Lead', 'Opportunity', 'Solution'].includes(upperFirst(objectName))) {
-      return { ...output, businessProcess: `${defaultRecordType}Process` };
-    }
-    return output;
-  }
+    const defaultRecordType = json['defaultRecordType'];
+    if (typeof defaultRecordType === 'string') {
+      // We need to keep track of these globally for when we generate the package XML.
+      allRecordTypes.push(name + '.' + upperFirst(defaultRecordType));
```
mshanemc, line 254: prefer template literals over adding strings

src/org/scratchOrgSettingsGenerator.ts
```diff
@@ -157,42 +203,16 @@ export default class SettingsGenerator {
...
-        }
+  private async writeObjectSettingsIfNeeded(
+    objectsDir: string,
+    allRecordTypes: string[],
+    allbusinessProcesses: string[]
+  ) {
+    if (this.objectSettingsData) {
+      for (const item of Object.keys(this.objectSettingsData)) {
+        const value: ObjectSetting = this.objectSettingsData[item];
+        const fileContent = this.createObjectFileContent(upperFirst(item), value, allRecordTypes, allbusinessProcesses);
```
mshanemc, line 214: examples of business logic that belong in `createObjectFileContent` are the `upperFirst`--that belongs on L254

so you can pass it the input and it provides the right output.  Otherwise, your tests require you to set upperFirst when sending it stuff

src/org/scratchOrgSettingsGenerator.ts
```diff
@@ -30,23 +29,56 @@ export enum RequestStatus {
...
 }
 
-export interface BusinessProcessFileContent extends JsonMap {
-  fullName: string;
-  isActive: boolean;
-  values: [
-    {
-      fullName: string;
-      default?: boolean;
+export const writePackageFile = ({
```
WillieRuemmele, line 45: 2 comments:

this method doesn't write anything, it just returns a JSON object representation right, so it should probably be `create...` not `write...`

NIT: it's actually creating a manifest, which is commonly a `package.xml`, might want to rename it to `createManifest` or something more appropriate

src/org/scratchOrgSettingsGenerator.ts
```diff
@@ -30,23 +29,56 @@ export enum RequestStatus {
...
+  allRecordTypes = [],
```
WillieRuemmele, line 46: `allRecordTypes` and `allBusinessProcesses` are defaulted to `[]` but marked optional. Usually you just to define a default, or as optional not both. If you define everything with a default you could remove some of the logic below, like `if(settingData)`, since you knew it would always be populated, even if was the default of `[]`

src/org/scratchOrgSettingsGenerator.ts
```diff
@@ -157,42 +203,16 @@ export default class SettingsGenerator {
...
-          });
-          this.writer.addToZip(businessProcessesXml, businessProcessesDir);
-        }
+  private async writeObjectSettingsIfNeeded(
+    objectsDir: string,
+    allRecordTypes: string[],
+    allbusinessProcesses: string[]
+  ) {
+    if (this.objectSettingsData) {
+      for (const item of Object.keys(this.objectSettingsData)) {
```
WillieRuemmele, line 212: NIT: this could be rewritten as `for (const [item, value] of Object.entries(this.objectSettingsData))` to avoid accessing the value below

src/org/scratchOrgSettingsGenerator.ts
```diff
@@ -209,62 +229,92 @@ export default class SettingsGenerator {
...
-    <version>${apiVersion}</version>
-</Package>`;
-    this.writer.addToZip(pkg, path.join(this.shapeDirName, 'package.xml'));
-  }
-
-  private createObjectFileContent(json: Record<string, unknown>) {
-    const output: ObjectSetting = {};
-    if (json.sharingModel) {
-      output.sharingModel = upperFirst(json.sharingModel as string);
+  private createObjectFileContent(
```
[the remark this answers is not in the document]
mshanemc: also, what should type should `createObjectFileContent` return?

src/org/scratchOrgSettingsGenerator.ts
```diff
@@ -209,62 +229,92 @@ export default class SettingsGenerator {
...
+    json: Record<string, unknown>,
+    allRecordTypes: string[],
+    allbusinessProcesses: string[]
+  ) {
+    let output = {
+      '@': {
+        xmlns: 'http://soap.sforce.com/2006/04/metadata',
+      },
+    } as JsonMap;
+    const sharingModel = json['sharingModel'];
```
WillieRuemmele, line 243: is a string accessor required here? why not `json.sharingModel` anymore? Or extend the type definition with 
`json: Record<string, unknown> & { sharingModel?: string },` and avoid casting it `as string`

mshanemc (commented):
nice progress.  Those tests for the package.xml turned out nice!

TODO:
address PR feedback
write some tests that assert the object files

WillieRuemmele (changes requested):

src/org/scratchOrgSettingsGenerator.ts
maggiben: Actually the way it's written it makes perfect sense if you read the test and the way conditionals are being used, as far I know it breaks no rules and allows for:

1. A default value allows to write a very simple conditional like so: `if (allRecordTypes.length > 0)` which saves us from asserting the possible `undefined` value.
2. An optional argument allows to write tests without the need to pass in empty arrays like so: `createObjectFileContent({ apiVersion: '54' });`
Unless there is a good reason not to have them written this way I want to keep it this way.

src/org/scratchOrgSettingsGenerator.ts
maggiben: You're confusing the function which is my fault because I've created a class method and a global function named the same way, I've renamed such functions so that not to confuse readers.

maggiben:
> nice progress. Those tests for the package.xml turned out nice!
> 
> TODO: address PR feedback write some tests that assert the object files

@mshanemc PR feedback addressed, test for object files have been created see `scratchOrgSettingsGeneratorTest.ts` L540 & L557

src/org/scratchOrgSettingsGenerator.ts
```diff
@@ -209,9 +233,13 @@ export default class SettingsGenerator {
     allbusinessProcesses: string[]
   ) {
     if (this.objectSettingsData) {
-      for (const item of Object.keys(this.objectSettingsData)) {
-        const value: ObjectSetting = this.objectSettingsData[item];
-        const fileContent = this.createObjectFileContent(upperFirst(item), value, allRecordTypes, allbusinessProcesses);
+      for (const [item, value] of Object.entries(this.objectSettingsData)) {
+        const fileContent = this.createRecordTypeAndBusinessProcessFileContent(
+          upperFirst(item),
```
mshanemc, line 238: pass in the item, and let upperFirst happen in the function.

you can make it const inside there if you need upperFirst in several places

src/org/scratchOrgSettingsGenerator.ts
```diff
@@ -229,8 +257,8 @@ export default class SettingsGenerator {
     }
   }
 
-  private createObjectFileContent(
-    name: string,
+  private createRecordTypeAndBusinessProcessFileContent(
```
mshanemc, line 260: this should be a function,
return assertable json for testing, 
have a typed return value,
not do xml conversion,
and have tests that assert the various outcomes are correctly put in the json

everything that made the package.xml writing and testing awesome applies to the object files, too

mshanemc (changes requested):

src/org/scratchOrgSettingsGenerator.ts
WillieRuemmele: 1. it's only not breaking rules because the argument into the method is an object, if you broke them apart, so that the method required 4 arguments, it would throw an error 
image

2. if they all had defaults, you could remove the check, because `Object.keys` or however you were iterating over them would continue to work, it would loop 0 times

maybe it's a valid use case of an optional argument having a default, or they should just be optional. I think each option should follow the same pattern. It seems like you could simplify some of the logic below by changing some of this

mshanemc (approved):

mshanemc:
QA notes:

### this branch linked into plugin-org w/ `org:beta:create`
✅ create an org with a features/settings-only scratch def
❌  use [object settings from dev guide](https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_scratch_orgs_def_file.htm)

```
ERROR running force:org:beta:create:  A scratch org was created with username test-ltojdnfmflim@example.com, but the settings failed to deploy due to: 
Error parsing file: Element {http://soap.sforce.com/2006/04/metadata}businessProcesses invalid at this location in type RecordType
```

Screen Shot 2022-07-06 at 3 28 01 PM

mshanemc (dismissed):
yep, these changes fixed the issue I found

mshanemc:
QA, round 2

use [object settings from dev guide](https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_scratch_orgs_def_file.htm)
✅ deploys with no errors or warnings
✅ oppty is private in the org's SharingSettings

PR #616, opened 2022-07-08
fix: add structure writer interface and refactor

peternhale:
settings generator modified to allow 2 writers zip and directory
@W-10833077@


src/org/scratchOrgSettingsGenerator.ts
```diff
@@ -294,10 +297,31 @@ export default class SettingsGenerator {
     }
   }
 
+  public async createDeployPackageContents(apiVersion: string) {
```
shetzel, line 300: All public APIs should have jsDoc

src/util/directoryWriter.ts
```diff
@@ -0,0 +1,51 @@
...
+import { Readable, pipeline as cbPipeline } from 'stream';
+import * as fs from 'fs';
+import * as os from 'os';
+import * as path from 'path';
+import { promisify } from 'util';
+import { StructuredWriter } from './structuredWriter';
+
+const pipeline = promisify(cbPipeline);
+
+export class DirectoryWriter implements StructuredWriter {
```
shetzel, line 17: The class and all public methods should have jsDoc

src/util/directoryWriter.ts
```diff
@@ -0,0 +1,51 @@
...
+      const writeStream = fs.createWriteStream(destPath);
+      await pipeline(contents, writeStream);
+    } else if (typeof contents === 'string') {
+      fs.writeFileSync(destPath, contents);
+    } else if (contents instanceof Buffer) {
+      fs.writeFileSync(destPath, contents);
+    }
+  }
+
+  public finalize(): Promise<void> {
```
shetzel, line 40: This could return the promise from the `pipeline()` call in `addToStore` if we need to keep that method sync.

src/util/zipWriter.ts
```diff
@@ -9,30 +9,36 @@ import { createWriteStream, createReadStream } from 'fs';
...
   private buffers: Buffer[] = [];
 
-  public constructor(private rootDestination?: string) {
+  public constructor(private readonly rootDestination?: string) {
     super({ objectMode: true });
     void pipeline(this.zip, this.getOutputStream());
   }
 
-  public addToZip(contents: string | Readable | Buffer, path: string): void {
+  public async addToStore(contents: string | Readable | Buffer, path: string): Promise<void> {
```
shetzel, line 28: This is a breaking change unless this is totally new code, and it doesn't appear to be.

shetzel (changes requested):

shetzel (approved):

PR #617, opened 2022-07-10
Changes for bundling

nawforce:
This provides a partial fix for #298. There are two sets of changes.  

In src/messages.ts we change `importMessagesDirectory` so it won't error if a messages directory is not found. It returns a status flag indicating if the directory was located if needed. Then in src/exported.ts we provide all the arguments to` Messages.importMessagesDirectory` so that `importMessagesDirectory` does not need to locate the package.json file.

In the second set of changes (all the other files) we move message loading out of static scopes so they are loaded as needed which appears to be the original intent of the design.

This is only a partial fix as if you want to bundle you now need to arrange for message file loading to occur yourself in a way that is supported by the bundler you are using and this must happen before you use any features of the library. You can use `Messages.setLoaderFunction` to provide a custom loader. With webpack I have tested the changes via loading with require, such as `require('@salesforce/core/messages/scratchOrgCreate.json')` and constructing a Map in the usual way. In this case the require argument must be a literal string for webpack to be able to bundle the file for you.


nawforce:
Example webpack config I used for testing

[webpack.config.js.txt](https://github.com/forcedotcom/sfdx-core/files/9086723/webpack.config.js.txt)

mshanemc:
@nawforce @lukecotter there's a PR open that's meant to deal with this in a more elegant way (and can also work for other libraries/plugins).

https://github.com/forcedotcom/sfdx-core/pull/758

I would love to get one of y'all webpackers to let me know if this works for your needs, since we don't use it ourselves.

lukecotter:
@mshanemc 

Sorry for the slow reply, I only just got round to testing this.
Looks like this change did the trick for this issues however I have hit a couple more when trying to bundle.

I have raised two PRs in the libraries causing the problem
https://github.com/jsforce/jsforce/pull/1343
https://github.com/forcedotcom/node-bunyan/pull/14

**JSforce**
JSforce has some types being exported  from a types file and imported in another.
babel strips the types out and leave the imports, which bundlers do not like.
my fix use the type only anotation to make sure the imports are removed too.
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export

**node-bunyan**
node bunyan was hiding some of the requires which led to runtime failure when the JSforce issue is resolved.

Let me know what you think of the changes it would be great to get them in so I can use these libraries in the [debug log analyzer](https://github.com/certinia/debug-log-analyzer)

PR #624, opened 2022-07-25
remove mkdirp in favor of native dep

maggiben:
Description:

Remove `mkdirp` dependency in favor of native code.

Work Item
@W-11286801@

### DIFF
--- a/src/global.ts
+++ b/src/global.ts
@@ -5,10 +5,11 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
+import * as fs from 'fs';
 import * as os from 'os';
 import * as path from 'path';
 import { env } from '@salesforce/kit';
-import * as mkdirp from 'mkdirp';
+import { SfError } from './sfError';
 
 /**
  * Represents an environment mode.  Supports `production`, `development`, `demo`, and `test`
@@ -98,6 +99,11 @@ export class Global {
    */
   public static async createDir(dirPath?: string): Promise<void> {
     dirPath = dirPath ? path.join(Global.SFDX_DIR, dirPath) : Global.SFDX_DIR;
-    await mkdirp(dirPath, '700');
+    try {
+      await fs.promises.mkdir(dirPath, { recursive: true });
+      await fs.promises.chmod(dirPath, '700');
+    } catch (error) {
+      throw new SfError(`Failed to create directory or set permissions for: ${dirPath}`);
+    }
   }
 }
--- a/src/global.ts
+++ b/src/global.ts
@@ -100,7 +100,7 @@ export class Global {
   public static async createDir(dirPath?: string): Promise<void> {
     dirPath = dirPath ? path.join(Global.SFDX_DIR, dirPath) : Global.SFDX_DIR;
     try {
-      if (process.platform === "win32") {
+      if (process.platform === 'win32') {
         await fs.promises.mkdir(dirPath, { recursive: true });
       } else {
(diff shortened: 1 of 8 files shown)


src/global.ts
```diff
@@ -98,6 +99,11 @@ export class Global {
    */
   public static async createDir(dirPath?: string): Promise<void> {
     dirPath = dirPath ? path.join(Global.SFDX_DIR, dirPath) : Global.SFDX_DIR;
-    await mkdirp(dirPath, '700');
+    try {
+      await fs.promises.mkdir(dirPath, { recursive: true });
+      await fs.promises.chmod(dirPath, 0o700);
```
cristiand391, line 104: the permissions can be specified when creating the folder, see: https://nodejs.org/api/fs.html#fspromisesmkdirpath-options

peternhale (approved):

src/global.ts
```diff
@@ -98,6 +99,10 @@ export class Global {
    */
   public static async createDir(dirPath?: string): Promise<void> {
     dirPath = dirPath ? path.join(Global.SFDX_DIR, dirPath) : Global.SFDX_DIR;
-    await mkdirp(dirPath, '700');
+    try {
+      await fs.promises.mkdir(dirPath, { recursive: true, mode: 0o700 });
```
mshanemc, line 103: this is worth a special UT test on windows (or manual check via repl) since the `mode` option isn't supported.  It's not clear from the node docs if that means "it doesn't do anything" or "it throws"

https://nodejs.org/api/fs.html#fsmkdirpath-options-callback

mshanemc (commented):
see comment about windows file perms.

src/global.ts
peternhale, line 103: Make it safe
``` suggestion
      if (process.platform === "win32") {
        await fs.promises.mkdir(dirPath, { recursive: true });
      } else {
        await fs.promises.mkdir(dirPath, { recursive: true, mode: 0o700 });
      }
```

src/logger.ts
```diff
@@ -415,9 +414,7 @@ export class Logger {
       await fs.promises.access(logFile, fs.constants.W_OK);
     } catch (err1) {
       try {
-        await mkdirp(path.dirname(logFile), {
-          mode: '700',
-        });
+        await fs.promises.mkdir(path.dirname(logFile), { recursive: true, mode: 0x700 });
```
peternhale, line 417: Make it platform safe here too

peternhale: When using recursive === true, the operation will not throw if the path already exists, so the catch should not be a `noop`;

See https://nodejs.org/api/fs.html#fspromisesmkdirpath-options

peternhale (changes requested):

src/logger.ts
```diff
@@ -457,9 +458,8 @@ export class Logger {
       fs.accessSync(logFile, fs.constants.W_OK);
     } catch (err1) {
       try {
-        mkdirp.sync(path.dirname(logFile), {
-          mode: '700',
-        });
+        fs.mkdirSync(path.dirname(logFile));
+        fs.chmodSync(path.dirname(logFile), 0o700);
```
peternhale, line 462: @maggiben does chmodSync have the same limitation for changing file mode on Windows?

maggiben: I can use `mod` in `mkdirSync` and do it only if NOT win32

src/global.ts
maggiben: `mod` only runs if platform is NOT win32

peternhale (approved):

cristiand391 (approved):

PR #629, opened 2022-08-02
fix: add interop between templates config/env var

WillieRuemmele:
@W-11445939@

adds interop between old/new config value for `plugin-templates`

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -40,6 +40,7 @@ const messages = Messages.load('@salesforce/core', 'config', [
   'restDeploy',
   'instanceUrl',
   'disable-telemetry',
+  'customOrgMetadataTemplates',
 ]);
 
 const SFDX_CONFIG_FILE_NAME = 'sfdx-config.json';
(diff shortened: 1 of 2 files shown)


src/config/config.ts
```diff
@@ -271,6 +272,12 @@ export const SFDX_ALLOWED_PROPERTIES = [
       failedMessage: messages.getMessage('invalidBooleanConfigValue'),
     },
   },
+  {
+    key: SfdxPropertyKeys.CUSTOM_ORG_METADATA_TEMPLATES,
+    newKey: OrgConfigProperties.ORG_CUSTOM_METADATA_TEMPLATES,
+    deprecated: true,
```
randi274, line 278: Should this be deprecated: false, since this is still being used?

WillieRuemmele:
### QA Notes:
---

✅ :`config:set customOrgMetadataTemplates`
✅ :`config:unset customOrgMetadataTemplates`
✅ : set/unset `org-custom-metadata-templates` in `.sf/config.json`
✅ : picked up correct config with `ConfigAggregator`

mdonnalley (approved):

randi274 (approved):
Looks good! Ran QA on this together synced up with templates. Important to note that just the old value (customOrgMetadataTemplates) is the one that we need set in VS Code, as that's the one that gets read from the config aggregator in templates.

PR #639, opened 2022-09-15
fix: add helper functions to project for pkging

peternhale:
### What does this PR do?
Add some additional functions in SfProject for packaging needs
@W-11748608@

### DIFF
--- a/src/sfProject.ts
+++ b/src/sfProject.ts
@@ -303,7 +303,7 @@ export class SfProjectJson extends ConfigFile {
    * for data other than the names.
    */
   public getUniquePackageNames(): string[] {
-    return this.getUniquePackageDirectories().map((pkgDir) => pkgDir.name) || [];
+    return this.getUniquePackageDirectories().map((pkgDir) => pkgDir.name);
   }
 
   /**
@@ -605,17 +605,6 @@ export class SfProject {
   }
 
   /**
-   * Returns the package directory that has the given property whose value equals the given value.
-   *
-   * @param property
-   * @param value
-   */
-  public getPackageDirectoryWithProperty(property: string, value: string): Optional<NamedPackageDir> {
-    const packageDirs = this.getPackageDirectories();
-    return packageDirs.find((packageDir) => Reflect.get(packageDir, property) === value);
-  }
-
-  /**
    * Returns the package directory.
    *
    * @param packageName Name of the package directory.  E.g., 'force-app'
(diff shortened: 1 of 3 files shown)


src/sfProject.ts
```diff
@@ -544,6 +604,17 @@ export class SfProject {
     return packageDir ? packageDir.name : undefined;
   }
 
+  /**
+   * Returns the package directory that has the given property whose value equals the given value.
+   *
+   * @param property
+   * @param value
+   */
+  public getPackageDirectoryWithProperty(property: string, value: string): Optional<NamedPackageDir> {
```
shetzel, line 613: Is this a common usecase?  This is essentially a filter.  I.e., `getPackageDirectories().filter(pd => pd.prop === value)` where the filter returns a single match.

src/sfProject.ts
```diff
@@ -656,6 +727,34 @@ export class SfProject {
 
     return this.projectConfig;
   }
+
+  public async hasPackageAliases(): Promise<boolean> {
+    return await this.getSfProjectJson().hasPackageAliases();
```
shetzel, line 732:
```suggestion
    return this.getSfProjectJson().hasPackageAliases();
```
based on: https://jakearchibald.com/2017/await-vs-return-vs-return-await/

src/sfProject.ts
```diff
@@ -656,6 +727,34 @@ export class SfProject {
...
+  }
+  /**
+   * Returns a read-only list of `packageDirectories` within sfdx-project.json, first reading
+   * and validating the file if necessary. i.e. modifying this array will not affect the
+   * sfdx-project.json file.
+   */
+  public getPackageAliases(): Nullable<Dictionary<string>> {
```
shetzel, line 739: this is already defined above

shetzel (changes requested):

src/sfProject.ts
```diff
@@ -302,21 +303,79 @@ export class SfProjectJson extends ConfigFile {
...
   public hasMultiplePackages(): boolean {
-    return this.getContents().packageDirectories && this.getContents().packageDirectories.length > 1;
+    return this.getContents()?.packageDirectories?.length > 1;
+  }
+
+  /**
+   * Has at least one package alias defined in the project.
+   */
+  public async hasPackageAliases() {
+    return Object.keys(this.getContents().packageAliases || {}).length > 0;
```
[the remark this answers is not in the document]
peternhale: Just keeping things aligned with the current complexity of the class. I do imagine more to follow.

src/sfProject.ts
peternhale: Right now the function I wanted to replace from pkg utils is used twice, both are looking up which package dir is the default. I will remove this for now and we can  remove refs to `getConfigPackageDirectory` in pkg utils with calls to project.getDefaultPackage

src/sfProject.ts
peternhale: The other function is on SfProjectJson. Once again, mirroring the current impl

shetzel (approved):

mdonnalley (approved):

PR #660, opened 2022-09-23
fix: fix package dir functions

peternhale:
@W-11786892@

### DIFF
--- a/src/sfProject.ts
+++ b/src/sfProject.ts
@@ -363,6 +363,11 @@ export class SfProjectJson extends ConfigFile {
    * @param packageDir
    */
   public addPackageDirectory(packageDir: NamedPackageDir): void {
+    // there is no notion of uniqueness in package directory entries
+    // so an attempt of matching an existing entry is a bit convoluted
+    // an entry w/o a package or id is considered a directory entry for which a package has yet to be created
+    // so first attempt is to find a matching dir entry that where path is the same and id and package are not present
+    // if that fails, then find a matching dir entry package is present and is same as the new entry
     const dirIndex = this.getContents().packageDirectories.findIndex((pd) => {
       const withId = pd as NamedPackageDir & { id: string };
       return (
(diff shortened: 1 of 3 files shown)


src/sfProject.ts
```diff
@@ -284,13 +284,13 @@ export class SfProjectJson extends ConfigFile {
    * for packaging operations that want to do something for each package entry.
    */
   public getUniquePackageDirectories(): NamedPackageDir[] {
-    const visited: Dictionary<boolean> = {};
+    const visited: Set<string> = new Set<string>();
```
WillieRuemmele, line 287: maybe a bit more than the change intended, but couldn't this be rewritten as `return new Set<string>(this.getPacageDirectoriesSync())` 

we shouldn't need to worry about adding duplicate values to a Set

src/sfProject.ts
```diff
@@ -611,7 +615,15 @@ export class SfProject {
    */
   public getPackage(packageName: string): Optional<NamedPackageDir> {
     const packageDirs = this.getPackageDirectories();
-    return packageDirs.find((packageDir) => packageDir.name === packageName);
+    return packageDirs.find((packageDir) => packageDir.package === packageName || packageDir.name === packageName);
```
WillieRuemmele, line 618: could this have any unintended consequences? it would be safer to check the original condition first right?

src/sfProject.ts
peternhale: The goal is to retain order if dir entries found in pjson. Placing contents into any collection would not guarantee original order.

src/sfProject.ts
peternhale: I think both are wrong. I am so confused by what to compare in an entry.

WillieRuemmele (approved):

src/sfProject.ts
```diff
@@ -362,9 +362,13 @@ export class SfProjectJson extends ConfigFile {
...
    */
-  public addPackageDirectory(packageDir: PackageDir) {
-    const dirIndex = this.getContents().packageDirectories.findIndex((pkgDir) => {
-      return pkgDir.package === packageDir.package;
+  public addPackageDirectory(packageDir: NamedPackageDir): void {
+    const dirIndex = this.getContents().packageDirectories.findIndex((pd) => {
+      const withId = pd as NamedPackageDir & { id: string };
+      return (
+        (withId.path === packageDir.path && !withId.id && !withId.package) ||
+        (!!packageDir.package && packageDir.package === withId.package)
```
shetzel, line 370: Perhaps a comment explaining why we need to check 3 fields for matching?

shetzel (approved):

WillieRuemmele (approved):

PR #665, opened 2022-09-30
fix: add tooling query support to autoFetchQuery

peternhale:
### What does this PR do?
Add an option to specify that tooling query function should be used for the requested query.
@W-11786892@


src/org/connection.ts
```diff
@@ -342,19 +342,21 @@ export class Connection<S extends Schema = Schema> extends JSForceConnection<S>
...
+    queryOptions: Partial<QueryOptions & { tooling: boolean }> = { tooling: false }
   ): Promise<QueryResult<T>> {
     const config: ConfigAggregator = await ConfigAggregator.create();
     // take the limit from the calling function, then the config, then default 10,000
     const maxFetch: number =
       (config.getInfo(OrgConfigProperties.ORG_MAX_QUERY_LIMIT).value as number) || queryOptions.maxFetch || 10000;
 
+    const tooling = queryOptions?.tooling;
+    delete queryOptions.tooling;
+
```
mshanemc, line 354:
```suggestion
    const { tooling, ...queryOptionsWithoutTooling }  = queryOptions;
```

and then use the new constant below.

logic: generally avoid mutating parameters passed to functions (especially objects passed by reference) because whoever calls this may not be done with the object and shouldn't be aware that you're changing it.

mshanemc (changes requested):

WillieRuemmele (approved):

PR #669, opened 2022-10-07
fix: ensure the hubOrg passes is actually a DevHub, throw error if not

WillieRuemmele:
### What does this PR do?
throws a nicer error message when the `hubOrg` is not actually enabled as a `DevHub`

# Before
`ERROR running force:org:create:  The requested resource does not exist`

# After

`ERROR running force:org:create:  Ensure that org, with id: test-podeygqvfh7u@example.com and username 00D530000008eONEAY is enabled as a DevHub`

*Messaging to be reviewed by Juliet

@W-11838125@

### DIFF
--- a/messages/scratchOrgInfoApi.md
+++ b/messages/scratchOrgInfoApi.md
@@ -16,4 +16,4 @@ Successfully created org with ID: %s and name: %s. Unfortunately, source trackin
 
 # hubOrgIsNotDevHub
 
-Ensure that org with username: %s and id %s is enabled as a DevHub
+Make sure that the org with username %s and ID %s is enabled as a Dev Hub. To enable it, open the org in your browser, navigate to the Dev Hub page in Setup, and click Enable.
--- a/messages/scratchOrgInfoApi.md
+++ b/messages/scratchOrgInfoApi.md
@@ -13,3 +13,7 @@ We've deprecated OrgPreferences. Update the scratch org definition file to repla
 # SourceStatusResetFailureError
 
 Successfully created org with ID: %s and name: %s. Unfortunately, source tracking isn’t working as expected. If you run force:source:status, the results may be incorrect. Try again by creating another scratch org.
+
+# hubOrgIsNotDevHub
+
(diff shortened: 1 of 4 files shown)


messages/scratchOrgInfoApi.md
```diff
@@ -13,3 +13,7 @@ We've deprecated OrgPreferences. Update the scratch org definition file to repla
 # SourceStatusResetFailureError
 
 Successfully created org with ID: %s and name: %s. Unfortunately, source tracking isn’t working as expected. If you run force:source:status, the results may be incorrect. Try again by creating another scratch org.
+
+# hubOrgIsNotDevHub
+
+Ensure that org with username: %s and id %s is enabled as a DevHub
```
jshackell-sfdc, line 19:
```suggestion
Make sure that the org with username %s and ID %s is enabled as a Dev Hub. To enable it, open the org in your browser, navigate to the Dev Hub page in Setup, and click Enable.
```

jshackell-sfdc: Let me know if this is too long.   I just figured it was easy to explain how to enable an org as a Dev Hub; users might be baffled as to how to go about it.

cristiand391 (approved):

src/org/scratchOrgInfoApi.ts
```diff
@@ -277,6 +277,9 @@ export const requestScratchOrgCreation = async (
   scratchOrgRequest: ScratchOrgInfo,
   settings: SettingsGenerator
 ): Promise<SaveResult> => {
+  if (!hubOrg.isDevHubOrg()) {
```
cristiand391, line 280: note: this method relies on the `isDevHub` prop to be set in the auth file, this prop is set based on a server check when creating and saving the auth file after this change was merged: https://github.com/forcedotcom/sfdx-core/pull/560#discussion_r853445695

if a user auth'd to a hub before that change ⬆️ was merged and starts to get the error then a re-auth should solve it.

PR #678, opened 2022-10-14
fix: getPackageFromPath now handles nested package paths

peternhale:
### What does this PR do?
Modifies getPackageFromPath to better handle matching logic between a package dir entry's fullPath property and the path passed to the function.
@W-11903179@

### DIFF
--- a/src/sfProject.ts
+++ b/src/sfProject.ts
@@ -4,7 +4,7 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
-import { basename, dirname, isAbsolute, normalize, sep } from 'path';
+import { basename, dirname, isAbsolute, normalize, resolve, sep } from 'path';
 import * as fs from 'fs';
 import { defaults, env } from '@salesforce/kit';
 import { Dictionary, ensure, JsonMap, Nullable, Optional } from '@salesforce/ts-types';
@@ -602,9 +602,15 @@ export class SfProject {
    */
   public getPackageFromPath(path: string): Optional<NamedPackageDir> {
     const packageDirs = this.getPackageDirectories();
-    const match = packageDirs.find(
-      (packageDir) => basename(path) === packageDir.path || path.includes(packageDir.fullPath)
-    );
+    // resolve the given path
+    const fullPath = resolve(path);
+    const match = packageDirs.find((packageDir) => {
+      // fullPath will not have a trailing slash, so remove it from packageDir.fullPath, if it exists
+      const fullPathSansTrailingSep = packageDir.fullPath.replace(/(\\|\/)$/, '');
+      return (
+        basename(path) === packageDir.path || fullPath === fullPathSansTrailingSep || path.includes(packageDir.fullPath)
+      );
+    });
     return match;
   }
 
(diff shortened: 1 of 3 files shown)


src/sfProject.ts
```diff
@@ -602,9 +602,15 @@ export class SfProject {
...
+    // resolve the given path
+    const fullPath = resolve(path);
+    const match = packageDirs.find((packageDir) => {
+      // fullPath will not have a trailing slash, so remove it from packageDir.fullPath, if it exists
+      const fullPathSansTrailingSep = packageDir.fullPath.replace(/(\\|\/)$/, '');
+      return (
+        basename(path) === packageDir.path || fullPath === fullPathSansTrailingSep || path.includes(packageDir.fullPath)
+      );
+    });
     return match;
```
mshanemc, line 614: suggestion (add comments if you want to explain the trailing slash part)
```ts
return this.getPackageDirectories().find(
      (packageDir) =>
        basename(path) === packageDir.path || path.includes(packageDir.fullPath) || packageDir.fullPath.startsWith(resolve(path))
    );
```

mshanemc: it puts the `resolve` last so its avoided if possible

peternhale: > it puts the `resolve` last so its avoided if possible

Not following. Are you asking that the use of resolve should be done after the other conditions are checked?

cristiand391 (approved):

src/sfProject.ts
cristiand391: If you wanna go with Shane's suggestion make sure to change this:

```diff
-path.includes(packageDir.fullPath)
+packageDir.fullPath.includes(path)

the pkg dir fullPath always have a trailing slash:
https://github.com/forcedotcom/sfdx-core/blob/main/src/sfProject.ts#L246

File: src/sfProject.ts [hunk 2]
```diff
@@ -602,9 +602,15 @@ export class SfProject {
...
-    const match = packageDirs.find(
-      (packageDir) => basename(path) === packageDir.path || path.includes(packageDir.fullPath)
-    );
+    // resolve the given path
+    const fullPath = resolve(path);
+    const match = packageDirs.find((packageDir) => {
+      // fullPath will not have a trailing slash, so remove it from packageDir.fullPath, if it exists
+      const fullPathSansTrailingSep = packageDir.fullPath.replace(/(\\|\/)$/, '');
+      return (
+        basename(path) === packageDir.path || fullPath === fullPathSansTrailingSep || path.includes(packageDir.fullPath)
```
cristiand391, line 611: if I run `sfdx force:package:beta:version:create --json -c -v user@domain.com -x -b refs/pull/46/merge -d folder_a/folder_b -f ./config/project-scratch-def.json -t Tag__20221008.7a -w 90`

then `path` will be `folder_a/folder_b`, but `packageDir.fullPath` is an absolute path so this:

```typescript
path.includes(packageDir.fullPath)
```

will always return false.

plugin-packaging:
https://github.com/salesforcecli/plugin-packaging/blob/main/src/commands/force/package/beta/version/create.ts#L187

peternhale: will replace path.includes with fullPath.includes

cristiand391 (approved):

src/sfProject.ts
cristiand391: @peternhale since it's possible that `getPackageFromPath` doesn't return any package, plugin-packaging should handle that case too:

https://github.com/salesforcecli/plugin-packaging/blob/main/src/commands/force/package/beta/version/create.ts#L187-#L180

TS doesn't complain because the type is assiged on line 180 but if the dir passed isn't a package sfdx will throw something like `Cannot acccess package from undefined`

PR #700, opened 2022-11-15
feat: getConnection accepts an optional version

mshanemc:
### What does this PR do?
getConnection accepts an optional version
used for apiversion flag on sfCommand-based commands

@W-12068411@

### DIFF
--- a/src/org/org.ts
+++ b/src/org/org.ts
@@ -873,6 +873,8 @@ export class Org extends AsyncOptionalCreatable<Org.Options> {
 
   /**
    * Returns the JSForce connection for the org.
+   * side effect: If you pass it an apiVersion, it will set it on the Org
+   * so that future calls to getConnection() will also use that version.
    *
    * @param apiVersion The API version to use for the connection.
    */
--- a/src/org/org.ts
+++ b/src/org/org.ts
@@ -873,8 +873,13 @@ export class Org extends AsyncOptionalCreatable<Org.Options> {
 
   /**
    * Returns the JSForce connection for the org.
+   *
+   * @param apiVersion The API version to use for the connection.
    */
-  public getConnection(): Connection {
+  public getConnection(apiVersion?: string): Connection {
+    if (apiVersion && this.connection.getApiVersion() !== apiVersion) {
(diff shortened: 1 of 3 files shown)


peternhale (approved):

src/org/org.ts
```diff
@@ -873,8 +873,13 @@ export class Org extends AsyncOptionalCreatable<Org.Options> {
 
   /**
    * Returns the JSForce connection for the org.
+   *
+   * @param apiVersion The API version to use for the connection.
    */
-  public getConnection(): Connection {
+  public getConnection(apiVersion?: string): Connection {
```
peternhale, line 879: @mshanemc after noodling on this, my only reservation is the implicit permanent state change to the original connection. Maybe deep clone the connection, set the api version and return the clone.

peternhale (changes requested):

src/org/org.ts
mshanemc: Pete and agreed to make this clearer via jsdoc instead.

src/org/org.ts
```diff
@@ -873,8 +873,15 @@ export class Org extends AsyncOptionalCreatable<Org.Options> {
...
   /**
    * Returns the JSForce connection for the org.
+   * side effect: If you pass it an apiVersion, it will set it on the Org
+   * so that future calls to getConnection() will also use that version.
+   *
+   * @param apiVersion The API version to use for the connection.
    */
-  public getConnection(): Connection {
+  public getConnection(apiVersion?: string): Connection {
+    if (apiVersion && this.connection.getApiVersion() !== apiVersion) {
```
iowillhoit, line 882: Suggestion: Have this log if `getApiVersion` and `apiVersion` are the same. Help prevent folks from providing this unnecessarily. 

```js
if (apiVersion) {
     if (this.connection.getApiVersion() === apiVersion) {
          logger.warn(`Default API version is already ${apiVersion}`);
    } else { 
          this.connection.setApiVersion(apiVersion);
    }
}

iowillhoit (dismissed):

iowillhoit (approved):

peternhale (approved):

PR #707, opened 2022-11-23
fix: remove refs to legacy client id

peternhale:
@W-11659812@

Removes references to legacy client id and us current default client id


test/unit/org/authInfoTest.ts
```diff
@@ -153,7 +153,6 @@ describe('AuthInfo', () => {
         // verify the returned object doesn't have secrets
         expect(() => walkAndSearchForSecrets(toJsonMap(fields) || {}, decryptedRefreshToken)).to.not.throw();
 
-        expect(strObj).does.not.include(ensureString(testOrg.defaultConnectedAppInfo.clientSecret));
```
WillieRuemmele, line 156: why don't we need these assertions anymore? Because clientSecret = `''`?

WillieRuemmele (commented):
Changes seem safe. Hopefully we're far enough into the "distant future" to where we won't feel any effects

test/unit/org/authInfoTest.ts
peternhale: Yep

WillieRuemmele (dismissed):

shetzel (approved):

RodEsp (approved):

PR #712, opened 2022-11-30
Sh/connection perf

shetzel:
### What does this PR do?
Saves DNS resolution status and max API version per Connection instance.  The API version cache is used from more places within Connection.

@W-12129008@
https://github.com/forcedotcom/cli/issues/1713

### DIFF
--- a/src/org/connection.ts
+++ b/src/org/connection.ts
@@ -239,7 +239,7 @@ export class Connection<S extends Schema = Schema> extends JSForceConnection<S> 
    */
   public async retrieveMaxApiVersion(): Promise<string> {
     // Check saved value first, then cache.
-    if (this.maxApiVersion || (this.maxApiVersion = this.getCachedApiVersion())) {
+    if ((this.maxApiVersion ??= this.getCachedApiVersion())) {
       return this.maxApiVersion;
     }
 
--- a/src/org/connection.ts
+++ b/src/org/connection.ts
@@ -10,7 +10,7 @@
 import { URL } from 'url';
 import { AsyncResult, DeployOptions, DeployResultLocator } from 'jsforce/api/metadata';
 import { Duration, env, maxBy } from '@salesforce/kit';
-import { asString, ensure, isString, JsonCollection, JsonMap, Nullable, Optional } from '@salesforce/ts-types';
+import { asString, ensure, isString, JsonCollection, JsonMap, Optional } from '@salesforce/ts-types';
 import {
   Connection as JSForceConnection,
   ConnectionConfig,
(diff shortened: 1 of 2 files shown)


src/org/connection.ts
```diff
@@ -82,6 +82,12 @@ export class Connection<S extends Schema = Schema> extends JSForceConnection<S>
   // All connections are tied to a username
   private username!: string;
 
+  // Save the DNS resolution result of this connection's instance URL.
+  private resolvable!: boolean;
+
+  // Save the max API version of this connection's org.
+  private maxApiVersion!: string;
+
```
mshanemc, line 90: let's avoid the untrue TS assertions.  It's better to check for the definedness of these props each time than to guarantee they're always defined.

you could always set `resolvable` to false, then make it true you've resolved it once.  (maybe a better name is hasResolved?)

maxApiVersion could definitely be undefined.

mshanemc (changes requested):

src/org/connection.ts
```diff
@@ -235,18 +238,38 @@ export class Connection<S extends Schema = Schema> extends JSForceConnection<S>
...
    */
   public async retrieveMaxApiVersion(): Promise<string> {
+    if (this.maxApiVersion) {
+      return this.maxApiVersion;
+    }
+
+    // check API version cache
+    const cachedApiVersion = this.getCachedApiVersion();
+    if (cachedApiVersion) {
+      return cachedApiVersion;
```
mshanemc, line 248: let's store the value from cache into `this.maxApiVersion`

mshanemc:
QA Notes:

using sfdx-core linked into sdr linked into plugin source, running `DEBUG=* bin/dev source:deploy` so I can see logs from myDomainResolver and maxApiVersion stuff.

✅  extra calls go away and I see these instead
```
sf:connection DEBUG API version cache last checked on 11/30/2022, 4:22:27 PM (now is 11/30/2022, 4:40:04 PM) +58ms
sf:connection DEBUG Using cached API version: 56.0 +0ms
sf:connection DEBUG Connection created with apiVersion 56.0 +0ms
``` 
 set the timestamp on the AuthFile back a year (  "instanceApiVersionLastRetrieved": "11/30/2021, 4:43:51 PM". )
✅ I see resolver
✅ I see call to max url
✅ file gets updated

set the version on the AuthFile back to 54.
✅ it uses that version, no calls to maxApi or resolver

set the version back and set an old timestamp
✅ timestamp and version are updated as expected

set timestamp,version back and SFDX_IGNORE_API_VERSION_CACHE
❓ updates timestamp, cache but makes all the calls to maxApiVersion (expected) and resolver (_**not expected**_) each time
     -- after conferring with Steve, this'll be a fix in the source plugin to pass SDR a connection instead of a username

remove those 2 props from the AuthFile
✅ they were re-added and used on subsequent calls

remove only the timestamp
✅ it was re-added and used

--apiversion 50.0
✅ uses api version 50.0.  Checks max version from the file as expected.

mshanemc (approved):

PR #721, opened 2022-12-06
feat: backward compatible messages.createError

mshanemc:
### What does this PR do?
messages.createError will tinker with the error key/name so that it ends in Error.  That's cool, unless you're trying to avoid breaking changes to JSON error output.

now you can avoid this pattern where you have to "fix" the name.

https://github.com/salesforcecli/plugin-trust/blob/3d6123010f0fc50203e1064ccd238b43c725c614/src/commands/plugins/trust/verify.ts#L77

@W-12173611@

### DIFF
--- a/src/messages.ts
+++ b/src/messages.ts
@@ -548,14 +548,43 @@ export class Messages<T extends string> {
     exitCodeOrCause?: number | Error,
     cause?: Error
   ): SfError {
-    const structuredMessage = this.formatMessageContents('error', key, tokens, actionTokens);
-    return new SfError(
-      structuredMessage.message,
-      structuredMessage.name,
-      structuredMessage.actions,
-      exitCodeOrCause,
-      cause
-    );
+    const { message, name, actions } = this.formatMessageContents({
+      type: 'error',
+      key,
+      tokens,
+      actionTokens,
+    });
+    return new SfError(message, name, actions, exitCodeOrCause, cause);
+  }
+
+  /**
+   * SfError wants error to end with the word Error.  Use this to create errors while preserving their existing name (for compatibility reasons).
+   *
+   * @deprecated Use `createError` instead unless you need to preserver the error name to avoid breaking changes.
+   * `error.name` will be the upper-cased key, remove prefixed `error.` and will always end in Error.
+   * `error.actions` will be loaded using `${key}.actions` if available.
+   *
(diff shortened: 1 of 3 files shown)


src/messages.ts
```diff
@@ -548,14 +548,43 @@ export class Messages<T extends string> {
...
+      type: 'error',
+      key,
+      tokens,
+      actionTokens,
+    });
+    return new SfError(message, name, actions, exitCodeOrCause, cause);
+  }
+
+  /**
+   * SfError wants error to end with the word Error.  Use this to create errors while preserving their existing name (for compatibility reasons).
```
cromwellryan, line 561:
```suggestion
   * SfError wants error names to end with the suffix Error.  Use this to create errors while preserving their existing name (for compatibility reasons).
```

src/messages.ts
```diff
@@ -548,14 +548,43 @@ export class Messages<T extends string> {
...
+   *
+   * @deprecated Use `createError` instead unless you need to preserver the error name to avoid breaking changes.
```
cromwellryan, line 563:
```suggestion
   * @deprecated Use `createError` instead unless you need to preserve the error name to avoid breaking changes.
```

src/messages.ts
```diff
@@ -548,14 +548,43 @@ export class Messages<T extends string> {
...
+   * `error.name` will be the upper-cased key, remove prefixed `error.` and will always end in Error.
```
cromwellryan, line 564: > and will always end in Error
Is this still accurate given the purpose of this variation?

src/messages.ts
```diff
@@ -548,14 +548,43 @@ export class Messages<T extends string> {
...
+   * `error.actions` will be loaded using `${key}.actions` if available.
+   *
+   * @param key The key of the error message.
+   * @param tokens The error message tokens.
+   * @param actionTokens The action messages tokens.
+   * @param exitCodeOrCause The exit code which will be used by SfdxCommand or the underlying error that caused this error to be raised.
+   * @param cause The underlying error that caused this error to be raised.
+   */
+  public createErrorButPreserveName(
```
cromwellryan, line 573: I came here initially to show my appreciate for intention revealing function names 😄.

cromwellryan (commented):
Mostly some grammatical things...

RodEsp (approved):

PR #722, opened 2022-12-08
fix: add password property to

peternhale:
### What does this PR do?
Adds password property to MockTestOrgData
@W-12083436@

### DIFF
--- a/src/testSetup.ts
+++ b/src/testSetup.ts
@@ -961,6 +961,7 @@ export class MockTestOrgData {
   public isDevHub?: boolean;
   public isScratchOrg?: boolean;
   public isExpired?: boolean | 'unknown';
+  public password?: string;
 
   public constructor(id: string = uniqid(), options?: { username: string }) {
     this.testId = id;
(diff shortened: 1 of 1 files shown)


src/testSetup.ts
```diff
@@ -961,6 +961,7 @@ export class MockTestOrgData {
   public isDevHub?: boolean;
   public isScratchOrg?: boolean;
   public isExpired?: boolean | 'unknown';
+  public password?: string;
```
WillieRuemmele, line 964: is password really optional? Seems to be set in the constructor right?

peternhale: Should be. I will remove init from ctor

src/testSetup.ts
```diff
@@ -975,6 +976,7 @@ export class MockTestOrgData {
     this.accessToken = `${this.testId}/accessToken`;
     this.refreshToken = `${this.testId}/refreshToken`;
     this.redirectUri = 'http://localhost:1717/OauthRedirect';
+    this.password = Math.random().toString(36).substring(2, 15) + Math.random().toString(36).substring(2, 15);
```
mshanemc: It makes sense to have all the "always there" properties on the Mock, since you're going to need them anyway.

Would anyone ever want a MockTestOrgData without a password (ex: to UT the output of `org:display`)?  

If so, they would do this (remove it from the mock after it's created)
https://github.com/forcedotcom/sfdx-core/blob/7a8ff9aebb19226b2d3df8c5b4b008cf9aaca38e/test/unit/org/authInfoTest.ts

instead of adding it to the mock if the need it via the public password prop?

mshanemc (approved):
comments I was making were addressed already.

WillieRuemmele (approved):

PR #728, opened 2022-12-13
fix: provide way to stub user data

peternhale:
### What does this PR do?
Add stubUsers to testSetup
@W-12083436@

### DIFF
--- a/src/testSetup.ts
+++ b/src/testSetup.ts
@@ -13,7 +13,7 @@
 import { randomBytes } from 'crypto';
 import { EventEmitter } from 'events';
 import { tmpdir as osTmpdir } from 'os';
-import { join as pathJoin, basename } from 'path';
+import { basename, join as pathJoin } from 'path';
 import * as util from 'util';
 import { SinonSandbox, SinonStatic, SinonStub } from 'sinon';
 
@@ -42,7 +42,7 @@ import { SfError } from './sfError';
 import { SfProject, SfProjectJson } from './sfProject';
 import { CometClient, CometSubscription, Message, StreamingExtension } from './status/streamingClient';
 import { OrgAccessor, StateAggregator } from './stateAggregator';
-import { AuthFields, Org, SandboxFields } from './org';
+import { AuthFields, Org, SandboxFields, User, UserFields } from './org';
 import { SandboxAccessor } from './stateAggregator/accessors/sandboxAccessor';
 import { AliasGroup } from './config/aliasesConfig';
 import { Global } from './global';
@@ -127,6 +127,7 @@ export class TestContext {
     Config?: ConfigStub;
     SfProjectJson?: ConfigStub;
     TokensConfig?: ConfigStub;
+    OrgUsersConfig?: ConfigStub;
   } = {};
   /**
    * A record of stubs created during instantiation.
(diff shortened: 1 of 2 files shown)


src/testSetup.ts
```diff
@@ -304,6 +305,41 @@ export class TestContext {
...
+   * The key is the username of the admin user and it must be included in the users array in order to obtain the orgId key for the remaining users.
+   * The admin user is excluded from the users array.
+   *
+   */
+  public stubUsers(users: Record<string, MockTestOrgData[]>): Promise<void> {
+    const mockUsers = Object.values(users).flatMap((orgUsers) => orgUsers as unknown as UserFields[]);
+    const userOrgsMap = new Map(
+      Object.entries(users).map(([adminUsername, orgs]) => {
+        const adminOrg = orgs.find((org) => org.username === adminUsername);
+        if (!adminOrg) {
```
WillieRuemmele, line 321: NIT: could be ternary

src/testSetup.ts
```diff
@@ -304,6 +305,41 @@ export class TestContext {
...
+        }
+        return [adminOrg.orgId, { usernames: orgs.filter((org) => org.username !== adminUsername) }];
+      })
+    );
+
+    stubMethod(this.SANDBOX, User.prototype, 'retrieve').callsFake(
+      (username): Promise<UserFields | undefined> => Promise.resolve(mockUsers.find((org) => org.username === username))
+    );
+
+    // stubMethod(this.SANDBOX, ConfigFile.prototype, 'hasFile').callsFake((orgId: string) => userOrgsMap.has(orgId));
```
WillieRuemmele, line 332: no longer needed?

src/testSetup.ts
```diff
@@ -304,6 +305,41 @@ export class TestContext {
...
 
+  /**
+   * Stub salesforce user authorizations.
+   *
+   * @param users The users to stub.
+   * The key is the username of the admin user and it must be included in the users array in order to obtain the orgId key for the remaining users.
+   * The admin user is excluded from the users array.
+   *
+   */
+  public stubUsers(users: Record<string, MockTestOrgData[]>): Promise<void> {
```
WillieRuemmele, line 316: should this be async? or just return `void`?

src/testSetup.ts
```diff
@@ -701,8 +737,8 @@ const _testSetup = (sinon?: SinonStatic): TestContext => {
  *    $$.stubAliases({ 'myTestAlias': 'user@company.com' });
  *
  *    // Will use the contents set above.
- *    const username = (await StateAggregator.getInstance()).aliases.resolveUsername('myTestAlias');
- *    expect(username).to.equal('user@company.com');
+ *    const username = (await StateAggregator.getInstance()).aliases.resolveUsername("myTestAlias");
```
WillieRuemmele, line 740: fine with these changes, but should we keep these commented out UTs? Why not remove them?

src/testSetup.ts
peternhale: Goes along with others, although no awaits

src/testSetup.ts
peternhale: let me check

peternhale: @WillieRuemmele looks like jsdoc to me. I'll fix the `"`

src/testSetup.ts
WillieRuemmele: why follow the others if there's no `await`s? will you have to `await stubUsers` since it returns a promise?

src/testSetup.ts
```diff
@@ -304,6 +305,41 @@ export class TestContext {
...
+  /**
+   * Stub salesforce user authorizations.
+   *
+   * @param users The users to stub.
+   * The key is the username of the admin user and it must be included in the users array in order to obtain the orgId key for the remaining users.
+   * The admin user is excluded from the users array.
+   *
+   */
+  public stubUsers(users: Record<string, MockTestOrgData[]>): Promise<void> {
+    const mockUsers = Object.values(users).flatMap((orgUsers) => orgUsers as unknown as UserFields[]);
```
mshanemc, line 317: TS is trying to tell you something about these types.

```
Type 'Promise<MockTestOrgData | undefined>' is not assignable to type 'Promise<UserFields | undefined>'.
  Type 'MockTestOrgData | undefined' is not assignable to type 'UserFields | undefined'.
    Type 'MockTestOrgData' is missing the following properties from type 'UserFields': id, lastName, alias, timeZoneSidKey, and 5 more.
```

Should there be an type for UserMock that is a MockTestOrgData but has the additional required props?  Should user mocks be their own type?

How would you use this mock if you don't have these properties (ex: LastName) that aren't on  MockTestOrgData?

They're [in a method](https://github.com/forcedotcom/sfdx-core/blob/f9d69c55fc5261b980e937c06eda6b3cd8a25d6c/src/testSetup.ts#L1057) called `getMockUserInfo` but the typing on that isn't clear either.  

It's used like https://github.com/forcedotcom/sfdx-core/blob/f9d69c55fc5261b980e937c06eda6b3cd8a25d6c/test/unit/org/userTest.ts#L36

I'm with you, where I'd rather give the consumer the option to set props

mshanemc (changes requested):

src/testSetup.ts
peternhale: @mshanemc the getMockUserInfo function returns an object with property names that have leading capital letters. Not what I want. I want user fields (UserFields). I'll work to remove the as unknown.

WillieRuemmele (approved):

PR #758, opened 2023-01-24
Sm/message-transpilation

mshanemc:
### What does this PR do?
- make messags constructor's secret 3rd prop public/readonly
- ttypescript transformer for reading message files at compile time
- wireit
- fix 1 missing message found while testing a discarded version of the transformer

@W-12422637@

---
Nut failure explanation (h/t @mdonnalley for explaining all this)
List of everything in salesforceCLI with this problem (it's the only one)
https://github.com/search?q=org%3Asalesforcecli%20Messages.loadMessages(%27%40salesforce%2Fcore%27&type=code

The NUT wants to not hardcode messages (good)
It calls `loadMessages(@salesforce/core` without having done the ImportDirectory step.  This was previously working because the NUT ran the command and it had loaded messages.

We would not want runtime code to use this pattern (refer to messages without using the Import step and presume their existence as a side effect of other code having run).

I'm modifying the nut on the plugin-org sf-ification PR chain to prevent this from failing in the future, but this'll be a known and acceptable TF until that merges.

https://github.com/salesforcecli/plugin-org/pull/550


tsconfig.json
```diff
@@ -3,7 +3,8 @@
   "compilerOptions": {
     "outDir": "./lib",
     "resolveJsonModule": true,
-    "rootDir": "./src"
+    "rootDir": "./src",
+    "plugins": [{ "transform": "./messageTransformer/messageTransformer.ts" }]
```
mshanemc, line 7: this property is only used by `ttypescript` / `ttsc`

https://github.com/cevek/ttypescript

peternhale (dismissed):
Looks cool @mshanemc. Only question is can this be disabled to revert back to original behavior?

mshanemc:
> Looks cool @mshanemc. Only question is can this be disabled to revert back to original behavior?

`yarn build` runs the thing that causes this to happen.  So if you wanted to opt out of it, you'd change the build script to just use `tsc` instead of `ttsc` and it would build like it does today.

https://github.com/forcedotcom/sfdx-core/pull/758/files#diff-099767c07c6a6b08bd73ea23c58cb970e1c2ab1ad5fc0a680aa2b738bc434673R7

messageTransformer/messageTransformer.ts
```diff
@@ -0,0 +1,93 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
```
gbockus-sf, line 2: nit: 2023

mshanemc: oh, that means I probably need to bump something else.

mdonnalley (approved):

PR #764, opened 2023-02-02
fix: conf file stubs for read and readSync

peternhale:
### What does this PR do?
assigns correct config file methods to stub references in TestContext.
@W-12083545@

### DIFF
--- a/src/testSetup.ts
+++ b/src/testSetup.ts
@@ -576,8 +576,9 @@ export const stubContext = (testContext: TestContext): Record<string, SinonStub>
 
   // Mock out all config file IO for all tests. They can restore individually if they need original functionality.
   // @ts-ignore
-  stubs.configRead = testContext.SANDBOXES.CONFIG.stub(ConfigFile.prototype, 'readSync').callsFake(readSync);
-  stubs.configReadSync = testContext.SANDBOXES.CONFIG.stub(ConfigFile.prototype, 'read').callsFake(read);
+  stubs.configRead = testContext.SANDBOXES.CONFIG.stub(ConfigFile.prototype, 'read').callsFake(read);
+  // @ts-ignore
+  stubs.configReadSync = testContext.SANDBOXES.CONFIG.stub(ConfigFile.prototype, 'readSync').callsFake(readSync);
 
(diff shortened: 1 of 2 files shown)


mshanemc (dismissed):

peternhale:
@mshanemc I would like another look. I went ahead and enabled test-strict for tests, so ...

test/unit/lifecycleEventsTest.ts
```diff
@@ -156,7 +162,7 @@ describe('lifecycleEvents', () => {
   it('will use a newer version and transfer the listeners', () => {
     // the original
     const lifecycle = Lifecycle.getInstance();
-    lifecycle.on('test7', async (result) => {
+    lifecycle.on('test7', async (result: Record<string, unknown>) => {
```
mshanemc, line 165: `on` can have a generic that provides the event type

```suggestion
    lifecycle.on<Record<string, unknown>>('test7', async (result) => {
```

test/unit/config/configFileTest.ts
```diff
@@ -195,7 +195,7 @@ describe('Config', () => {
         config.unlinkSync();
         assert(false, 'should throw');
       } catch (e) {
-        expect(e.name).to.equal('TargetFileNotFound');
```
mshanemc, line 198: you can also use chai assert to assert that it is the error of the expected type.  It's bonus a useful test assertion *and* narrows TS so all the props are valid

`assert(e instanceOf Error)` (or could be SfError if you know it should be that)

test/unit/crypto/cryptoKeyFailuresTest.ts
```diff
@@ -76,11 +77,13 @@ if (os.platform() === 'darwin') {
       try {
         await shouldThrow(Crypto.create(), 'Should have thrown an SetCredentialError for Crypto.init()');
       } catch (err) {
-        expect(err.name).to.equal('SetCredentialError');
```
mshanemc, line 79: oh, this would be a good one to assert `instanceof SfError` so then you get the `actions` typing, etc.

mshanemc (approved):
approved with non-blocking comments.

mshanemc (approved):

PR #774, opened 2023-02-21
fix: improve error contents for device login

peternhale:
### What does this PR do?
Improves error content for errors encountered during device login
@W-12484720@

### DIFF
--- a/src/deviceOauthService.ts
+++ b/src/deviceOauthService.ts
@@ -48,8 +48,8 @@ async function makeRequest<T extends JsonMap>(options: HttpRequest): Promise<T> 
   const rawResponse = await new Transport().httpRequest(options);
   const response = parseJsonMap<T>(rawResponse.body);
   if (response.error) {
-    const errorDescription = response.error_description ? `(${response.error_description as string})` : '';
-    const error = response.error ? `(${response.error as string})` : 'Unknown';
+    const errorDescription = typeof response.error_description === 'string' ? response.error_description : ''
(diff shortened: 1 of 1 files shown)


WillieRuemmele (approved):

src/deviceOauthService.ts
```diff
@@ -48,7 +48,9 @@ async function makeRequest<T extends JsonMap>(options: HttpRequest): Promise<T>
   const rawResponse = await new Transport().httpRequest(options);
   const response = parseJsonMap<T>(rawResponse.body);
   if (response.error) {
-    const err = new SfError('Request Failed.');
+    const errorDescription = response.error_description ? `(${response.error_description as string})` : '';
+    const error = response.error ? `(${response.error as string})` : 'Unknown';
```
mshanemc, line 52: This is one solution if we can't put good typings on <T>
```suggestion
    const errorDescription = typeof response.error_description === 'string' ? response.error_description : ''
    const error = typeof response.error === 'string' ?  response.error : 'Unknown';
```

src/deviceOauthService.ts
```diff
@@ -48,7 +48,9 @@ async function makeRequest<T extends JsonMap>(options: HttpRequest): Promise<T>
...
+    const err = new SfError(`Request Failed: ${error} ${errorDescription}`);
     err.data = Object.assign(response, { status: rawResponse.statusCode });
```
mshanemc, line 54: prefer ...spread to avoid mutating the response

mshanemc (approved):
approved with comments

PR #810, opened 2023-04-10
feat: support stronger ciphers

shetzel:
### What does this PR do?
Supports stronger ciphers for encryption/decryption.

@W-12422652@

QA Notes
**MAKE A BACKUP OF YOUR ~/.sfdx DIRECTORY BEFORE DOING ANYTHING.  Use the backup to revert everything done during testing.**

Link this library to plugin-auth and plugin-org.
Enable Crypto debug output.  `export DEBUG=sf:crypto` and `export SF_LOG_LEVEL=trace`
You'll probably want to set `SF_USE_GENERIC_UNIX_KEYCHAIN=true` for initial testing.
Run commands from those plugins without SF_CRYPTO_V2 env var.
Run commands from those plugins with SF_CRYPTO_V2=false.
Run commands from those plugins with SF_CRYPTO_V2=true.
Move the ~/.sfdx dir so that a new key is generated, then:
Run commands from those plugins with SF_CRYPTO_V2=true.

See the doc in the WI for more usecases.


src/crypto/crypto.ts
```diff
@@ -165,28 +212,15 @@ export class Crypto extends AsyncOptionalCreatable<CryptoOptions> {
...
-        if (os.platform() === 'darwin' && !useGenericUnixKeychain) {
-          error.actions = [messages.getMessage('macKeychainOutOfSync')];
-        }
-        throw error;
-      }
-      return dec;
-    });
+    // When everything is v2, we can remove the else
+    if (this.isV2Crypto()) {
+      return this.decryptV2({ tag, iv, secret, tokens });
```
mshanemc, line 220: what's the need for passing iv/secret to the decrypt methods?

For example, decryptV2 could calculate them from the tokens.  If they both took the tokens, there wouldn't be all the conditionals here.

src/crypto/crypto.ts
```diff
@@ -29,6 +42,30 @@ const ENCRYPTED_CHARS = /[a-f0-9]/;
...
+let cryptoLogger: Logger;
+const getCryptoLogger = (): Logger => {
+  if (!cryptoLogger) {
+    cryptoLogger = Logger.childFromRoot('crypto');
+  }
+  return cryptoLogger;
+};
+
+type CryptoVersion = 'v1' | 'v2';
+let cryptoVersion: CryptoVersion;
```
mshanemc, line 54: you can derive this type from (existing code `keyof typeof IV_BYTES`) or using the proposed `VERSIONS` structure, `keyof typeof VERSIONS`

mshanemc (commented):
left some comments to get the thinking going.

src/crypto/crypto.ts
```diff
@@ -37,9 +74,16 @@ interface CredType {
...
+  tag: string;
+  iv: string;
+  secret: string;
+};
+type DecryptConfigV2 = DecryptConfigV1 & { tokens: string[] };
+
+const makeSecureBuffer = (password: string | undefined, encoding: 'utf8' | 'hex'): SecureBuffer<string> => {
   const newSb = new SecureBuffer<string>();
-  newSb.consume(Buffer.from(ensure(password), 'utf8'));
+  newSb.consume(Buffer.from(ensure(password), encoding));
```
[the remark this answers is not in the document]
cristiand391: I guess it was because of the keychain code where password is `password?: string`:
https://github.com/forcedotcom/sfdx-core/blob/9e32558d895cae448e834e275f3ffbcaa59a1628/src/crypto/crypto.ts#L57-L69

and in the else block, `SecureBuffer.value` returns `Optional<string>`.

@shetzel `makeSecureBuffer` isn't exported so maybe we can make the change to enforce password is a string where `makeSecureBuffer` is called

cristiand391 (approved):

cristiand391:
**QA notes**:

deleted the `sfdx` entry from keychain
and linked this branch of core to auth/org/user plugins

### fresh start with v2 key
backed up `.sfdx` and `.sf`, then deleted them.
`SF_CRYPTO_V2=true`

mac (native keychain):

✅ auth to a hub with `org login web`, access/refresh tokens in auth file are encrypted using new cipher.
✅ created scratch org and generated a pwd for it, tokens and pwd encrypted 
✅ AT and pwd from `org display` are correctly decrypted: used with curl to hit the scratch org limits endpoint/can auth with user/pwd
✅ auth with sfdxURL: grabbed it from the scratch org, logged out of all orgs and auth'd using it.
✅ JWT auth

linux and windows (generic keychain)
same results as on mac +
✅ `~/.sfdx/key.json` contains 64char key.

### existing v1 key + auth files with crypto v2 enabled

✅ sfdx-core users v1 crypto even with `SF_CRYPTO_V2=true`, auth/org commands work

PR #817, opened 2023-04-18
Sm/unified config aggregator

mshanemc:
> this is part of a set of PRs.  See https://salesforce.quip.com/kRtwAJN99VYE for explanation, sequence, and QA 

### What does this PR do?
one ConfigAggregator that handles old and new configs/envs.  See https://salesforce.quip.com/kRtwAJN99VYE for more details

besides this PR, this WI also requires a change to sf-plugins-core to always load ConfigAggregator but this sfdx-core PR needs to happen first

@W-13048680@


src/config/configAggregator.ts
```diff
@@ -165,13 +166,22 @@ export class ConfigAggregator extends AsyncOptionalCreatable<ConfigAggregator.Op
...
+   * If you use a deprecated property, a warning will be emitted and it will attempt to resolve the new property's value
    *
    * **Throws** *{@link SfError}{ name: 'UnknownConfigKeyError' }* An attempt to get a property that's not supported.
    *
    * @param key The key of the property.
    */
   public getPropertyValue<T extends AnyJson>(key: string): Optional<T> {
-    if (this.getAllowedProperties().some((element) => key === element.key)) {
+    const match = this.getAllowedProperties().find((element) => key === element.key);
+    if (match?.deprecated && match.newKey) {
```
peternhale, line 177: Appears that the warning will only be emitted when a deprecated key has a new key. What about deprecation if a config value that does not have al alternative? Should we emit the warning when one is using a config key that does not have an alternative?

src/config/envVars.ts
```diff
@@ -9,10 +9,12 @@ import { Dictionary, Nullable } from '@salesforce/ts-types';
 import { camelCase, snakeCase } from 'change-case';
 import { Env } from '@salesforce/kit';
 import { Messages } from '../messages';
-import { Global } from '../global';
+import { Lifecycle } from '../lifecycleEvents';
+// import { Global } from '../global';
```
peternhale, line 13: remove

peternhale (changes requested):

src/config/configAggregator.ts
mshanemc: we would not want to emit this particular warning/message.

I am not aware of any plans to deprecate and not replace a config

src/config/configAggregator.ts
```diff
@@ -137,7 +138,7 @@ export class ConfigAggregator extends AsyncOptionalCreatable<ConfigAggregator.Op
    *
    * @param key The config key.
    */
-  public static getValue(key: string): ConfigInfo {
+  public static getValue(key: string): ConfigInfo | undefined {
```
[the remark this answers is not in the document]
mshanemc: I think so.  Not in actual functionality but to a compiler, if they don't have skipLibChecks, then it might.

I could back this out and let the consumers stumbled into the potential `undefineds` the hard way, and save the type fix for the next major.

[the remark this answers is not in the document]
WillieRuemmele: I would consider that a breaking change, anything where you'd have to fix code to consume a new version 😦 

I wish we could introduce this now, but think we can only introduce type narrowing changes without a major 🙁 

we can leave a comment in there for now 🤷‍♂️ - could even link to this thread

mshanemc:
merging with known failures
1. plugin-settings because NUT changes there reflect these changes
2. plugin-templates.  I don't understand why the types are colliding.  For safety, I built/compiled/ran NUTs on that plugin with the `qa` version of sfdx-core and sf-plugins-core.  🟢

PR #822, opened 2023-04-25
fix: exclude status Del% in querySandboxProcess

jayree:
### What does this PR do?

Fix filter in querySandboxProcess function for sandboxes with status `Deleting` or `Deleted`. `'D'` is not a valid value for `Status` see: https://developer.salesforce.com/docs/atlas.en-us.api_tooling.meta/api_tooling/tooling_api_objects_sandboxprocess.htm

Due to an unknown reason a SOQL filter by `Status` doesn't seem work (Status: `Completed` equals to `'1'` instead of `Completed` in SOQL)

### DIFF
--- a/src/org/org.ts
+++ b/src/org/org.ts
@@ -1383,10 +1383,17 @@ export class Org extends AsyncOptionalCreatable<Org.Options> {
    * @private
    */
   private async querySandboxProcess(where: string): Promise<SandboxProcessObject> {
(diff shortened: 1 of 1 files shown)


shetzel (approved):

shetzel:
These changes were brought into https://github.com/forcedotcom/sfdx-core/pull/944
Thanks @jayree !!

PR #824, opened 2023-04-27
chore: strict null changes

WillieRuemmele:
### What does this PR do?

strict-nullifies new changes in `main`

[skip-validate-pr]


mshanemc (approved):
you *could* do a check that the request has `url` and `query` properties to get rid of some of the `as string` stuff (and possibly provide more useful exceptions).  

I don't know this code very well so it's hard to know if the types are just too loose OR if you really could get requests with those missing.

PR #834, opened 2023-05-11
Sm/handle unknown federated configs

mshanemc:
### What does this PR do?
one config (`RestDeploy`) is only in sfdx-core under its old name, but its "new name" equivalent moved moved to PDR's ConfigMeta https://github.com/salesforcecli/plugin-deploy-retrieve/blob/9043c4e82287d8b8ed6da3d91988d80dd4fdc994/src/configMeta.ts#L22

This makes ConfigAggregator a little smarter so it can fall back and not throw when it recognizes that scenario.

https://github.com/forcedotcom/cli/issues/2127
[@W-13189905@](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001RCgi0YAD/view)

QA: use the repro from the github issue 
you can yarn link this into plugin-signups and then plugins:link plugin-signups in your CLI

### DIFF
--- a/src/config/configAggregator.ts
+++ b/src/config/configAggregator.ts
@@ -207,9 +207,12 @@ export class ConfigAggregator extends AsyncOptionalCreatable<ConfigAggregator.Op
         }
       }
       return match;
-    } else {
-      throw messages.createError('unknownConfigKey', [key]);
     }
+    const matchFromNewKey = this.getAllowedProperties().find((element) => key === element.newKey);
+    if (matchFromNewKey) {
+      return matchFromNewKey;
+    }
+    throw messages.createError('unknownConfigKey', [key]);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


WillieRuemmele (dismissed):

WillieRuemmele (approved):

WillieRuemmele:
### QA Notes
---
link this into plugin-signups into @salesforce/cli
before:
```
 ➜  sf org list shape                     
Error (1): Unknown config name: org-metadata-rest-deploy.
```

after: ✅ 
```
 ➜  sf org list shape

=== Org Shapes
```
✅ : when setting `restDeploy=true`
✅ : `org-metadata-rest-deploy=false`
✅ : `SF_ORG_METADATA_REST_DEPLOY` env var

PR #842, opened 2023-05-26
Sm/aliases-without-config-file

mshanemc:
### What does this PR do?
stateAggregator.aliasAccessor without configFile.  
every set/unset creates an fs operation with file locking (new library dependency, [3PP](https://gus.lightning.force.com/lightning/r/ADM_Third_Party_Software__c/a0qEE0000007AaXYAU/view))
deprecate things that only AliasAccessor was using (aliasConfig, configGroup) to get to configFile
testSetup has special "mocks" for aliases now that they're not using configFile (they're actually real fs ops to the tmp dir)

this would be a breaking change to the public API because the inheritane chain from aliasAccessor contained a LOT of methods we aren't using and now it won't.

@W-12065678@ https://github.com/forcedotcom/cli/issues/1806


src/stateAggregator/accessors/aliasAccessor.ts
```diff
@@ -148,23 +203,87 @@ export class AliasAccessor extends AsyncOptionalCreatable {
...
+      }
+      this.aliasStore = fileContentsRawToAliasStore(await readFile(this.fileLocation, 'utf-8'));
+    } catch (e) {
+      if (e instanceof Error && 'code' in e && e.code === 'ENOENT') {
+        this.aliasStore = new Map<string, string>();
+        await mkdir(dirname(this.fileLocation), { recursive: true });
+        await this.saveAliasStoreToFile();
+        return;
+      }
+      if (useLock) await unlock(this.fileLocation);
```
shetzel, line 231: This is an edge case, but seems possible:
`lock` throws an error that is not an ENOENT (maybe an EPERM error) so the lock is never obtained.  In the catch we try to unlock and that throws an error, and is the one reported to the user.  The user wouldn't know why the lock couldn't be obtained.  To prevent this, we might need to use a try/catch/finally and the finally has a try/catch for the unlock.

mshanemc: I think putting the original `lock` outside the try should surface an error better.

PR #861, opened 2023-06-13
fix: make typescript a regular dep

mshanemc:
### What does this PR do?
because the messageTransformer uses TS, that needs to be a regular (nonDev) dep

https://github.com/salesforcecli/plugin-deploy-retrieve/actions/runs/5257572625/jobs/9500678747?pr=659

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,7 +238,7 @@ In your plugin or library,
 ```json
 {
   ...
-  "plugins": [{ "transform": "@salesforce/core", "import": "messageTransformer" }]
+  "plugins": [{ "transform": "@salesforce/core/lib/messageTransformer", "import": "messageTransformer" }]
 }
 ```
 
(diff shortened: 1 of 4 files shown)


cristiand391 (approved):
Approved with one question:

sfdx-core imports `sinon` in testSetup:
https://github.com/forcedotcom/sfdx-core/blob/main/src/testSetup.ts#L18

which is also shipped in the npm package, but it doesn't include `sinon` as a dependency.

Is the error at runtime being caused by this export?
https://github.com/forcedotcom/sfdx-core/blob/main/src/exported.ts#L13

I'm curious if that's the case then we could not include `typescript` as a dep and suggest to do a more specific import like we do for test setup:

```
import { TestContext, MockTestOrgData } from '@salesforce/core/lib/testSetup';
```

cristiand391 (approved):

PR #862, opened 2023-06-15
fix: identify sandboxes during authentication

shetzel:
### What does this PR do?
Enhances `AuthInfo.identifyPossibleScratchOrgs()` to also identify sandboxes.
Adds another public method to `Org` to allow querying the `SandboxProcess` sObject by org ID.

@W-11831532@

QA suggestions:
Create a sandbox from the UI.  Auth to it from the CLI.  Check the auth file for the org and that a sandbox specific file was created in the global dir.  Attempting to delete the sandbox should work, since it can now be properly identified as a sandbox during authentication.


src/org/authInfo.ts
```diff
@@ -419,7 +419,48 @@ export class AuthInfo extends AsyncOptionalCreatable<AuthInfo.Options> {
             logger.debug(`error updating auth file for ${orgAuthInfo.getUsername()}`, error);
           }
         } catch (error) {
-          logger.error(`Error connecting to devhub ${hubAuthInfo.username}`, error);
+          if (error instanceof Error && error.name === 'SingleRecordQuery_NoRecords' && fields.orgId) {
```
mshanemc, line 422: line 395, the `getDevHubAuthInfos` is going to filter out orgs that aren't devHubs.  We'd want a better function for filtering out anything we know is a scratch org or sandbox, and then check the remaining stuff.

The isDevHub is a good filter to making the list of orgs we'd want to ask about scratch orgs

src/org/authInfo.ts
```diff
@@ -419,7 +419,48 @@ export class AuthInfo extends AsyncOptionalCreatable<AuthInfo.Options> {
...
+                await orgAuthInfo.save({
+                  ...fields,
+                  isScratch: false,
+                  isSandbox: true,
+                });
+              } catch (err) {
+                logger.debug(`error updating auth file for: ${orgAuthInfo.getUsername()}`, err);
+              }
+
+              try {
```
mshanemc, line 439: would anything bad happen if you failed to update the AuthFile but then updated the SandboxConfig?

src/org/authInfo.ts
```diff
@@ -419,7 +419,48 @@ export class AuthInfo extends AsyncOptionalCreatable<AuthInfo.Options> {
...
+                  sandboxName: sbxProcess.SandboxName,
+                  sandboxProcessId: sbxProcess.Id,
+                  sandboxInfoId: sbxProcess.SandboxInfoId,
+                  timestamp: new Date().toISOString(),
+                } as SandboxFields;
+
+                const stateAggregator = await StateAggregator.getInstance();
+                stateAggregator.sandboxes.set(fields.orgId, sfSandbox);
+                logger.debug(`writing sandbox auth file for: ${orgAuthInfo.getUsername()} with ID: ${fields.orgId}`);
+                await stateAggregator.sandboxes.write(fields.orgId);
```
mshanemc, line 454: this seems safe from the ConfigFile bugs because it should only ever hit one of these.

Just be sure that the auth command is calling this `identifyPossibleScratchOrgs` NOT in parallel with anything else that might be touching the authFile.

mshanemc (changes requested):
I think it's worth solving the non-devhub + hasSandboxes filter problem.

src/org/authInfo.ts
shetzel: Making those changes.

src/org/authInfo.ts
shetzel: Good question.  That's most likely a bad state though so probably just abort the mission (`return;`) if there's an error updating the auth file.  I'll make that change.

src/org/authInfo.ts
shetzel: I checked all places I could find and nothing was done in parallel so hopefully it's ok.

src/org/authInfo.ts
```diff
@@ -392,21 +392,29 @@ export class AuthInfo extends AsyncOptionalCreatable<AuthInfo.Options> {
...
 
-    logger.debug('getting devHubs');
+    logger.debug('getting devHubs and prod orgs to identify scratch orgs and sandboxes');
 
     // TODO: return if url is not sandbox-like to avoid constantly asking about production orgs
     // TODO: someday we make this easier by asking the org if it is a scratch org
 
     const hubAuthInfos = await AuthInfo.getDevHubAuthInfos();
+    // Get a list of org auths that are known not to be devhubs, scratch orgs, or sandboxes.
+    const possibleProdOrgs = await AuthInfo.listAllAuthorizations(
```
mshanemc, line 405: since it's possible that a DevHub is also a ProdOrg, you could include DevHubs in the filter.

Then instead of having to `queryScratchOrg` each DevHub and then catch to ask about sandbox, you could ask about everything on the first round.

mshanemc (approved):

src/org/authInfo.ts
```diff
@@ -440,6 +456,59 @@ export class AuthInfo extends AsyncOptionalCreatable<AuthInfo.Options> {
...
+        // set the sandbox config value
+        const sfSandbox = {
+          sandboxUsername: fields.username,
+          sandboxOrgId: fields.orgId,
+          prodOrgUsername: possibleProdOrg.username,
+          sandboxName: sbxProcess.SandboxName,
+          sandboxProcessId: sbxProcess.Id,
+          sandboxInfoId: sbxProcess.SandboxInfoId,
+          timestamp: new Date().toISOString(),
+        } as SandboxFields;
```
mshanemc, line 498: is the `SandboxFields` type wrong that timestamp isn't part of it?  Can that be fixed, or at least leave a comment about why the type assertion is necessary?

mshanemc:
QA Notes:

TBD create a scratch org the normal way, then get password
TBD logout from scratch org
TBD re-auth to scratch org, is marked as scratch
TBD re-auth to an existing prod org, inspect auth file.  expect Not scratch/sbox
TBD create sbox the normal way
TBD logout from sbox, re-auth, is marked as sbox

mshanemc:
QA notes: merged based on demos

mshanemc (approved):

PR #874, opened 2023-06-26
Wr/unset alias config

WillieRuemmele:
### What does this PR do?
when `Org.delete` is called, it will now unset any config values that reference the org (username or alias) and will unset an alias keys that are set to the org's username
https://github.com/forcedotcom/cli/issues/2237
@W-13644800@

### DIFF
--- a/src/org/org.ts
+++ b/src/org/org.ts
@@ -1140,9 +1140,7 @@ export class Org extends AsyncOptionalCreatable<Org.Options> {
 
       // unset any configs referencing this org
       const config = await Config.create();
-      [...existingAliases, username]
-        .map((name) => config.getKeysByValue(name))
-        .map((keys) => keys.map((k) => config.unset(k)));
+      [...existingAliases, username].flatMap((name) => config.getKeysByValue(name)).map((key) => config.unset(key));
       await config.write();
 
       const activeScratchOrgRecordId = (
--- a/src/org/org.ts
+++ b/src/org/org.ts
@@ -1131,7 +1131,20 @@ export class Org extends AsyncOptionalCreatable<Org.Options> {
 
     try {
       const devHubConn = devHub.getConnection();
-      const username = this.getUsername();
+      const username = ensureString(this.getUsername());
+
+      // unset any aliases referencing this org
+      const stateAgg = await StateAggregator.getInstance();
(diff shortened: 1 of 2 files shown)


mshanemc (approved):

mshanemc:
QA: core PR branch yarn-linked into plugin-org, plugins-linked into sf

create scratch org w/ alias, assign it another alias.
org delete scratch (-o username)
❌  `This directory does not contain a valid Salesforce DX project`

-- I think you're going to need to either use ConfigAggregator for these OR handle there not being a local (project) config

mshanemc:
QA: core PR branch yarn-linked into plugin-org, plugins-linked into sf

create scratch org w/ alias WR3 and set-defualt, assign it 2nd alias WR4. 
Also set as global target-org (using second alias)
org delete scratch (-o username)
✅ auth file is deleted
✅ 2 aliases no longer appear in alias list
✅  sf config list doesn't show it

same test as above, but in a project.  Delete using -o username
✅ auth file is deleted
✅ 2 aliases no longer appear in alias list
✅  sf config list doesn't show it

same test, Delete using -o alias (alias is the second alias, not the one used for org create)
✅ auth file is deleted
✅ 2 aliases no longer appear in alias list
✅  sf config list doesn't show it

create and delete org with no alias and no config 
✅ looks good!

src/config/configAggregator.ts
```diff
@@ -334,6 +334,13 @@ export class ConfigAggregator extends AsyncOptionalCreatable<ConfigAggregator.Op
     return this.config;
   }
 
+  public async unsetByValue(key: string): Promise<void> {
+    this.localConfig?.getKeysByValue(key).map((k) => this.localConfig?.unset(k));
```
mshanemc, line 338: My only concern about this is that parallel org deletes (think `org list --clean`) could possibly muck up the config files because of the well-known concurrency bug.

it would be nice if we only did `write()` if we knew that config had been modified.  

`config.unset` return true if it does unset the thing, so you could do something like

```ts
const globalIsModified = this.globalConfig?.getKeysByValue(key).map((k) => this.globalConfig?.unset(k)).some(Boolean)
```

mshanemc (approved):

mshanemc:
QA: retested using tests from above, all still good in project and not in project

✅

PR #876, opened 2023-06-28
Sm/new logger

mshanemc:
> I *think* https://github.com/forcedotcom/sfdx-core/pull/882 will fix the plugin-auth NUT failure you see in this PR

### What does this PR do?
logger via pino instead of bunyan fork.

some useful background on pino: https://www.nearform.com/blog/pino7-0-0-pino-transport-worker_thread-transport/
docs: https://getpino.io/#/

many of the existing methods on Logger class become no-ops
getBunyanLogger will return a pino Logger.  It'll still work for some things (ex: trace(), debug()) but probably breaks if you're trying to do "fancy" things to the logger (add/remove streams/serializers/etc)

the log filtering is now its own reusable function so it can be re-used .  It's also quite a bit faster.

[@W-13586057@](https://gus.lightning.force.com/a07EE00001Tl80JYAR)

https://github.com/forcedotcom/cli/issues/2209
https://github.com/forcedotcom/cli/issues/2196 (haven't tested this one, but it's worth checking.  We don't swallow uncaught exceptions in the new logger)
https://github.com/forcedotcom/cli/issues/1706
https://github.com/forcedotcom/cli/issues/1699

and also 3 closed issues
https://github.com/forcedotcom/cli/issues/1928
https://github.com/forcedotcom/cli/issues/2198
https://github.com/forcedotcom/cli/issues/1408


src/logger/logger.ts
```diff
@@ -0,0 +1,670 @@
...
+import { Logger as PinoLogger, pino } from 'pino';
+import { Env } from '@salesforce/kit';
+import { isKeyOf, isString } from '@salesforce/ts-types';
+import { Global, Mode } from '../global';
+import { SfError } from '../sfError';
+import { unwrapArray } from '../util/unwrapArray';
+import { MemoryLogger } from './memoryLogger';
+import { cleanup } from './cleanup';
+
+/**
```
mshanemc, line 20: most of the bunyan-specific stuff here is deprecated.  Most become no-ops.

src/lifecycleEvents.ts
```diff
@@ -6,12 +6,12 @@
  */
 
 import { AnyJson, Dictionary } from '@salesforce/ts-types';
-import * as Debug from 'debug';
 import { compare } from 'semver';
 // needed for TS to not put everything inside /lib/src
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-ignore
 import * as pjson from '../package.json';
+import { Logger } from './logger/logger';
```
mshanemc, line 14: use Logger instead of debug for logging.  this allows debug to be removed

src/logger/filters.ts
```diff
@@ -0,0 +1,90 @@
+/*
```
mshanemc, line 1: I refactored out the "filters" from the original logger into its own function,
and then refactored it to speed things up a bit.

When the perf tests run, they log a LOT through to the fs.  That becomes a stream bottleneck, and backpressure eventually makes its way from fs => transport stream => worker  thread => logger => main process.  improving this redactor perf should keep the logger fast.

Screenshot 2023-06-29 at 2 33 50 PM

src/logger/transformStream.ts
```diff
@@ -0,0 +1,56 @@
+/*
```
mshanemc, line 1: this loads on a worker thread, and pino streams messages through it before pipelining them to the final destination (file, or pretty terminal output)

The exported types aren't good, and the pipeline doesn't know what precedes it (node.PipelineSource uses `any`) I tried to clean this up but couldn't get them all.

.github/workflows/test.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     needs: linux-unit-tests
     uses: salesforcecli/github-workflows/.github/workflows/unitTestsWindows.yml@main
   nuts:
-    needs: linux-unit-tests
+    # needs: linux-unit-tests
```
iowillhoit, line 18: Can this be added back?

src/logger/cleanup.ts
```diff
@@ -0,0 +1,58 @@
...
+ */
+import * as fs from 'node:fs';
+import { join } from 'node:path';
+import { Global } from '../global';
+import { Logger } from './logger';
+
+/**
+ * the odds of running are 1 in CLEAN_ODDS
+ * ex: CLEAN_ODDS=100 implies 1 in 100
+ * ex: CLEAN_ODDS=100 implies 1 in 1 (run every time)
```
iowillhoit, line 15: Should this be `CLEAN_ODDS=1`?

src/logger/logger.ts
```diff
@@ -0,0 +1,668 @@
...
+    .map((v: string) => v.toLowerCase());
+
+  // The sfdx root logger singleton
+  private static rootLogger?: Logger;
+
+  /**
+   * @deprecated.  Has no effect
+   */
+
+  public readonly logRotationCount = new Env().getNumber('SF_LOG_ROTATION_COUNT') ?? 2;
```
iowillhoit, line 231: We could support this by comparing it against `SF_LOG_ROTATION_PERIOD` in the `cleanup` method. Right now it does not look like users are able to configure keeping logs older than 7 days (ignoring the cleanup odds)

src/logger/logger.ts
iowillhoit: Now removed since we decided to bump major

iowillhoit (dismissed):

iowillhoit (approved):

PR #909, opened 2023-08-02
fix: use debug levels that are numbers from the env

mshanemc:
### What does this PR do?

the env used for logger level (SF_LOG_LEVEL) was used to accepting a number.  Found this while trying `--dev-debug` and not getting the expected result with the new logger
https://github.com/salesforcecli/cli/blob/7de1e6eab96d4a2d768d6720664d9644e4842177/src/flags.ts#L31

new function handles passed-in options and env scenarios where there might be a number, and where there migt be a number that doesn't match a real level (round up to the nearest real level) to change that to a string.

Also accepts string values ex: 'debug' | 'trace' etc.

[@W-13876266@](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001XVR1JYAX/view)

### DIFF
--- a/src/logger/logger.ts
+++ b/src/logger/logger.ts
@@ -507,8 +507,6 @@ const levelFromOption = (value?: string | number): string => {
       return numberToLevel(value);
     case 'string':
       return value;
-    case 'undefined':
-      return pino.levels.labels[Logger.DEFAULT_LEVEL];
     default:
       return pino.levels.labels[Logger.DEFAULT_LEVEL];
   }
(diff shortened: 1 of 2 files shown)


src/logger/logger.ts
```diff
@@ -488,3 +488,33 @@ const getWriteStream = (level = 'warn'): pino.TransportSingleOptions => {
...
+};
+
+const levelFromOption = (value?: string | number): string => {
+  switch (typeof value) {
+    case 'number':
+      return numberToLevel(value);
+    case 'string':
+      return value;
+    case 'undefined':
+      return pino.levels.labels[Logger.DEFAULT_LEVEL];
```
shetzel, line 511: nit: could remove this case since the default is the same.

shetzel (dismissed):

shetzel (approved):

PR #992, opened 2023-11-16
fix: set stack to fullStack if set

shetzel:
### What does this PR do?
`SfError.fullStack` is unknown by most (if not all) consumers of it.  The reason for creating an SfError with a caused by error is to add the cause error stack to the new stack.  Not sure what other reason there would be.  This change sets the `SfError.stack` to be the same as `SfError.fullStack`

[skip-validate-pr]

### DIFF
--- a/src/sfError.ts
+++ b/src/sfError.ts
@@ -72,6 +72,9 @@ export class SfError extends NamedError {
     } else {
       this.exitCode = 1;
     }
+    if (this.fullStack) {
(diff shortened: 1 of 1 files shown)


shetzel:
Closing in favor of a larger refactor effort of `SfError` to take advantage of the native `Error.cause`.  See also: https://salesforce-internal.slack.com/archives/G02K6C90RBJ/p1700164249793629

PR #1015, opened 2024-01-08
fix: handle html server response

shetzel:
### What does this PR do?
Handles an HTML response from the server and throws a helpful error.  Also, if the error is just a string or doesn't have the data property set it handles it as well.

@W-14756399@

### DIFF
--- a/src/deviceOauthService.ts
+++ b/src/deviceOauthService.ts
@@ -10,7 +10,7 @@
 import Transport from 'jsforce/lib/transport';
 import { AsyncCreatable, Duration, parseJsonMap } from '@salesforce/kit';
 import { HttpRequest, OAuth2Config } from 'jsforce';
-import { ensureString, JsonMap, Nullable } from '@salesforce/ts-types';
+import { ensureString, isString, JsonMap, Nullable } from '@salesforce/ts-types';
 import * as FormData from 'form-data';
 import { Logger } from './logger/logger';
 import { AuthInfo, DEFAULT_CONNECTED_APP_INFO } from './org/authInfo';
@@ -39,6 +39,12 @@ export interface DeviceCodePollingResponse extends JsonMap {
   issued_at: string;
 }
 
+interface DeviceCodeAuthError extends SfError {
+  error: string;
+  error_description: string;
+  status: number;
+}
+
 async function wait(ms = 1000): Promise<void> {
   return new Promise((resolve) => {
     setTimeout(resolve, ms);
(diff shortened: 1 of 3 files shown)


src/deviceOauthService.ts
```diff
@@ -39,6 +39,12 @@ export interface DeviceCodePollingResponse extends JsonMap {
   issued_at: string;
 }
 
+interface DeviceCodeAuthError extends SfError {
+  error: string;
+  error_description: string;
+  status: number;
+}
+
 async function wait(ms = 1000): Promise<void> {
```
mshanemc, line 48: "while we're here" suggestion: drop this function and use kit/sleep?

src/deviceOauthService.ts
```diff
@@ -47,7 +53,15 @@ async function wait(ms = 1000): Promise<void> {
 
 async function makeRequest<T extends JsonMap>(options: HttpRequest): Promise<T> {
   const rawResponse = await new Transport().httpRequest(options);
-  const response = parseJsonMap<T>(rawResponse.body);
+  let response: T;
+  if (rawResponse?.headers && rawResponse.headers['content-type'] === 'text/html') {
+    response = {
```
mshanemc, line 58: you could throw in the case of html error, rather than try to shoehorn it into T but with an error/error_description to cause it to throw on line 70.

as written, it would still barf the whole html page to the user (html => error_description => SfError message => final `throw` from `poll`).

When I did the jsforce fix for this in http-api, I put them html content to the logs.  You could, if throwing an SfError, put it in the `data` property if you think users might want it when using `--json`.

src/deviceOauthService.ts
```diff
@@ -175,21 +189,21 @@ export class DeviceOauthService extends AsyncCreatable<OAuth2Config> {
     );
     try {
       return await makeRequest<DeviceCodePollingResponse>(httpRequest);
-    } catch (e) {
-      /* eslint-disable @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/restrict-template-expressions */
-      const err: any = (e as SfError).data;
-      if (err.error && err.status === 400 && err.error === 'authorization_pending') {
+    } catch (e: unknown) {
+      const err = (
+        e instanceof SfError ? e.data : SfError.wrap(isString(e) ? e : 'unknown').data
```
mshanemc, line 194: checking the type `as`sertion logic:
in the case of string error, 
  .data would be `undefined`
  so then err would be `undefined`

Then the type for `err` should be `DeviceCodeAuthError | undefined`.

[you've got all the `?` in place to handle that properly, I'm thinking of the next person who has to touch this and TS lets them refer to `err` without them realizing that it *could* be undefined and then the dishonest `as`sertion leads to a runtime TypeError ]

src/deviceOauthService.ts
```diff
@@ -39,15 +39,23 @@ export interface DeviceCodePollingResponse extends JsonMap {
...
+interface DeviceCodeAuthError extends SfError {
+  error: string;
+  error_description: string;
+  status: number;
 }
 
 async function makeRequest<T extends JsonMap>(options: HttpRequest): Promise<T> {
   const rawResponse = await new Transport().httpRequest(options);
+
+  if (rawResponse?.headers && rawResponse.headers['content-type'] === 'text/html') {
```
mshanemc, line 51:
```suggestion
  if (rawResponse?.headers?.['content-type'] === 'text/html') {
```

mshanemc (approved):

mshanemc:
QA: using `yarn:link` of this branch into plugin-auth

```
./bin/dev.js org login device --instance-url https://computing-connect-6970-dev-ed.scratch.my.salesforce.com/services/data/v50.0
Error (1): HTTP response contains html content. Check that the org exists and can be reached.
```
✅ good human error message
✅ json output contains the full html response in its data property.
Screenshot 2024-01-09 at 1 46 25 PM

❓  exitCode is `1`.

mshanemc (approved):

PR #1026, opened 2024-02-13
Mz/make bundleable

mingxuanzhangsfdx:
### What does this PR do?
make the code bundle-able without affecting the existing workflow.

@W-15022712@
https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001jzB1UYAU/view

### DIFF
--- a/examples/test/tsconfig.json
+++ b/examples/test/tsconfig.json
@@ -4,6 +4,7 @@
   "compilerOptions": {
     "noEmit": true,
     "skipLibCheck": true,
-    "resolveJsonModule": true
+    "resolveJsonModule": true,
+    "esModuleInterop": true
   }
 }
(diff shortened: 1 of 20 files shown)


peternhale (approved):

examples/test/tsconfig.json
```diff
@@ -4,6 +4,7 @@
   "compilerOptions": {
     "noEmit": true,
     "skipLibCheck": true,
-    "resolveJsonModule": true
+    "resolveJsonModule": true,
+    "esModuleInterop": true
```
mingxuanzhangsfdx, line 8: 'esModuleInterop' makes typescript compiler treat commonjs similar to ES6 modules. 

I made the changes in all tsconfig.json. This way we abandon usage like `import * as Faye from 'faye';` which esbuild does not support for commonjs module. Even if there is no change of code here in examples/test, it is still good to use it.

src/deviceOauthService.ts
```diff
@@ -11,7 +11,7 @@ import Transport from 'jsforce/lib/transport';
 import { AsyncCreatable, Duration, parseJsonMap, sleep } from '@salesforce/kit';
 import { HttpRequest, OAuth2Config } from 'jsforce';
 import { ensureString, isString, JsonMap, Nullable } from '@salesforce/ts-types';
-import * as FormData from 'form-data';
+import FormData from 'form-data';
```
mingxuanzhangsfdx, line 14: All changes like this are to make the import statement compatible with esbuild without hurting the existing compiling process of typescript compiler.

PR #1030, opened 2024-02-19
chore: esbuild artifacts

mingxuanzhangsfdx:
### What does this PR do?
This PR creates a github workflow to make esbuild transpile the project and then publish a package @salesforce/core-bundle to npm.

It should be merged after https://github.com/forcedotcom/sfdx-core/pull/1026

[W-15022848](https://gus.lightning.force.com/a07EE00001jzD09YAE)

@W-15022848@

### DIFF
--- a/.github/workflows/esbuild-publish.yml
+++ b/.github/workflows/esbuild-publish.yml
@@ -1,0 +1,48 @@
+name: esbuild Compilation & npm Publish Workflow
+
+on:
+  workflow_dispatch:
+    inputs:
+      branch:
+        description: 'Set the branch to use for automation tests'
+        type: string
+        required: false
+        default: 'main'
+      nodeVersion:
+        description: version of node to use.  It's better to specify latest, lts/* or lts/-1 than to hardode numbers
+        type: string
+        default: lts/*
+        required: false
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.branch }}
+      - uses: actions/setup-node@v4
+        with:
+          node-version: ${{ inputs.nodeVersion }}
+          registry-url: 'https://registry.npmjs.org'
+      - name: Install esbuild Dependencies
+        run: |
+          yarn add esbuild@^0.19.5 -D
+          yarn add esbuild-plugin-pino@^2.1.0 -D
+          yarn add npm-dts@^1.3.12 -D
+          yarn add esbuild-plugin-tsc@^0.4.0 -D
+      - name: Install All Dependencies
+        run: |
+          yarn install --network-timeout 600000
+      - name: Update for Bundling
+        run: |
+          node scripts/updateForBundling.js
+      - name: Generate Bundle
+        run: |
+          node scripts/build.js
+      - name: Publish a Package
+        run: |
+          npm config set //registry.npmjs.org/:_authToken=$NODE_AUTH_TOKEN
+          npm publish
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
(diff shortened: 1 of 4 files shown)


.github/workflows/esbuild-publish.yml
```diff
@@ -0,0 +1,48 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.branch }}
+      - uses: actions/setup-node@v4
+        with:
+          node-version: ${{ inputs.nodeVersion }}
+          registry-url: 'https://registry.npmjs.org'
+      - name: Install esbuild Dependencies
```
mingxuanzhangsfdx, line 28: These dependencies are required by esbuild

scripts/updateForBundling.js
```diff
@@ -0,0 +1,67 @@
...
+    try {
+      const packageJson = JSON.parse(data);
+
+      // Update package name if necessary
+      if (packageJson.name && packageJson.name === '@salesforce/core') {
+        packageJson.name = '@salesforce/core-bundle';
+      }
+
+      // Remove 'prepack' and 'prepare' scripts
+      if (packageJson.scripts) {
```
mingxuanzhangsfdx, line 22: These two fields are deleted because it will block the publishing process of esbuild.

scripts/updateForBundling.js
```diff
@@ -0,0 +1,67 @@
...
+      console.error(`Error reading package.json: ${err}`);
+      return;
+    }
+
+    try {
+      const packageJson = JSON.parse(data);
+
+      // Update package name if necessary
+      if (packageJson.name && packageJson.name === '@salesforce/core') {
+        packageJson.name = '@salesforce/core-bundle';
```
mingxuanzhangsfdx, line 18: produce a separate package called core-bundle.

.github/workflows/esbuild-publish.yml
```diff
@@ -0,0 +1,48 @@
...
+        run: |
+          yarn add esbuild@^0.19.5 -D
+          yarn add esbuild-plugin-pino@^2.1.0 -D
+          yarn add npm-dts@^1.3.12 -D
+          yarn add esbuild-plugin-tsc@^0.4.0 -D
```
mshanemc, line 33: you *could* just do 1 add, like

yarn add -D esbuild@^0.19.5 esbuild-plugin-pino@^2.1.0 etc.

.github/workflows/esbuild-publish.yml
```diff
@@ -0,0 +1,48 @@
...
+      - name: Install All Dependencies
+        run: |
+          yarn install --network-timeout 600000
```
mshanemc, line 36: within our actions we have https://github.com/salesforcecli/github-workflows/blob/main/.github/actions/yarnInstallWithRetries/action.yml which has retry logic wrapped around it.  That reduces network/cache errors from yarn

mshanemc (approved):

.github/workflows/esbuild-publish.yml
```diff
@@ -0,0 +1,48 @@
...
+      - name: Update for Bundling
+        run: |
+          node scripts/updateForBundling.js
+      - name: Generate Bundle
+        run: |
+          node scripts/build.js
+      - name: Publish a Package
+        run: |
+          npm config set //registry.npmjs.org/:_authToken=$NODE_AUTH_TOKEN
```
cristiand391, line 45: I don't think this will work, but if it does make sure the token isn't included in the logs. 

IIRC, we use `actions/setup-node` which writes `NODE_AUTH_TOKEN` to the npmrc file.

mingxuanzhangsfdx: I tried the .npmrc method first but it did not work properly. I am sure it won't display in the logs. Here is an example in my forked repo. https://github.com/mingxuanzhangsfdx/sfdx-core/actions/runs/7955097070/job/21713519127#step:10:1

mshanemc (approved):

PR #1031, opened 2024-02-22
feat: add support for refreshing sandboxes

shetzel:
### What does this PR do?

- Allows refreshing sandboxes.
- Handles updating auth files of refreshed sandboxes.
- Fixes a bug where authenticating after successful sandbox status polling did not update the auth file to be a sandbox, and would therefore not be properly identified as a sandbox by `org list`.

@W-15042183@

### DIFF
--- a/src/org/org.ts
+++ b/src/org/org.ts
@@ -111,6 +111,19 @@ export type SandboxProcessObject = {
   ApexClassId?: string;
   EndDate?: string;
 };
+const sandboxProcessFields = [
+  'Id',
+  'Status',
+  'SandboxName',
+  'SandboxInfoId',
+  'LicenseType',
+  'CreatedDate',
+  'CopyProgress',
+  'SandboxOrganization',
+  'SourceId',
+  'Description',
+  'EndDate',
+];
 
 export type SandboxRequest = {
   SandboxName: string;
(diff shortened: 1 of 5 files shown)


src/org/org.ts
```diff
@@ -227,6 +261,62 @@ export class Org extends AsyncOptionalCreatable<Org.Options> {
...
+      async: false,
+      interval: Duration.seconds(30),
+    }
+  ): Promise<SandboxProcessObject> {
+    this.logger.debug(sandboxInfo, 'RefreshSandbox called with SandboxInfo');
+    const refreshResult = await this.connection.tooling.update('SandboxInfo', sandboxInfo);
+    this.logger.debug(refreshResult, 'Return from calling tooling.update');
+
+    if (!refreshResult.success) {
+      throw messages.createError('sandboxInfoRefreshFailed', [JSON.stringify(refreshResult)]);
```
mdonnalley, line 284: is `refreshResult` guaranteed to be JSON? Just wondering if we get something unexpected back from the server (like HTML) then it would be an ugly message for the user.

mdonnalley (approved):

src/org/org.ts
shetzel: Good question.  As long as this uses the jsforce version that catches unexpected HTML responses to SOAP requests it should throw a much better error.

jshackell-sfdc (approved):

PR #1035, opened 2024-03-07
fix: re-enable code challenge support

shetzel:
### What does this PR do?
CLI auth should default to using code challenges and code verifiers for PKCE.  This used to be the case but somehow it was disabled.  This adds it back and also adds unit tests to ensure it stays that way.

@W-15198289@

QA:
All auth that happens as part of NUTs will verify that it doesn't break anything.  To ensure that it works with an org that has PKCE set, open your org, to to setup, search for "oauth", click on "OAuth and OpenID Connect Settings", toggle the "Require Proof Key for Code Exchange (PKCE) Extension" on.  Then do `sf org login web` targeting that org.


mdonnalley (approved):

mdonnalley:
**QA**

- linked sfdx-core to plugin-org, linked plugin-auth to sf
- Enabled `Require Proof Key for Code Exchange (PKCE) Extension` on org
✅  `sf org login web`

PR #1038, opened 2024-03-13
add properties for PS/PSL package metadata access to package dir

KSCCO:
### What does this PR do?
Update PackageDir to include packageMetadataAccess properties that can be set in sfdx-project.json to allow a package developer to assign permission sets and permission set licenses to the build org user.

The equivalent schema change was made in PR: 
https://github.com/forcedotcom/schemas/pull/82

@W-14842023@

### DIFF
--- a/src/sfProject.ts
+++ b/src/sfProject.ts
@@ -39,6 +39,10 @@ export type PackageDir = {
   dependencies?: PackageDirDependency[];
   includeProfileUserLicenses?: boolean;
   package?: string;
+  packageMetadataAccess?:  {
+     permissionSets: string | string[];
+     permissionSetLicenses: string | string[];
+  };
   path: string;
   postInstallScript?: string;
   postInstallUrl?: string;


src/sfProject.ts
```diff
@@ -39,6 +39,10 @@ export type PackageDir = {
   dependencies?: PackageDirDependency[];
   includeProfileUserLicenses?: boolean;
   package?: string;
+  packageMetadataAccess?:  {
+     permissionSets: string | string[];
+     permissionSetLicenses: string | string[];
```
shetzel, line 44: Can it just be `string[]` instead of both?  It's easier to code for an array versus an array or string.  You have to do type checks wherever you use it rather than just treat it as an array.  Also, are both required when `packageMetadataAccess` is defined or are they optional?

KSCCO: I think it can just be string[]. I'll verify and make the change.

I believe these combinations make sense for inclusion in packageMetadataAccess:
PermissionSet & PermissionSetLicense
PermissionSet (only)

Defining only PermissionSetLicense without a PermissionSet probably does not make sense since it won't actually grant any permissions to the user. However, there shouldn't be any harm in allowing it. And I could even believe there might be some reason for wanting to do so that's not obvious to me.

shetzel: Based on DMs with Kevin, let's just keep it like it is for consistency with the other permset properties (ApexTestAccess)

shetzel (approved):

PR #1040, opened 2024-03-21
fix(auth-server): handle preflight requests

cristiand391:
### What does this PR do?

Updates the auth web server to handle preflight requests.

Chrome v123 (promoted to stable 2 days ago) now does a preflight request (`OPTIONS` req) to the server before the auth request, the server only handled `GET` request and was throwing on anything else, blocking the web auth flow.

repro:

1) update chrome to v123
2) try auth with `sf org login web`, type auth creds in browser and click on login
3) see failure on browser and `sf` exits with `Invalid request method: OPTIONS` error.

on the network tab you can see the preflight request that causes the failure:
Screenshot 2024-03-22 at 14 35 40

testing:
link this branch into plugin-auth and follow the repro steps, the preflight request is handled correctly and auth can follow.

@W-15313611@
https://github.com/forcedotcom/cli/issues/2785


mshanemc (approved):

src/webOAuthServer.ts
```diff
@@ -399,6 +405,22 @@ export class WebServer extends AsyncCreatable<WebServer.Options> {
...
+  /**
+   * Preflight request:
+   *
+   * https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request
+   * https://www.w3.org/TR/2020/SPSD-cors-20200602/#resource-preflight-requests
+   */
+  public handlePreflightRequest(response: http.ServerResponse): void {
+    // We don't validate the origin here because:
+    // 1. The default login URL (login.salesforce.com) will not match after a redirect or if user choose a custom domain in login.
+    // 2. There's no fixed list of auth URLs we could check against.
```
cristiand391, line 417: point 1 refers to the oauth config object that can be accessed here.

> If the value of the [Origin](https://www.w3.org/TR/2020/SPSD-cors-20200602/#http-origin) header is not a [case-sensitive](https://www.w3.org/TR/2020/SPSD-cors-20200602/#case-sensitive) match for any of the values in [list of origins](https://www.w3.org/TR/2020/SPSD-cors-20200602/#list-of-origins) do not set any additional headers and terminate this set of steps.
Always matching is acceptable since the [list of origins](https://www.w3.org/TR/2020/SPSD-cors-20200602/#list-of-origins) can be unbounded.

WillieRuemmele:
### QA Notes
---
> I was unable to reproduce this on an M1 mac with chrome 123...
> I hoped on a call with Cristian, and we walked through it together

✅ : able to `auth web login` to an org
✅ : OPTIONS request => 204

PR #1046, opened 2024-04-04
feat: improved sfError

mshanemc:
### What does this PR do?
- bump tsconfig target/lib/etc for es2022 so we get native error.cause
- sfError extends Error instead of kit/NamedError
  - adds the `name` property  
  - adds the fullStack method...which now operates on stack of `error.cause` by recursing through them to build its "caused by" 
  - add a new `.create` method that's less painful than the 5-prop overloaded constructor and which supports the `context` and `data` properties
  - wrap method takes an `unknown` so you can wrap things in a `catch` block without having to worry about their type (previously only supported `string | Error | SfError`).  Will wrap any "other" things in an error to preserve its info in the new error's `cause`  
  - the constructor (arg5) and the new `.create` also allow `cause` to be `unknown` and will narrow it to an error (and otherwise throw) for that same convenience in catch blocks.
  - `code` property always returns a string [it's either code (which can only be a string) or name (also required, string)] and now the types say that

@W-14507125@

### DIFF
--- a/src/sfError.ts
+++ b/src/sfError.ts
@@ -123,12 +123,16 @@ export class SfError<T extends ErrorDataProperties = ErrorDataProperties> extend
     const sfError =
       err instanceof Error
         ? // a basic error with message and name.  We make it the cause to preserve any other properties
-          new SfError<T>(err.message, err.name, undefined, err)
+          SfError.create<T>({
+            message: err.message,
+            name: err.name,
+            cause: err,
+          })
         : // ok, something was throws that wasn't error or string.  Convert it to an Error that preserves the information as the cause and wrap that.
-          SfError.wrap<T>(
-            new TypeError(`SfError.wrap received type ${typeof err} but expects type Error or string`, { cause: err })
-          );
-
+          SfError.create<T>({
+            message: `SfError.wrap received type ${typeof err} but expects type Error or string`,
+            cause: err,
+          });
     // If the original error has a code, use that instead of name.
     if (hasString(err, 'code')) {
       sfError.code = err.code;
--- a/src/sfError.ts
+++ b/src/sfError.ts
@@ -14,7 +14,7 @@ export type SfErrorOptions<T extends ErrorDataProperties = ErrorDataProperties> 
   data?: T;
   cause?: Error;
   context?: string;
-  actions: string[];
+  actions?: string[];
 };
 
 type ErrorDataProperties = AnyJson;
--- a/src/sfError.ts
+++ b/src/sfError.ts
@@ -129,10 +129,10 @@ export class SfError<T extends ErrorDataProperties = ErrorDataProperties> extend
             cause: err,
           })
         : // ok, something was throws that wasn't error or string.  Convert it to an Error that preserves the information as the cause and wrap that.
-          SfError.create<T>({
-            message: `SfError.wrap received type ${typeof err} but expects type Error or string`,
-            cause: err,
-          });
+          SfError.wrap<T>(
+            new Error(`SfError.wrap received type ${typeof err} but expects type Error or string`, { cause: err })
+          );
+
     // If the original error has a code, use that instead of name.
     if (hasString(err, 'code')) {
       sfError.code = err.code;
--- a/src/sfError.ts
+++ b/src/sfError.ts
@@ -5,8 +5,19 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
-import { NamedError } from '@salesforce/kit';
 import { AnyJson, hasString, isString, JsonMap } from '@salesforce/ts-types';
+
+export type SfErrorOptions<T extends ErrorDataProperties = ErrorDataProperties> = {
+  message: string;
+  exitCode?: number;
+  name?: string;
+  data?: T;
+  cause?: Error;
+  context?: string;
+  actions: string[];
+};
+
+type ErrorDataProperties = AnyJson;
 
 /**
  * A generalized sfdx error which also contains an action. The action is used in the
@@ -24,7 +35,8 @@ import { AnyJson, hasString, isString, JsonMap } from '@salesforce/ts-types';
  * throw new SfError(message.getMessage('myError'), 'MyErrorName');
  * ```
  */
-export class SfError<T = unknown> extends NamedError {
+export class SfError<T extends ErrorDataProperties = ErrorDataProperties> extends Error {
+  public readonly name: string;
   /**
    * Action messages. Hints to the users regarding what can be done to fix related issues.
    */
@@ -59,19 +71,25 @@ export class SfError<T = unknown> extends NamedError {
    */
   public constructor(
     message: string,
-    name?: string,
+    name = 'SfError',
     actions?: string[],
     exitCodeOrCause?: number | Error,
     cause?: Error
   ) {
-    cause = exitCodeOrCause instanceof Error ? exitCodeOrCause : cause;
-    super(name ?? 'SfError', message || name, cause);
+    const derivedCause = exitCodeOrCause instanceof Error ? exitCodeOrCause : cause;
+    super(message);
+    this.name = name;
+    this.cause = derivedCause;
     this.actions = actions;
     if (typeof exitCodeOrCause === 'number') {
       this.exitCode = exitCodeOrCause;
     } else {
       this.exitCode = 1;
     }
+  }
+
+  public get fullStack(): string | undefined {
+    return recursiveStack(this).join('\nCaused by: ');
   }
 
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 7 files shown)


src/sfError.ts
```diff
@@ -5,9 +5,20 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
-import { NamedError } from '@salesforce/kit';
 import { AnyJson, hasString, isString, JsonMap } from '@salesforce/ts-types';
 
+export type SfErrorOptions<T extends ErrorDataProperties = ErrorDataProperties> = {
```
mshanemc, line 10: type for the new `.create` method param obj

src/sfError.ts
```diff
@@ -5,9 +5,20 @@
...
+  message: string;
+  exitCode?: number;
+  name?: string;
+  data?: T;
+  cause?: Error;
+  context?: string;
+  actions?: string[];
+};
+
+type ErrorDataProperties = AnyJson;
```
mshanemc, line 20: constrain the data Prop...see the `"DANGER"` note on the original's line 145

src/sfError.ts
```diff
@@ -24,7 +35,8 @@ import { AnyJson, hasString, isString, JsonMap } from '@salesforce/ts-types';
  * throw new SfError(message.getMessage('myError'), 'MyErrorName');
  * ```
  */
-export class SfError<T = unknown> extends NamedError {
+export class SfError<T extends ErrorDataProperties = ErrorDataProperties> extends Error {
```
mshanemc, line 38: leave NamedError out of this

src/sfError.ts
```diff
@@ -24,7 +35,8 @@ import { AnyJson, hasString, isString, JsonMap } from '@salesforce/ts-types';
...
+  public readonly name: string;
```
mshanemc, line 39: brought from NamedError

src/sfError.ts
```diff
@@ -74,6 +88,10 @@ export class SfError<T = unknown> extends NamedError {
     }
   }
 
+  public get fullStack(): string | undefined {
```
mshanemc, line 91: this existed on NamedError.  Now it recurses through the `cause` prop's error chain

mshanemc: https://github.com/forcedotcom/kit/blob/963b1d1f34833fbd889a89f18ed82194aeb7f5a5/src/errors.ts#L34

src/sfError.ts
```diff
@@ -83,26 +101,39 @@ export class SfError<T = unknown> extends NamedError {
...
-  public static wrap(err: Error | string): SfError {
+  public static wrap<T extends ErrorDataProperties = ErrorDataProperties>(err: unknown): SfError<T> {
     if (isString(err)) {
-      return new SfError(err);
+      return new SfError<T>(err);
     }
 
     if (err instanceof SfError) {
-      return err;
+      return err as SfError<T>;
```
mshanemc, line 122: <T> only matters if you want to type the .data prop

test/tsconfig.json
```diff
@@ -5,6 +5,10 @@
     "noEmit": true,
     "skipLibCheck": true,
     "resolveJsonModule": true,
-    "esModuleInterop": true
+    "esModuleInterop": true,
+    "lib": ["ES2022"],
```
mshanemc, line 9: necessary to get the native error.cause

mshanemc: we already set engines to node18+ but now we type/compile for it

test/unit/sfErrorTest.ts
```diff
@@ -70,6 +133,55 @@ describe('SfError', () => {
...
       expect(mySfError).to.equal(existingSfError);
     });
+
+    describe('handling "other" stuff that is not Error', () => {
+      it('undefined', () => {
+        const wrapMe = undefined;
+        const mySfError = SfError.wrap(wrapMe);
+        expect(mySfError).to.be.an.instanceOf(SfError);
+        expect(mySfError.message === 'An unexpected error occurred');
+        expect(mySfError.name === 'TypeError');
```
WillieRuemmele, line 143: Does it _really_ make sense than an `SfError.wrap(undefined).name==='TypeError'`?

WillieRuemmele (commented):
it all seems like a lot of usability improvements

src/sfError.ts
```diff
@@ -83,26 +101,39 @@ export class SfError<T = unknown> extends NamedError {
...
     }
 
-    const sfError = new SfError(err.message, err.name, undefined, err);
+    const sfError =
+      err instanceof Error
+        ? // a basic error with message and name.  We make it the cause to preserve any other properties
+          new SfError<T>(err.message, err.name, undefined, err)
+        : // ok, something was throws that wasn't error or string.  Convert it to an Error that preserves the information as the cause and wrap that.
+          SfError.wrap<T>(new TypeError('An unexpected error occurred', { cause: err }));
```
shetzel, line 130: Why `TypeError` and not just `Error`?

shetzel (approved):

src/sfError.ts
cristiand391: this method was only used (at least in `salesforcecli`) by sfdxCommand:
https://github.com/search?q=org%3Asalesforcecli%20%2F%5C.fullStack%2F&type=code

we could remove it and recommend to use `util.inspect(err)`, it checks `error.cause` and pretty-print the stacktrace while also removing duplicates:
https://nodejs.org/dist/latest-v20.x/docs/api/errors.html#errorcause

```js
const util = require('node:util')
const badServer = new Error('Bad server')
const cause = new Error('The remote HTTP server responded with a 500 status', { cause: badServer });
const symptom = new Error('The message failed to send', { cause });

console.log(util.inspect(symptom));
```
```
Error: The message failed to send
    at Object.<anonymous> (/Users/cdominguez/code/gh/sf/command/index.js:4:17)
    at Module._compile (node:internal/modules/cjs/loader:1376:14)
    ... 4 lines matching cause stack trace ...
    at node:internal/main/run_main_module:28:49 {
  [cause]: Error: The remote HTTP server responded with a 500 status
      at Object.<anonymous> (/Users/cdominguez/code/gh/sf/command/index.js:3:15)
      at Module._compile (node:internal/modules/cjs/loader:1376:14)
      ... 4 lines matching cause stack trace ...
      at node:internal/main/run_main_module:28:49 {
    [cause]: Error: Bad server
        at Object.<anonymous> (/Users/cdominguez/code/gh/sf/command/index.js:2:19)
        at Module._compile (node:internal/modules/cjs/loader:1376:14)
        at Module._extensions..js (node:internal/modules/cjs/loader:1435:10)
        at Module.load (node:internal/modules/cjs/loader:1207:32)
        at Module._load (node:internal/modules/cjs/loader:1023:12)
        at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:135:12)
        at node:internal/main/run_main_module:28:49
  }
}
```

test/unit/sfErrorTest.ts
mshanemc: I think so.  You're passing something that is of an unexpected type and we can't do much with it.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError
> A TypeError may be thrown when:
> an operand or argument passed to a function is incompatible with the type expected by that operator or function; or

What would you prefer?

src/sfError.ts
mshanemc: see above comment about TypeError.  Sounds like it's confusing to 2 people now.

I'd argue it's fairly idiomatic in TS (typically, type-checking an any/unknown)
see https://github.com/search?q=throw+new+TypeError+language%3ATypeScript&type=code&l=TypeScript

src/sfError.ts
mshanemc: yeah, I think I like that.  I did a broader github search and couldn't find anything else using it, either.

I didn't know about `util.inspect`... that's cool

src/sfError.ts
```diff
@@ -59,13 +72,17 @@ export class SfError<T = unknown> extends NamedError {
...
     actions?: string[],
     exitCodeOrCause?: number | Error,
-    cause?: Error
+    cause?: unknown
   ) {
-    cause = exitCodeOrCause instanceof Error ? exitCodeOrCause : cause;
-    super(name ?? 'SfError', message || name, cause);
+    if (typeof cause !== 'undefined' && !(cause instanceof Error)) {
+      throw new TypeError(`The cause, if provided, must be an instance of Error. Received: ${typeof cause}`);
+    }
```
cristiand391, line 82: related to the WI about correctly bubbling up the erro from iso-git > STL > SDR:

STL throws an `SfError` with data in it, SDR will wrap it and rethrow (new SfError...) but you have to remember to call `error.setData(originalError)` or the data is lost. 
Could we check here in the constructor if `cause instanceof SfError` and save possible `cause.data` in the new error?

mshanemc: a "parent" error could call `setData` or mutating the data prop and losing the `cause` data.

but, it's a good idea.

src/sfError.ts
shetzel: To me, a TypeError means that the author of the code made a mistake and someone needs to fix a client side bug.  Usually this would mean the CLI team.  W.r.t "plugin health" it would be categorized as a bug for the CLI team (if called from the sf CLI).  If this is intended, then it's fine.  If the meaning of the error is anything else, we should just use `Error`.  Although, since it's immediately being wrapped by SfError the command framework probably wouldn't interpret this as a TypeError and would categorize it differently.  In the end it probably doesn't matter much.

src/sfError.ts
mshanemc: but maybe this is the least bad option. 
```
 this.data = resolvedCause.data as T;
```

PR #1050, opened 2024-04-08
Sm/core7-jsforce

mshanemc:
### What does this PR do?
swap jsforce 

@W-15387805@


package.json
```diff
@@ -1,13 +1,75 @@
...
   "types": "lib/index.d.ts",
   "license": "BSD-3-Clause",
   "engines": {
     "node": ">=18.0.0"
   },
+  "exports": {
+    ".": {
+      "types": "./lib/index.d.ts",
+      "require": "./lib/index.js",
+      "import": "./lib/index.js",
```
cristiand391, line 15: node docs says these 2 should be mutually exclusive:

> "import" - matches when the package is loaded via import or import(), or via any top-level import or resolve operation by the ECMAScript module loader. Applies regardless of the module format of the target file. Always mutually exclusive with "require".

https://nodejs.org/api/packages.html#conditional-exports

but here it's fine because sfdx-core isn't ESM-yet, right?

cristiand391 (approved):
approved with one question about conditional exports

package.json
mshanemc: it's a good question.  They're all cjs for now--if we did decide to dual-publish these, we'd have `require` and `import` pointing at different files.

It worked with both an ESM (PDR) and non-ESM (SDR) like this.  Maybe it's overkill unless we dual-publish?

shetzel: It feels a bit like overkill since there's nothing conditional about it currently and it's 60 extra lines in package.json.

src/deviceOauthService.ts
```diff
@@ -7,9 +7,9 @@
 /* eslint-disable camelcase */
 /* eslint-disable @typescript-eslint/ban-types */
 
-import Transport from 'jsforce/lib/transport';
+import Transport from '@jsforce/jsforce-node/lib/transport';
```
shetzel, line 10: Can we change this to some publicly exported Class/Function in jsforce-node?  If not, do we need to make a change in jsforce-node?

src/org/authInfo.ts
```diff
@@ -26,8 +26,8 @@ import {
   Nullable,
   Optional,
 } from '@salesforce/ts-types';
-import { OAuth2Config, OAuth2, TokenResponse } from 'jsforce';
-import Transport from 'jsforce/lib/transport';
+import { OAuth2Config, OAuth2, TokenResponse } from '@jsforce/jsforce-node';
+import Transport from '@jsforce/jsforce-node/lib/transport';
```
shetzel, line 30: Same comment as for `deviceOauthService`.  Also, seems like we can combine multiple lines of imports into 1 statement.

shetzel (approved):

PR #1064, opened 2024-05-03
fix: remove namespace if nonamespace option is set

mdonnalley:
### What does this PR do?

Remove `namespace` from ScratchOrgInfo payload if the `nonamespace` option is true

@W-15675379@
https://github.com/forcedotcom/cli/issues/2855

### DIFF
--- a/src/org/scratchOrgInfoGenerator.ts
+++ b/src/org/scratchOrgInfoGenerator.ts
@@ -37,7 +37,7 @@ export type ScratchOrgInfoPayload = {
   package2AncestorIds: string;
   features: string | string[];
   connectedAppConsumerKey: string;
-  namespace: string;
+  namespace?: string;
   connectedAppCallbackUrl: string;
   durationDays: number;
 } & PartialScratchOrgInfo;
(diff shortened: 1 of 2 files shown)


src/org/scratchOrgInfoGenerator.ts
```diff
@@ -216,12 +216,16 @@ export const generateScratchOrgInfo = async ({
     // project is not required
   }
 
+  const { namespace: originalNamespace, ...payload } = scratchOrgInfoPayload;
+
+  const namespace = originalNamespace ?? sfProject?.get('namespace');
```
cristiand391, line 221: so, `originalNamespace` comes from the definition file and takes precedence over `namespace` defined in sfdx-project.json. 

it's doc'd for sfdx-project json but not for the definition file here
https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_scratch_orgs_def_file.htm

I'll ask Juliet if we should add a note about it in docs.

cristiand391 (approved):

cristiand391:
**QA notes**:

repro'd the issue with sf v2.40.7, running:
```
sf org create scratch -d -f config/project-scratch-def.json -a dreamhouse --duration-days 1 --no-namespace
```
and specifying the namespace in the scratch def. file, then `sf org list --verbose` to view the namespace in the scratch org.

`--no-namespace` works if you specify the namespace in `sfdx-project.json`.

✅ namespace in scratch org def. file, `--no-namespace` not passed, org created with namespace
✅ namespace in `sfdx-project.json`, `--no-namespace` not passed, org created with namespace
✅ namespace in scratch org def. file, `--no-namespace` passed, org created without namespace
✅ namespace in `sfdx-project.json`, `--no-namespace` passed, org created without namespace

PR #1079, opened 2024-06-05
fix: add a did you mean action to AuthInfo

WillieRuemmele:
### What does this PR do?
adds a "did you mean" suggestion to the `SfError` thrown when we can't find a passed in username/alias when constructing an `AuthInfo` class.

@W-15921235@


src/org/authInfo.ts
```diff
@@ -813,7 +814,16 @@ export class AuthInfo extends AsyncOptionalCreatable<AuthInfo.Options> {
     }
     // If a username with NO oauth options, ensure authorization already exist.
     else if (username && !authOptions && !(await this.stateAggregator.orgs.exists(username))) {
-      throw messages.createError('namedOrgNotFound', [username]);
+      throw SfError.create({
+        name: 'NamedOrgNotFoundError',
+        message: messages.getMessage('namedOrgNotFound', [username]),
+        actions: [
+          `It looks like you mistyped the username or alias. Did you mean "${findSuggestion(username, [
```
mdonnalley, line 821: Could it be misleading to tell the user that they mistyped the username or alias? What happens if they provide a username or alias that simply doesn't exist? 

Maybe simplifying it to just `Did you mean "<suggestion>"?` would be the most flexible

Also do we need to remove the action if no suggestion is found?

src/org/authInfo.ts
```diff
@@ -813,7 +814,16 @@ export class AuthInfo extends AsyncOptionalCreatable<AuthInfo.Options> {
...
+            ...(await this.stateAggregator.orgs.list()).map((f) => f.split('.json')[0]),
```
mdonnalley, line 822: is splitting the filename better than getting the `username` prop from the contents? Or can we not guarantee that that property will be there? More curious than anything

mdonnalley (changes requested):
See QA

src/org/authInfo.ts
WillieRuemmele: Juliet and I had some back and forth on this. We thought the current messaging was friendlier, but we also had exactly what you suggested 🤷‍♂️ 

I'll remove it if nothing is passed, or we couldn't deduce any similar names

src/org/authInfo.ts
WillieRuemmele: yeah, username is optional in the contents

src/org/authInfo.ts
mdonnalley: I'll happily defer to Juliet on the message 👍

mdonnalley:
QA #2

🟢 no aliases and no orgs, `sf org display -o foo`

```
Error (2): Parsing --target-org
	No authorization information found for foo.
See more help with --help
```

PR #1081, opened 2024-06-07
fix: sfProject types matches schemas repo

mshanemc:
### What does this PR do?
uses the types from salesforce/schemas for sfProject and its PackageDirs.

This might be a breaking change because consumers might have been using looser types
- or packaging lib, which uses types in crazy ways https://github.com/forcedotcom/packaging/pull/590

pino 9 (major version bump) drops support for older node version.  Shouldn't affect us.

@W-10271718@
@W-5571825@


src/sfProject.ts
```diff
@@ -760,3 +724,22 @@ export class SfProject {
       .map(([key]) => key);
   }
 }
+
+/** differentiate between the Base PackageDir (path, maybe default) and the Packaging version (package and maybe a LOT of other fields) by whether is has the `package` property */
+export const isPackagingDirectory = (packageDir: PackageDir): packageDir is PackagePackageDir =>
```
mshanemc, line 729: the are exported to help consumers use the types safely.

You can tell from the amount of ` as ` in packagingLib that the types weren't helping.

package.json
```diff
@@ -53,9 +53,9 @@
   ],
   "dependencies": {
```
WillieRuemmele, line 54: is this a major version bump?  I don't see it in in commits with `!` or in the version here?

src/sfProject.ts
```diff
@@ -67,16 +35,10 @@ export type NamedPackageDir = PackageDir & {
...
-  packageDirectories: PackageDir[];
-  namespace?: string;
-  sourceApiVersion?: string;
-  sfdcLoginUrl?: string;
-  signupTargetLoginUrl?: string;
-  oauthLocalPort?: number;
-  plugins?: { [k: string]: unknown };
-  packageAliases?: { [k: string]: string };
-};
+export type NamedPackagingDir = PackagePackageDir & NamedDirAdditions;
```
WillieRuemmele, line 38: What's the difference between "package" and "packaging" - would be nice to have comments here explaining

NIT: `NamedDirAdditions` also unreasonably iirks me, maybe just `NameAndPath` so that it becomes `PackagePackageDir & NameAndPath`?

src/sfProject.ts
```diff
@@ -762,3 +713,26 @@ export class SfProject {
...
+/** differentiate between the Base PackageDir (path, maybe default) and the Packaging version (package and maybe a LOT of other fields) by whether is has the `package` property */
+export const isPackagingDirectory = (packageDir: PackageDir): packageDir is PackagePackageDir =>
+  isPackagingDir(packageDir);
+
+/** differentiate between the Base PackageDir (path, maybe default) and the Packaging version (package and maybe a LOT of other fields) by whether is has the `package` property */
+export const isNamedPackagingDirectory = (packageDir: NamedPackageDir): packageDir is NamedPackagingDir =>
+  isPackagingDir(packageDir);
+
+const isPackagingDir = (packageDir: PackageDir | NamedPackageDir): boolean =>
+  'package' in packageDir && typeof packageDir.package === 'string';
```
WillieRuemmele, line 726: isn't `typeof packageDir.package` always a string?

WillieRuemmele (commented):
would like clarification around the anticipated release (is it major)

I'd also like more comments around the types, and when they should apply... but since we're probably the only ones using them, maybe it's a moot point 🤷‍♂️

src/sfProject.ts
mshanemc: it could be 'undefined'

src/sfProject.ts
mshanemc: in sfdx-project.json they're called `packageDirectories` even if you're not using them for packaging.  `Packaging` refers to 1gp/2gp packages

I like you idea but will use `NameAndFullPath` since path is already on the basic one

package.json
mshanemc: yeah, there's a `feat!` commit in there with BREAKING CHANGES: etc.

mshanemc: https://github.com/forcedotcom/sfdx-core/commit/66083f7154a4b9da8f351375150e2c2c5b494bdc

src/sfProject.ts
WillieRuemmele: but it's typed as a required string? 
image


just checking `&& packageDir.package` should work too right? 🤷‍♂️

WillieRuemmele (approved):

MIGRATING_V7-V8.md
```diff
@@ -0,0 +1,15 @@
+# Migrating `@salesforce/core` from v7 to v8
+
+v8 uses the types from `@salesforce/schemas`, for SfProject/SfProjectJson.
+
+The schemas types are more accurate about the PackageDir type (what `sfdx-project.json` packageDirectories is an array of). It is a union type of
+
+1. simply `path`, with an optional default property
+1. 1, along with the `package` property and lots of optional properties used for packaging.
```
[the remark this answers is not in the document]
mshanemc: doesn't have to be.  markdown renders them correctly like this, and the you can change the order without retyping them.

PR #1092, opened 2024-06-20
fix: prevent validations on encrypted values

mdonnalley:
### What does this PR do?

Prevent validation from being re-run on encrypted values

**Testing**

Create a plugin using `sf dev generate plugin` and add this command: 

```typescript
import { SfCommand, Flags } from '@salesforce/sf-plugins-core';
import { Config } from '@salesforce/core';

// eslint-disable-next-line sf-plugin/command-example
export default class StoreToken extends SfCommand<void> {
  // eslint-disable-next-line sf-plugin/no-hardcoded-messages-commands
  public static readonly summary = 'Programmatically set a token in the global config';

  public static readonly flags = {
    token: Flags.string({
      required: true,
      // eslint-disable-next-line sf-plugin/no-hardcoded-messages-flags
      summary: 'a token to store',
    }),
  };

  public async run(): Promise<void> {
    const { flags } = await this.parse(StoreToken);

    const config = await Config.create({ isGlobal: true, encryptedKeys: ['local-web-server-identity-token'] });
    Config.addAllowedProperties([
      {
        key: 'local-web-server-identity-token',
        description: 'Token for local web server identity',
        encrypted: true,
        input: {
          validator: (value): boolean => typeof value === 'string' && value.startsWith('123'),
          failedMessage: 'Token must start with 123',
        },
      },
    ]);
    config.set('local-web-server-identity-token', flags.token);

    await config.write();

    this.log('Done!');
  }
}
```

Without this change, it will fail validation since the `validator` is rerun during write (which calls `this.cryptProperties(true)`, which calls `set`, which is was calls all the validators)

With the change, the validator is called just once during the initial set for encrypted properties

@W-16049112@

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -420,6 +420,16 @@ export class Config extends ConfigFile<ConfigFile.Options, ConfigProperties> {
     return keyBy(Config.allowedProperties, 'key');
   }
 
+  private static findProperty(key: string): ConfigPropertyMeta {
+    const property = Config.allowedProperties.find((allowedProp) => allowedProp.key === key);
+
+    if (!property) {
+      throw messages.createError('unknownConfigKey', [key]);
+    }
+
+    return property;
+  }
+
   /**
    * Read, assign, and return the config contents.
    */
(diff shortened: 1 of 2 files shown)


mshanemc (approved):

mshanemc:
error replicates using QA as described

✅ --token abc still fails 
✅ global config file has an encrypted value

PR #1099, opened 2024-07-09
util.format list issue

iowillhoit:
### What does this PR do?

> [!NOTE]
> We decided to postpone this change to collect information via telemetry. Once all missing specifiers have been corrected, we will enforce them. Follow up work will be completed in [@W-16197665@](https://gus.my.salesforce.com/apex/ADM_WorkLocator?bugorworknumber=W-16197665)

> [!CAUTION]
> This is technically a breaking change. If we have a message somewhere that is missing a specifier, the tokens will no longer be rendered. This is a difficult thing to validate across our codebase. We could do a few things:
> - Wait until the next major release of `sfdx-core`
> - Add telemetry for when a specifier is missing and monitor it over the coming weeks
> - Only skip formatting if `messages` has a length greater than `1`

This is a solution to prevent tokens from being rendered in a message when a specifier (e.g. `%s`) does not exist

If the message does not contain a specifier, node's `util.format` still appends the token to the end ([docs](https://nodejs.org/api/util.html#utilformatformat-args:~:text=If%20there%20are%20more%20arguments%20passed%20to%20the%20util.format()%20method%20than%20the%20number%20of%20specifiers%2C%20the%20extra%20arguments%20are%20concatenated%20to%20the%20returned%20string%2C%20separated%20by%20spaces))

This issue was discovered in `sf whatsnew --help` where the token for the specifier `%s` was being rendered at the end of each example. This was happening because the `markdownLoader` automatically splits bulleted lists into arrays.

 We now check for the presence of a specifier and only format the message if one is present.

### BEFORE:
Screenshot 2024-07-09 at 11 08 10 AM

### AFTER:
Screenshot 2024-07-09 at 11 08 39 AM

### Testing
- Pull branch
- `yarn build`
- `yarn link`
- `cd` to plugin-info
- `yarn link @salesforce/core`
- `bin/run.js whatsnew --help`

[@W-16188160@](https://gus.my.salesforce.com/apex/ADM_WorkLocator?bugorworknumber=W-16188160)

### DIFF
--- a/src/messages.ts
+++ b/src/messages.ts
@@ -588,8 +588,16 @@ export class Messages<T extends string> {
     }
     const messages = ensureArray(msg);
     return messages.map((message) => {
-      ensureString(message);
-      return util.format(message, ...tokens);
+      const msgStr = ensureString(message);
+      // If the message does not contain a specifier, util.format still appends the token to the end.
+      // The 'markdownLoader' automatically splits bulleted lists into arrays.
+      // This causes the token to be appended to each line regardless of the presence of a specifier.
+      // Here we check for the presence of a specifier and only format the message if one is present.
+      // https://nodejs.org/api/util.html#utilformatformat-args
+      // https://regex101.com/r/8Hf8Z6/1
+      const specifierRegex = new RegExp('%[sdifjoO]{1}', 'gm');
+
+      return specifierRegex.test(msgStr) ? util.format(msgStr, ...tokens) : msgStr;
     });
   }
 }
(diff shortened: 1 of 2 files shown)


src/messages.ts
```diff
@@ -588,8 +588,16 @@ export class Messages<T extends string> {
...
     return messages.map((message) => {
-      ensureString(message);
-      return util.format(message, ...tokens);
+      const msgStr = ensureString(message);
+      // If the message does not contain a specifier, util.format still appends the token to the end.
+      // The 'markdownLoader' automatically splits bulleted lists into arrays.
+      // This causes the token to be appended to each line regardless of the presence of a specifier.
+      // Here we check for the presence of a specifier and only format the message if one is present.
+      // https://nodejs.org/api/util.html#utilformatformat-args
+      // https://regex101.com/r/8Hf8Z6/1
```
mdonnalley, line 597: Having a link to the regex is brilliant 💡

mdonnalley (dismissed):

mdonnalley:
QA

🟢 tokenized input only occurs once in markdown list

iowillhoit:
Here is an example console.logging a pseudo telem event (ignore the Error, it is unrelated)
Screenshot 2024-07-09 at 1 44 30 PM

PR #1108, opened 2024-07-22
fix: wrap-error-like-objects

mshanemc:
### What does this PR do?
spreading an error, like https://github.com/salesforcecli/plugin-deploy-retrieve/blob/af41a48180f944e01f1c5e0b8ad79148b89c90c2/src/commands/project/deploy/quick.ts#L132 will pass a param that is not an instance of SfError which cause SfError.wrap to wrap it in another error.  The human output doesn't show the wrapped error, which makes it harder for users to know what's wrong.

https://github.com/forcedotcom/cli/issues/2962
@W-16297490@

### DIFF
--- a/src/sfError.ts
+++ b/src/sfError.ts
@@ -136,17 +136,12 @@ export class SfError<T extends ErrorDataProperties = ErrorDataProperties> extend
     }
 
     const sfError =
-      err instanceof Error
-        ? // a basic error with message and name.  We make it the cause to preserve any other properties
-          SfError.create<T>({
-            message: err.message,
-            name: err.name,
-            cause: err,
-          })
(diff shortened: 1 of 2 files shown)


mdonnalley (approved):

mdonnalley:
**QA**

setup:
- checkout and build branch
- link to PDR
- link PDR to sf
- cd to dreamhouse-lwc and create scratch org
- `sf project deploy validate --source-dir force-app/main/default/classes`
- `sf project deploy start --source-dir force-app/main/default/classes`
- `sf project deploy quick --job-id 0AfO300000EhvJAKAZ` (id from `validate` output)

🟢 fails with expected error
```
Error (CannotQuickDeployError): Job ID can't be used for quick deployment. Possible reasons include the deployment hasn't been validated, has already been deployed, or the validation expired because you ran it more than 10 days ago.
```

PR #1113, opened 2024-07-29
feat(generateScratchOrgInfo): allow shema key

alanjaouen:
### What does this PR do?

This pr remove $schema key from parsed scratchorg definition. allowing to pass check `scratchOrgInfoPayload must be heads down camelcase.` and so preventing `InvalidJsonCasing` exceptions

`InvalidJsonCasing` are raised if definition file contains `$schema` key (from [documentation](https://forcedotcom.github.io/schemas/))

### DIFF
--- a/src/org/scratchOrgInfoGenerator.ts
+++ b/src/org/scratchOrgInfoGenerator.ts
@@ -322,6 +322,8 @@ const parseDefinitionFile = async (definitionFile: string): Promise<Record<strin
   try {
     const fileData = await fs.readFile(definitionFile, 'utf8');
     const defFileContents = parseJson(fileData) as Record<string, unknown>;
+    // remove key '$schema' from the definition file
+    delete defFileContents['$schema'];
     return defFileContents;
   } catch (err) {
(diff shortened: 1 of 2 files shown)


mshanemc:
Cool, thanks for noticing and fixing.

Are you willing/able to add a unit test for that?  It should probably go in this `describe` block https://github.com/forcedotcom/sfdx-core/blob/e8b0980455c73c1e17fe90b4b0dcfbde176950d9/test/unit/org/scratchOrgInfoGenerator.test.ts#L600-L601
(pass it an object with your `$schema` property and verify that it's removed.

Your code looks fine and I'm sure it works, but the test will preserve the behavior through future refactorings.

alanjaouen:
Thanks for feedback

I added a test for this, let me know if there is a better way to make this test

mshanemc:
Looks great, thanks for doing that.  

I reorganized the tests a bit and moved it to a branch on our repo so all the integration tests can run.
https://github.com/forcedotcom/sfdx-core/pull/1124

PR #1116, opened 2024-08-01
fix: wait to read until file is unlocked

mdonnalley:
### What does this PR do?

Wait to read config files until file is unlocked. This fixes an [issue](https://github.com/forcedotcom/cli/issues/2965) where parallel command executions caused files reads to return empty.

#### QA

In a linked plugin (with this branch of sfdx-core linked), create a new command that creates an `AuthInfo` and calls `.save()`

```typescript
import { SfCommand, Flags } from '@salesforce/sf-plugins-core';
import { Messages, AuthInfo } from '@salesforce/core';

Messages.importMessagesDirectoryFromMetaUrl(import.meta.url);
const messages = Messages.loadMessages('@salesforce/plugin-parallel-auth', 'hello.world');

export default class World extends SfCommand<void> {
  public static readonly summary = messages.getMessage('summary');
  public static readonly description = messages.getMessage('description');
  public static readonly examples = messages.getMessages('examples');

  public static readonly flags = {
    org: Flags.requiredOrg(),
  };

  public async run(): Promise<void> {
    const { flags } = await this.parse(World);

    const username = flags.org.getUsername();
    if (!username) {
      return;
    }

    this.log(`Updating auth info for ${username} (PID: ${process.ppid})`);
    const authInfo = await AuthInfo.create({ username });
    authInfo.update({ username });
    await authInfo.save();
  }
}
```

In dreamhouse-lwc, create a scratch org then run the command 10 times in parallel:

```
sf hello world --org my-org & \
sf hello world --org my-org & \
sf hello world --org my-org & \
sf hello world --org my-org & \
sf hello world --org my-org & \
sf hello world --org my-org & \
sf hello world --org my-org & \
sf hello world --org my-org & \
sf hello world --org my-org & \
sf hello world --org my-org & \
wait
```

**NOTE: this repros the issue ~50% of the time**

I would consider it successful if you don't see any JSON parsing issues in 10 consecutive attempts

@W-16322519@


mshanemc (approved):
approved.  feedback was done via PRs.

QA: wrote a NUT and had Mike re-check the changes using his original local replication.

PR #1136, opened 2024-09-11
fix: update error message for sfdx auth url

shetzel:
### What does this PR do?
The error message in sfdx-core needs to be updated so that the "force://" protocol language is clearer.

Also updated the `buildRefreshTokenConfig()` method to properly wrap and throw the original error.

@W-16726261@

### DIFF
--- a/messages/core.md
+++ b/messages/core.md
@@ -34,6 +34,10 @@ Errors encountered:
 
 Error authenticating with the refresh token due to: %s
 
+# invalidSfdxAuthUrlError
+
+Invalid SFDX auth URL. Must be in the format "force://<clientId>:<clientSecret>:<refreshToken>@<instanceUrl>". Note that the "instanceUrl" inside the SFDX auth URL doesn\'t include a protocol ("https://").
+
 # orgDataNotAvailableError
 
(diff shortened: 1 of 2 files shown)


messages/core.md
```diff
@@ -34,6 +34,10 @@ Errors encountered:
 
 Error authenticating with the refresh token due to: %s
 
+# invalidSfdxAuthUrlError
+
+Invalid SFDX auth URL. Must be in the format "force://<clientId>:<clientSecret>:<refreshToken>@<instanceUrl>". Note that the "instanceUrl" inside the SFDX auth URL doesn\'t include a protocol ("https://").
```
jshackell-sfdc, line 39:
```suggestion
Invalid SFDX authorization URL. Must be in the format "force://<clientId>:<clientSecret>:<refreshToken>@<instanceUrl>". Note that the "instanceUrl" inside the SFDX authorization URL doesn\'t include the protocol ("https://"). Run "org display --target-org" on an org to see an example of an SFDX authorization URL. 
```

jshackell-sfdc: Okay?  I updated the name of the URL to match what's in the "org login sfdx-url" help.  And I think it's okay not to put that whole stuff about commands always using HTTPS here.  One can assume at this point that they've run --help on "org login sfdx-url".

jshackell-sfdc (dismissed):
See my edit

WillieRuemmele (approved):

PR #1146, opened 2024-10-02
Mz/bundle test cover

mingxuanzhangsfdx:
### What does this PR do?
This PR adds several checks for bundling to prevent potential changes from breaking the process, including:
1. avoid dynamic imports.
i. report an error when detecting dynamic imports. —supported:dynamic-import=false and 
ii. set [log-override](https://esbuild.github.io/api/#log-override) as error for dynamic imports
2. pay attention to the usage of fs read from a local file - generally esbuild will not be able to bundle the local file referenced by fs directly, tsc can do so by specifying the artifacts in tsconfig.json
i. add a check in husky whenever there is a new commit to send warnings of every fs.read* usage in the codebase
3. if there are modules that are not bundle-able, and we are not aware of, the test workflow should fail.
4. the [relative path](https://github.com/forcedotcom/sfdx-core/blob/56e39c6da642fd66fed349bf85df056d2a006ad5/src/logger/logger.ts#L183) to get transformStream in pino-logger should not be changed.
i. check during husky commit hook
ii. report an error if the target reference does not exist in updateForBundling.js
iii. add explanation to where transformStream is referenced to raise developers’ caution
iv. during bundling, if const searchString = /\$\{process\.cwd\(\)\}\$\{require\("path"\)\.sep\}tmp-lib/g; is not detected in the bundled file, an error should be raised and let [Mingxuan Zhang](https://salesforce.quip.com/bXfAEAxCoUR) or others from IDEx team know to figure it out.

@W-16304459@


.husky/pre-commit
```diff
@@ -2,3 +2,4 @@
 . "$(dirname "$0")/_/husky.sh"
 
 yarn lint && yarn pretty-quick --staged
+node ./scripts/scanTs.js
```
mingxuanzhangsfdx, line 5: I added this extra step to scan all ts artifacts and list usages of fs.read*

peternhale (approved):

.github/workflows/perf.yml
```diff
@@ -19,6 +19,8 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
+        with:
```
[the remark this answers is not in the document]
peternhale: What is the minimum supported node version for sfdx-core? Is there a way to not change nodejs version?

peternhale (changes requested):

scripts/bundleWithCheck.js
```diff
@@ -0,0 +1,16 @@
+const { exec } = require('child_process');
+
+const commandToRun = 'yarn build && node ./scripts/build.js';
+
+// Run the command
+exec(commandToRun, (error, stdout, stderr) => {
+  // Combine stdout and stderr to check the entire output
+  const output = `${stdout}\n${stderr}`;
+  // Check if the output contains the error string of esbuild
+  if (output.includes('[require-resolve-not-external]')) {
```
WillieRuemmele, line 10: what is the error output like, would this statement be false if there were multiple errors in that array? you could just check for `require-resolve-not-external`?

scripts/bundlingUtils.js
```diff
@@ -0,0 +1,135 @@
+const fs = require('fs');
+const path = require('path');
+
+// Function to update package.json
```
WillieRuemmele, line 4: could this comment be updated to why we need to update the pjson?  - same for all similar jsdoc-style comments

scripts/bundlingUtils.js
```diff
@@ -0,0 +1,135 @@
...
+    if (err) {
+      console.error(`Error reading package.json: ${err}`);
+      return;
+    }
+
+    try {
+      const packageJson = JSON.parse(data);
+
+      // Update package name if necessary
+      if (packageJson.name && packageJson.name === '@salesforce/core') {
```
WillieRuemmele, line 18: optional accessor? `packageJson?.name===`?

scripts/bundlingUtils.js
```diff
@@ -0,0 +1,135 @@
...
+        `Error: The target string "${targetString}" was not found in logger.ts.\n Please make sure to bundle transformStream by referencing the new path or reach out to IDEx Foundations Team.`
+      );
+      return;
+    }
+  });
+}
+
+// Function to update logger.ts
+function updateLoggerTs() {
+  const loggerPath = './src/logger/logger.ts';
```
WillieRuemmele, line 70: can this and `checkTransformStreamPath` be combined? they're doing a lot of the same? at least share some of the variables

scripts/bundleWithCheck.js
mingxuanzhangsfdx: The output from bundling process is like this 
image
So each error with details will list in a row.

WillieRuemmele (approved):

peternhale (approved):

PR #1147, opened 2024-10-03
fix: improved scratch org create authentication

shetzel:
### What does this PR do?
1. The `signupTargetLoginUrl` override used in sfdx-project.json files was not used for scratchOrgResumes, causing auth code exchange issues during resume.  It is now used.
2. encrypts the client secret written to the scratch-create-cache.json file.
3. If the auth code exchange fails when using a loginUrl that differs from `ScratchOrgInfo.LoginUrl` another attempt is made using `ScratchOrgInfo.LoginUrl`.
4. Improves debug logging for the loginUrl used during the auth code exchange

@W-16825003@

To QA:
- use a devhub with another connected app that has a consumer secret and use it to create scratch orgs.  Ensure the clientSecret value written to the cache file is encrypted.
- resume a scratch org create where the `signupTargetLoginUrl` is necessary and ensure that login URL override is used for the auth code exchange.
- check that another attempt is made if the auth code exchange fails with the first loginUrl if that loginUrl doesn't match ScratchOrgInfo.LoginUrl.

### DIFF
--- a/test/unit/org/scratchOrgInfoApi.test.ts
+++ b/test/unit/org/scratchOrgInfoApi.test.ts
@@ -35,14 +35,14 @@ const errorCodesMessages = Messages.loadMessages('@salesforce/core', 'scratchOrg
 
 const scratchOrgInfoId = '2SRK0000001QZxF';
 const TEMPLATE_SCRATCH_ORG_INFO: ScratchOrgInfo = {
-  LoginUrl: 'https://login.salesforce.com',
+  LoginUrl: 'https://scratch-org-info-login-url.salesforce.com',
   Snapshot: '1234',
   AuthCode: '1234',
   Status: 'New',
   SignupEmail: 'sfdx-cli@salesforce.com',
   SignupUsername: 'sfdx-cli',
   Username: 'sfdx-cli',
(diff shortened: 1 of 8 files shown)


test/unit/org/scratchOrgInfoApi.test.ts
```diff
@@ -466,29 +466,48 @@ describe('authorizeScratchOrg', () => {
...
 
     expect(result).to.be.equal(authInfo);
+    expect(authInfoStub.firstCall.args[0]).to.deep.equal({
+      username: TEMPLATE_SCRATCH_ORG_INFO.SignupUsername,
+      parentUsername: username,
+      oauth2Options: {
+        loginUrl: `https://${TEMPLATE_SCRATCH_ORG_INFO.SignupInstance}.salesforce.com`,
+        privateKeyFile: privateKey,
+      },
+    });
```
shetzel, line 484: None of the existing tests could possibly fail since `result` is a stubbed return setup by the tests.  I added expectations that should have been there from the start, as well as a few additional checks for the new code I added.

mdonnalley (approved):

PR #1166, opened 2025-02-14
fix: query the right `SandboxProcess` on refresh

cristiand391:
### What does this PR do?
draft

@W-17246841@

### DIFF
--- a/src/org/org.ts
+++ b/src/org/org.ts
@@ -346,17 +346,26 @@ export class Org extends AsyncOptionalCreatable<Org.Options> {
    */
   public async resumeSandbox(
     resumeSandboxRequest: ResumeSandboxRequest,
-    options: { wait?: Duration; interval?: Duration; async?: boolean } = {
+    options: { wait?: Duration; interval?: Duration; async?: boolean; refresh?: boolean } = {
       wait: Duration.minutes(0),
       async: false,
       interval: Duration.seconds(30),
+      refresh: false,
     }
   ): Promise<SandboxProcessObject> {
     this.logger.debug(resumeSandboxRequest, 'ResumeSandbox called with ResumeSandboxRequest');
     let sandboxCreationProgress: SandboxProcessObject;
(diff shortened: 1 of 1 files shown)


src/org/org.ts
```diff
@@ -1506,13 +1516,17 @@ export class Org extends AsyncOptionalCreatable<Org.Options> {
...
+  private async querySandboxProcess(where: string, refresh = false): Promise<SandboxProcessObject> {
     const soql = `SELECT ${sandboxProcessFields.join(
       ','
     )} FROM SandboxProcess WHERE ${where} ORDER BY CreatedDate DESC`;
-    const result = (await this.connection.tooling.query<SandboxProcessObject>(soql)).records.filter(
-      (item) => !item.Status.startsWith('Del')
-    );
+    const result = (await this.connection.tooling.query<SandboxProcessObject>(soql)).records.filter((item) => {
+      if (refresh) {
+        return !item.Status.startsWith('Completed');
```
cristiand391, line 1525: this is probably bad (resuming a finished refresh process would return no records, fix later!

src/org/org.ts
```diff
@@ -346,17 +346,26 @@ export class Org extends AsyncOptionalCreatable<Org.Options> {
...
+      refresh: false,
     }
   ): Promise<SandboxProcessObject> {
     this.logger.debug(resumeSandboxRequest, 'ResumeSandbox called with ResumeSandboxRequest');
     let sandboxCreationProgress: SandboxProcessObject;
     // seed the sandboxCreationProgress via the resumeSandboxRequest options
     if (resumeSandboxRequest.SandboxProcessObjId) {
-      sandboxCreationProgress = await this.querySandboxProcessById(resumeSandboxRequest.SandboxProcessObjId);
+      if (options.refresh) {
+        const sbxProcess = await this.querySandboxProcessById(resumeSandboxRequest.SandboxProcessObjId);
```
cristiand391, line 361: do we need this request? why isn't `SandboxInfoId` already in the cache?

PR #1177, opened 2025-03-31
fix: W-17889061 - update resume scratch 

soridalac:
### What does this PR do?
update the library when scratch still in New or Creating process
@W-17889061@

### DIFF
--- a/src/org/scratchOrgCreate.ts
+++ b/src/org/scratchOrgCreate.ts
@@ -104,7 +104,10 @@ const validateRetry = (retry: number): void => {
   }
 };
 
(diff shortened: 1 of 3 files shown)


cristiand391 (approved):

cristiand391:
**QA notes**:
this feature was requested by Vivek when scratch org creation was slow in prod, it's tricky to repro without a messy env so for QA I manually set the status field to make the code go into the polling for a few times then go back to the "finished" state to emulate a server-side scenario where creating a scratch took a few minutes.

✅ create scratch org async, then `org resume scratch --wait` polls for status and resumes auth/settings deploy after it's done
✅ times out if status is still new/error after `--wait` time passes.

PR #1181, opened 2025-04-11
fix: add retry limit option to polling client W-18203875

WillieRuemmele:
### What does this PR do?

adds `retryLimit` option to `PollingClient.options` to limit the amount of polling calls 

⚠️ : if set below what frequency * timeout = call count, will stop polling at the passed `retryLimit`

@W-18203875@


cristiand391 (approved):

cristiand391:
**QA notes**:

✅ long timeout with `retryLimit: 20`, exits after reaching retry limit
✅ short timeout with big retry limit, times out before reaching retry limit

PR #1188, opened 2025-05-14
feat(auth): allow to link connected apps to a user W-18394868

cristiand391:
### What does this PR do?
ref: 
* https://github.com/salesforcecli/plugin-auth/pull/1306
* https://github.com/salesforcecli/plugin-agent/pull/157

Doc: https://salesforce.quip.com/sUaFAknyDkue

This PR adds support for linking connected apps/ECAs to an authenticated user, allowing consumers to switch to a specific app for oauth-specific requirements like JWT-based access tokens, specific API scopes, etc.

#### AuthInfo
Adds a new `apps` property that contains an object of CA/ECAs with oauth info needed for requests.
For the scope of this PR, the only supported way to add new apps is via the web server flow (see linked plugin-auth PR ⬆️ )

#### Connection:
Adds an `app` property to support picking up a linked app in the provided auth info:
```ts
  const authInfo = await AuthInfo.create({
    username: 'johndoe@salesforce.com',
  });

  // get oauth creds from the main connected app in auth (opaque tokens)
  const conn = await Connection.create({
    authInfo,
  });

  // get oauth creds from the linked `agent-jwt-app`
  const jwtConn = await Connection.create({
    authInfo,
    app: 'agent-jwt-app',
  });
```

#### WebOAuthServer
Add support for linking apps to an authenticated users via the web server flow.
Before:
Web server only handled 1st time auth, always creating a new auth file
After
If you pass `app` and `username` params, after a successful auth in the browser it will save the oauth info in `authinfo.apps[app]`.

@W-18394868@


src/org/authInfo.ts
```diff
@@ -1277,6 +1338,14 @@ export namespace AuthInfo {
      * OAuth options.
      */
     oauth2Options?: JwtOAuth2Config;
+    apps?: Array<{
+      name: string;
+      accessToken: string;
+      refreshToken: string;
+      clientId: string;
+      clientSecret?: string;
+      // privateKey: string;
```
WillieRuemmele, line 1347: still needed?

WillieRuemmele (dismissed):

PR #1204, opened 2025-06-11
feat(W-18724886): add env vars for debug logger

mdonnalley:
### What does this PR do?

- Adds `SF_LOG_COLORIZE` to allow users to disable colored logs, defaults to true
- Adds `SF_LOG_STDERR` to allow users to send logs to stderr, defaults to false (meaning logs will go to stdout by default)

Having these two env vars allows us to successfully emit logs from the MCP server. Cursor and vscode seem to expect stringified JSON-RPC to be sent to stdout but not for stderr. They also don't render the ansi sequences into colors. So for simplicity's sake we're going to enable these two env vars whenever users enable logging on the MCP server

@W-18724886@

### DIFF
--- a/src/logger/logger.ts
+++ b/src/logger/logger.ts
@@ -452,11 +452,15 @@ export class Logger {
 
 /** return various streams that the logger could send data to, depending on the options and env  */
 const getWriteStream = (level = 'warn'): pino.TransportSingleOptions => {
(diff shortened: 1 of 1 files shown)


shetzel (approved):
nit: `env.getString()` and `env.getBoolean()` support passing a default.

PR #1205, opened 2025-06-17
feat: add features query to core -W-18539941

soridalac:
### What does this PR do?
Add Features query for SandboxInfo and SandboxProcess to the org library.
@W-18539941@

### DIFF
--- a/src/org/org.ts
+++ b/src/org/org.ts
@@ -110,6 +110,7 @@ export type SandboxProcessObject = {
   Description?: string;
   ApexClassId?: string;
   EndDate?: string;
+  Features?: string;
 };
 const sandboxProcessFields = [
   'Id',
@@ -123,6 +124,7 @@ const sandboxProcessFields = [
   'SourceId',
   'Description',
   'EndDate',
+  'Features',
 ];
 
 export type SandboxRequest = {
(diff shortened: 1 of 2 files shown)


src/org/org.ts
```diff
@@ -755,6 +764,90 @@ export class Org extends AsyncOptionalCreatable<Org.Options> {
...
+      )
+    ).Features;
+  }
+
+  /**
+   * Query SandboxInfo Features by ID
+   *
+   * @param id The SandboxInfo ID to query for
+   */
+  public async querySandboxInfoFeaturesById(id: string): Promise<string | undefined> {
```
shetzel, line 786: This method is not needed and should be deleted.  `querySandboxInfo()` should be used instead.

src/org/org.ts
```diff
@@ -755,6 +764,90 @@ export class Org extends AsyncOptionalCreatable<Org.Options> {
...
+    }
+    return result[0];
+  }
+
+  /**
+   * Get Features for a sandbox by name or ID, checking both SandboxProcess and SandboxInfo
+   *
+   * @param sandboxName The sandbox name or ID to query for
+   */
+  public async getSandboxFeatures(sandboxName: string): Promise<string | undefined> {
```
shetzel, line 820: This is too specific to have on the Org class and should be removed.  The only function to add should be `querySandboxInfo`

shetzel (changes requested):

shetzel (approved):

PR #1208, opened 2025-06-26
feat(W-18653307): use mutex to make ConfigAgg and StateAgg thread safe

mdonnalley:
### What does this PR do?

Use a mutex to ensure thread-safety when getting or delete instances of ConfigAggregator or StateAggregator

@W-18653307@


WillieRuemmele (approved):
only wondering if this is a breaking change, but if this breaks someone they're probably using these classes in an unsupported/unintended way anyways

iowillhoit (approved):

PR #1211, opened 2025-07-09
W-18988586 feat: support paths for ConfigAggregator

mshanemc:
### What does this PR do?
ConfigAggregator is a singleton
instead of caching 1 ConfigAggregator instance, it has one per sfdx project
to help with multiple project and VSCode when the user changes projects and the singleton has to be destroyed and recreated.

@W-18988586@

### DIFF
--- a/src/config/configAggregator.ts
+++ b/src/config/configAggregator.ts
@@ -77,7 +77,7 @@ export type ConfigInfo = {
  */
 export class ConfigAggregator extends AsyncOptionalCreatable<ConfigAggregator.Options> {
   protected static encrypted = true;
-  protected static instance: AsyncOptionalCreatable | undefined;
+  protected static instances = new Map<string, ConfigAggregator>();
 
   private static readonly mutex = new Mutex();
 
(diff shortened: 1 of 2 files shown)


src/config/configAggregator.ts
```diff
@@ -116,16 +116,27 @@ export class ConfigAggregator extends AsyncOptionalCreatable<ConfigAggregator.Op
 
   // Use typing from AsyncOptionalCreatable to support extending ConfigAggregator.
   // We really don't want ConfigAggregator extended but typescript doesn't support a final.
+  public static async create<T extends ConfigAggregator>(
```
mshanemc, line 119: this makes the types cleaner inside the method, but is kinda ugly
explanation: because this could be subclassed, the original let anything that extends AsyncOptionalCreatable use it.

we can have a tighter condition , where if we know the thing is really a ConfigAggregator (or its subclasses) the types can be more specific.

I'd also be happy to let `create` only work with ConfigAggregator but I don't understand heavy-OOP to know the implications.  There are a few `extends ConfigAggregator` I saw in the wild on github so I wanted to be careful with the public API.

src/config/configAggregator.ts
```diff
@@ -158,23 +174,22 @@ export class ConfigAggregator extends AsyncOptionalCreatable<ConfigAggregator.Op
...
    * asynchronously by calling {@link ConfigAggregator.reload}
    */
-  // Use typing from AsyncOptionalCreatable to support extending ConfigAggregator.
-  // We really don't want ConfigAggregator extended but typescript doesn't support a final.
-  private static getInstance<P, T extends AsyncOptionalCreatable<P>>(this: new () => T): T {
-    if (!ConfigAggregator.instance) {
-      ConfigAggregator.instance = new this();
-      (ConfigAggregator.instance as ConfigAggregator).loadPropertiesSync();
+  private static getInstance(projectPath = process.cwd()): ConfigAggregator {
+    if (!ConfigAggregator.instances.has(projectPath)) {
```
mshanemc, line 187: I think this has the potential to be sketchy, even before this PR.

because getValue is sync, we can't do the real `create` here.  So I think it's going to have a ConfigAggregator missing the options.configMeta stuff?

mdonnalley: Agreed - but I think that's why the jsdoc points out that you should call `reload` after

PR #1221, opened 2025-08-01
W-19183667 feat: web bundling, virtual fs

mshanemc:
### What does this PR do?
have a virtualFS that could be 
- node => regular node fs
- browser => memfs

use that fs everywhere in the library

fix global.isWeb to account for webWorkers (no `window`)
pino thread logger bypass based in isWeb

see comments for other misc changes
buncha fallback for undefineds in template literals

fileLocking can have fs injected into it, so we do
file locks become noop on the web

remove the deprecatd getCnames and its test

@W-19183667@

> 📓  there are no changes to the bundling code in this PR; all the polyfills and bundling steps for web happen in the vscode repo.  These changes are just to make the library bundling-capable by vscode ext.


src/fs/fs.ts
```diff
@@ -0,0 +1,68 @@
...
+
+import * as nodeFs from 'node:fs';
+// yes, we're going to import it even though it might not be used.
+// the alternatives were all worse without top-level await (iife, runtime errors from something trying to use it before it's initialized)
+import * as memfs from 'memfs';
+import type { VirtualFs } from './types';
+
+export let fs: VirtualFs;
+
+const isWeb = (): boolean => process.env.FORCE_MEMFS === 'true' || 'window' in globalThis || 'self' in globalThis;
```
shetzel, line 16: Why not use `Global.isWeb`?  E.g.,

`const isWeb = (): boolean => process.env.FORCE_MEMFS === 'true' || Global.isWeb`

shetzel (approved):

src/fs/fs.ts
shetzel: Ahh, because circular ref.  Got it.

shetzel (approved):

PR #1228, opened 2025-08-19
feat: support code builder web auth - W-16646649

shetzel:
### What does this PR do?
Supports Code Builder web auth.

@W-16646649@
@W-16646367@
@W-16645422@


src/webOAuthServer.ts
```diff
@@ -28,6 +28,8 @@ const messages = Messages.loadMessages('@salesforce/core', 'auth');
 // Server ignores requests for site icons
 const iconPaths = ['/favicon.ico', '/apple-touch-icon-precomposed.png'];
 
+const CODE_BUILDER_REDIRECT_URI = 'api/oauth2/salesforce/callback';
+
```
jonnyhork, line 32: @shetzel Sorry the callback URL changed a little to match the same format as the rest of our api.

in the connected app they are

```
https://codebuilder-platform-api.herokuapp.com/api/auth/salesforce/callback
https://codebuilder-platform-api-stage-c6fd5fccd5aa.herokuapp.com/api/auth/salesforce/callback
```

src/webOAuthServer.ts
```diff
@@ -193,10 +197,34 @@ export class WebOAuthServer extends AsyncCreatable<WebOAuthServer.Options> {
...
+        throw messages.createError('error.missingCodeBuilderUri');
+      }
+      const cbStateSha = env.getString('CODE_BUILDER_STATE_SHA');
+      if (!cbStateSha) {
+        throw messages.createError('error.missingCodeBuilderStateSha');
+      }
+      this.oauthConfig.state = JSON.stringify({
+        PORT: port,
+        CODE_BUILDER_URI: cbUri,
+        CODE_BUILDER_STATE_SHA: cbStateSha,
```
jonnyhork, line 219: I think we can slim this down a bit.

We can likely just look for and pass the encrypted state value and not worry about the `CODE_BUILDER_URI`

The state will contain either the CB URL or the instance id and we can redirect to the right place with either. Since its encrypted/decrypted on the server that should be enough to determine if the value is trust worthy.

messages/auth.md
```diff
@@ -24,6 +24,14 @@ Invalid request method: %s
 
 Invalid request uri: %s
 
+# warn.invalidClientId
+
+Code Builder web authentication is only supported with the Code Builder connected app. Ignoring client ID: %s
```
shetzel, line 29: note: this is not an error message seen by customers.  It's a warning message sent to a debug log.

cristiand391 (approved):

PR #1230, opened 2025-08-21
feat: add global to top-level exports W-19376651

mshanemc:
### What does this PR do?
add global to top-level exports

@W-19376651@


cristiand391 (approved):
fwiw, we are already exporting this in our index barrel file (so, not exposing some previously hidden class).

PR #1231, opened 2025-08-21
ci: no bundling W-19376651

mshanemc:
### What does this PR do?
don't do any bundle publishing
test bundleability as part of `test`

[@W-19376651@](https://gus.lightning.force.com/a07EE00002KOBrAYAX)


scripts/build.mjs
```diff
@@ -0,0 +1,26 @@
...
+  format: 'cjs',
+  supported: {
+    'dynamic-import': false,
+  },
+  logOverride: {
+    'unsupported-dynamic-import': 'error',
+  },
+  define: {
+    // this prevents the logger from writing to any files, obviating the need for pino-bundling stuff
+    'process.env.SF_DISABLE_LOG_FILE': "'true'",
```
WillieRuemmele, line 23: and this double quote `"'true'"`? is correct?

scripts/build.mjs
```diff
@@ -0,0 +1,26 @@
...
+  },
+  outdir: 'dist',
```
WillieRuemmele, line 25: why not `lib`?

scripts/build.mjs
mshanemc: yeah, steve asked the same thing on the other one.  it needs `'true'` the string instead of `true` the boolean.

Spent about 30 min on that a few weeks back before the AI told me that it needed this.  And it actually works!

scripts/build.mjs
mshanemc: and we want that 2-step process so that you can compile without bundling, and so that libraries can do that inlining of `messages` trick, which was one of the things that bundling didn't handle well before.

PR #1279, opened 2026-04-24
fix: incorrect enum value, swapped log rotation descriptions, wrong JSDoc, and stray artifacts in error messages

ZLeventer:
## Summary

Five bugs found across `org.ts`, `envVars.ts`, `envVars.md`, `configAggregator.ts`, and `scratchOrgInfoGenerator.md`:

### `src/org/org.ts:1868` — `TRIAL_EXPIRATION_DATE` enum has wrong string value

```ts
// before
TRIAL_EXPIRATION_DATE = 'trailExpirationDate',

// after
TRIAL_EXPIRATION_DATE = 'trialExpirationDate',
```

This enum value is used as a property key when reading auth fields from the org. The Salesforce API returns `trialExpirationDate`; the typo (`trail`) would cause trial expiration date lookups to silently return `undefined`.

### `src/config/envVars.ts:263` — `SFDX_LAZY_LOAD_MODULES` shows wrong description

```ts
// before
[EnvironmentVariable.SFDX_LAZY_LOAD_MODULES]: {
  description: getMessage(EnvironmentVariable.SFDX_USE_PROGRESS_BAR),  // copy-paste bug

// after
[EnvironmentVariable.SFDX_LAZY_LOAD_MODULES]: {
  description: getMessage(EnvironmentVariable.SFDX_LAZY_LOAD_MODULES),
```

`SFDX_LAZY_LOAD_MODULES` displayed the help text for `SFDX_USE_PROGRESS_BAR` instead of its own description. Every other entry in `SUPPORTED_ENV_VARS` correctly references its own key.

### `messages/envVars.md` — `sfdxLogRotationCount` / `sfdxLogRotationPeriod` descriptions are swapped (both SFDX_ and SF_ variants)

`sfdxLogRotationCount` had the rotation-period description ("Example '1d' will rotate logs daily"), and `sfdxLogRotationPeriod` had the count description ("3 backup files to keep"). Same swap in the `sfLogRotationCount` / `sfLogRotationPeriod` entries. Fixed by swapping both pairs back to their correct keys.

### `src/config/configAggregator.ts:291-293` — `getLocation()` JSDoc lists `GLOBAL` and `ENVIRONMENT` backwards

The JSDoc said `GLOBAL` for env vars and `ENVIRONMENT` for global config, which is the opposite of what the method actually returns. Corrected to match the implementation.

### `messages/scratchOrgInfoGenerator.md` — 5 error messages have trailing `",` artifacts

Remnants of a JSON-to-markdown migration. Each affected message would appear with a literal trailing `",` in user-facing error output (e.g., `"...scratch org definition.","`). Removed the artifacts from all 5 messages.

## Test plan
- [ ] `yarn build` passes
- [ ] Existing tests pass
- [ ] Manual: `sf env var list` should show correct descriptions for `SFDX_LAZY_LOAD_MODULES` and log rotation vars

### DIFF
--- a/messages/envVars.md
+++ b/messages/envVars.md
@@ -104,11 +104,11 @@ Level of messages that the CLI writes to the log file. Valid values are trace, d
 
 # sfdxLogRotationCount
 
-The default rotation period for logs. Example '1d' will rotate logs daily (at midnight).
+The number of backup rotated log files to keep. Example: '3' will have the base sf.log file, and the past 3 (period) log files.
 
 # sfdxLogRotationPeriod
 
-The number of backup rotated log files to keep. Example: '3' will have the base sf.log file, and the past 3 (period) log files.
+The default rotation period for logs. Example '1d' will rotate logs daily (at midnight).
 
 # sfdxMaxQueryLimit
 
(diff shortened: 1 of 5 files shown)


jshackell-sfdc (approved):
approve markdown updates only

PR #1290, opened 2026-05-20
feat: persist orgEdition during Org.create for all org types - W-22583951

peternhale:
## Summary

- Add `AuthInfo.determineOrg()` — a single shared method that queries the Organization table once and persists all org metadata (`orgEdition`, `isScratch`, `isSandbox`, `name`, `instanceName`, `namespacePrefix`, `trialExpirationDate`) to the auth JSON file
- All login commands (web, jwt, sfdx-url, access-token) now capture org metadata at login time via `initAuthOptions` → `determineOrg`
- `Org.create()` also triggers `determineOrg` in `Org.init()`, ensuring metadata is populated for any code path that instantiates an Org
- Remove `getNamespacePrefix` — a raw jsforce Transport call to the Organization table that is now redundant. The consolidated query uses the core `Connection.singleRecordQuery` abstraction instead
- Uses `orgEdition` as a sentinel: if already set in the auth file, the Organization query is skipped entirely

@W-22583951@

### Notes

**Problem:** The previous PR (`feat: add org edition (OrganizationType) to auth info fields`) added `orgEdition` to `AuthFields` and `updateLocalInformation()`, but `updateLocalInformation()` was never triggered during the login flow. Login commands work through `AuthInfo` and `WebOAuthServer` — they never instantiate an `Org`. Additionally, the existing code had multiple redundant Organization queries:

1. `getNamespacePrefix` — raw Transport call to `SELECT NamespacePrefix FROM Organization` during `initAuthOptions`
2. `Org.retrieveOrganizationInformation()` — full Organization query via `updateLocalInformation()` in `Org.init()`
3. `determineIfScratch()` / `determineIfSandbox()` — could each trigger `updateLocalInformation()` on cache miss

**Solution:** Consolidate into `AuthInfo.determineOrg()` which:
- Queries `SELECT Name, InstanceName, IsSandbox, TrialExpirationDate, NamespacePrefix, OrganizationType FROM Organization` — one round trip
- Persists all derived fields to the auth file
- Is called from three convergence points:
  - `initAuthOptions` (after `this.update(authConfig)`) — covers all login commands
  - `identifyPossibleScratchOrgs` — covers the login enrichment path
  - `Org.init()` — covers `Org.create()` consumers (plugin-org, plugin-deploy-retrieve, etc.)
- Subsequent calls are no-ops via the `orgEdition` sentinel check
- Since `isScratch`/`isSandbox` are now populated by `determineOrg`, calls to `determineIfScratch()`/`determineIfSandbox()` (e.g. from `tracksSource()`) will hit cache and avoid additional Organization queries

**Breaking change:** `getNamespacePrefix` was a private method, so no public API change. The `namespacePrefix` field is still populated — now via the consolidated Organization query rather than a separate raw Transport call.

### Automated tests
- [x] `determineOrg` queries Organization when `orgEdition` is missing from auth file
- [x] `determineOrg` skips query when `orgEdition` is already present (sentinel check)
- [x] `Org.create` succeeds even if `determineOrg` throws (best-effort)
- [x] `identifyPossibleScratchOrgs` updates auth file as a scratch org
- [x] `identifyPossibleScratchOrgs` updates auth file as a sandbox from possible prod orgs
- [x] `identifyPossibleScratchOrgs` does not update auth file when no dev hubs exist
- [x] `identifyPossibleScratchOrgs` does not update auth file when state already known
- [x] `identifyPossibleScratchOrgs` does not update auth file when no orgId present

### Manual tests
- [x] `sf org login web` — after completion, verified the orgEdition in the auth json was set to org edition of the target org
- [x] `sf org create scratch` — config requested Developer Edition. After completion, verified the orgEdition in the auth json was set to "Developer Edition"
- [x] `sf org create scratch` — config requested Enterprise Edition. After completion, verified the orgEdition in the auth json was set to "Enterprise Edition"
- [x] `sf org list` — this command verifies connectivity to each org and as a by-product of checking each, each org's existing json file was updated to include the new org edition

### DIFF
--- a/src/org/authInfo.ts
+++ b/src/org/authInfo.ts
@@ -9,7 +9,6 @@
 import { randomBytes } from 'node:crypto';
 import { resolve as pathResolve } from 'node:path';
 import * as os from 'node:os';
-import { Record as RecordType } from '@jsforce/jsforce-node';
 import { AsyncOptionalCreatable, env, isEmpty, parseJson, parseJsonMap } from '@salesforce/kit';
 import {
   AnyJson,
@@ -40,7 +39,7 @@ import { Messages } from '../messages';
 import { getLoginAudienceCombos, SfdcUrl } from '../util/sfdcUrl';
 import { findSuggestion } from '../util/findSuggestion';
 import { Connection, SFDX_HTTP_HEADERS } from './connection';
-import { Org, SandboxFields } from './org';
+import { Org, OrganizationInformation, SandboxFields } from './org';
 import { OrgConfigProperties } from './orgConfigProperties';
 
 Messages.importMessagesDirectory(__dirname);
@@ -406,6 +405,8 @@ export class AuthInfo extends AsyncOptionalCreatable<AuthInfo.Options> {
     // authInfo before it is necessary.
     const logger = await Logger.child('Common', { tag: 'identifyPossibleScratchOrgs' });
 
+    await AuthInfo.determineOrg(orgAuthInfo);
+
     // return if we already know the hub org, we know it is a devhub or prod-like, or no orgId present
     if (Boolean(fields.isDevHub) || Boolean(fields.devHubUsername) || !fields.orgId) return;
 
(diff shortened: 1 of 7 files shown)


soridalac (dismissed):

src/org/authInfo.ts
```diff
@@ -462,6 +463,33 @@ export class AuthInfo extends AsyncOptionalCreatable<AuthInfo.Options> {
...
+
+    try {
+      const conn = await Connection.create({ authInfo: orgAuthInfo });
+      const result = await conn.singleRecordQuery<OrganizationInformation>(
+        'SELECT Name, InstanceName, IsSandbox, TrialExpirationDate, NamespacePrefix, OrganizationType FROM Organization'
+      );
+      await orgAuthInfo.save({
+        [Org.Fields.NAME]: result.Name,
+        [Org.Fields.INSTANCE_NAME]: result.InstanceName,
+        [Org.Fields.NAMESPACE_PREFIX]: result.NamespacePrefix,
```
mshanemc: NS prefix can be added after the org is created.  The problem with the `orgEdition` sentinel would occur with a path like

1. auth an org (edition is set, NS is null)
2. link org to a DE with a NS, or create a NS on a DE
3. namespace is wrong and stays wrong without doing something drastic (ex: delete org and reauth, manually edit, etc).

mshanemc: everything else seems ok to do this was because they're immutable

mshanemc: I **think** once it has a NS that's immutable.

mshanemc (changes requested):

src/org/authInfo.ts
```diff
@@ -462,6 +463,33 @@ export class AuthInfo extends AsyncOptionalCreatable<AuthInfo.Options> {
     ]);
   }
 
+  /**
+   * Query the Organization object and persist org metadata fields if not already populated.
+   * Best-effort: silently returns if the org is unreachable or the query fails.
+   */
+  public static async determineOrg(orgAuthInfo: AuthInfo): Promise<void> {
```
mshanemc: do you have a use case in mind for needing to expose a new public static method on AuthInfo?  Who's going to call it and why?

As an alternative, you could export this as function and have Org call that when needed so that it's internal to sfdx-core so that we don't widen the API

peternhale: I'll refactor to an internal function

soridalac (dismissed):

mshanemc (approved):
approved code but didn't re-test.

NUT errors look auth related and probably worth looking into for this PR

PR #1295, opened 2026-06-10
@W-22620566: Change Lead business process picklist value to match Developer edition templates

smotwani11:
@W-22620566@
Summary

  - Change hardcoded Lead business process picklist value from "New - Not Contacted" to "Open - Not Contacted" to match Developer edition scratch org templates
  - Fixes settings deploy failures when objectSettings.lead.defaultRecordType is specified in definition files
  - Affects scratch org creation, package version create, and package convert

  Test plan

  - [x] Unit test added asserting Lead picklist value is "Open - Not Contacted"
  - [x] All 27 existing tests pass
  - [x] Manual E2E: sf org create scratch with Developer edition + Lead record type → succeeds
  - [x] Manual E2E: sf package convert with Developer edition definition file → generates correct Lead.object

### DIFF
--- a/src/org/scratchOrgSettingsGenerator.ts
+++ b/src/org/scratchOrgSettingsGenerator.ts
@@ -104,7 +104,7 @@ const calculateBusinessProcess = (
         businessProcessPicklistVal = 'New';
         break;
       case 'Lead':
-        businessProcessPicklistVal = 'New - Not Contacted';
+        businessProcessPicklistVal = 'Open - Not Contacted';
         break;
       case 'Opportunity':
         businessProcessPicklistVal = 'Prospecting';
--- a/test/unit/org/scratchOrgSettingsGenerator.test.ts
+++ b/test/unit/org/scratchOrgSettingsGenerator.test.ts
@@ -823,6 +823,37 @@ describe('scratchOrgSettingsGenerator', () => {
       expect(allRecordTypes).to.deep.equal(['Case.default']);
       expect(allbusinessProcesses).to.deep.equal(['Case.defaultProcess']);
     });
+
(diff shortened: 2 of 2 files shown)


smotwani11:
Overview
  
  What we changed

  One line in @salesforce/core → src/org/scratchOrgSettingsGenerator.ts:

  // Before
  businessProcessPicklistVal = 'New - Not Contacted';

  // After
  businessProcessPicklistVal = 'Open - Not Contacted';

  **Why we changed it**

The scratchOrgSettingsGenerator hardcodes picklist values for objects that require a business process when a record type is specified. For Lead, it used "New - Not Contacted" — a value that doesn't exist in any production scratch org template. The Developer edition template uses "Open - Not Contacted" instead.

When a user specifies objectSettings.lead.defaultRecordType in their definition file, the CLI generates a Lead.object XML containing a BusinessProcess that references this picklist value. The value gets deployed to the scratch/build org after creation. If the template doesn't have it → deploy fails:  

  [Error] Lead.DefaultProcess : Picklist value: New - Not Contacted not found

 This has been broken since the code was written (2019) and was never properly fixed (W-8967803, W-8937157 — both closed without resolution).

 **What this fixes**

  - sf org create scratch with Developer edition + Lead record type → was broken, now works
  - sf package version create with Lead record type in definition file → was broken, now works                                     
  - sf package convert with Lead record type in definition file → was broken, now works                                                             

 **What this doesn't fix**

Enterprise and Professional editions have completely different LeadStatus values (Contacted, Open, Qualified, Unqualified) — neither the old nor new value exists in  those templates. Those remain broken. The LeadsAndNotes team has a work item in their backlog to fix this issue permanently by addressing the root cause (hardcoded values that vary by edition template).

iowillhoit (approved):

PR #1301, opened 2026-06-18
fix: isWeb() no longer misdetects Bun as a browser environment @W-21890292@

### LINKED ISSUE #3535: isWeb() in fs.js incorrectly detects Bun as web environment, breaking keychain access
## Summary

When running `@salesforce/core` under the Bun runtime, the `isWeb()` check in `lib/fs/fs.js` returns `true`, causing the library to use `memfs` (in-memory filesystem) instead of `node:fs`. This breaks all filesystem-dependent functionality, including keychain access for org authentication.

## Root Cause

In `lib/fs/fs.js`, line 37:

```js
const isWeb = () => process.env.FORCE_MEMFS === 'true' || 'window' in globalThis || 'self' in globalThis;
```

Bun exposes `self` in `globalThis` as an alias for `globalThis` itself (for Web API compatibility), similar to how browsers expose `self`. This causes `'self' in globalThis` to return `true`, making `isWeb()` incorrectly identify Bun as a web/browser environment.

## Impact

All `fs` operations go through `memfs` instead of the real filesystem. The first visible symptom is typically:

```
MissingCredentialProgramError: Unable to find required security software: /usr/bin/security
```

This happens because `memfs.statSync('/usr/bin/security')` throws (the file doesn't exist in the in-memory filesystem), even though `/usr/bin/security` exists on disk.

## Reproduction

1. Install a project that depends on `@salesforce/core@^8.28.0`
2. Run it with Bun (e.g., `bun run my-cli-command`)
3. Any operation that triggers keychain/crypto access will fail with `MissingCredentialProgramError`

Verification:

```js
// Bun exposes self
bun -e "console.log('self' in globalThis)" // true

// Node.js does not
node -e "console.log('self' in globalThis)" // false
```

## Suggested Fix

The `isWeb()` check should use a more reliable detection method. Options:

1. Check for `window` (DOM) specifically, since `self` exists in both Web Workers and Bun:
   ```js
   const isWeb = () => process.env.FORCE_MEMFS === 'true' || (typeof window !== 'undefined' && typeof document !== 'undefined');
   ```

2. Exclude known server runtimes:
   ```js
   const isWeb = () => process.env.FORCE_MEMFS === 'true' || (('window' in globalThis || 'self' in globalThis) && !('process' in globalThis && process.versions?.node || process.versions?.bun));
   ```

3. Check for the absence of `node:fs` availability instead of checking globals.

## Environment

- `@salesforce/core`: 8.28.0
- Runtime: Bun 1.3.9
- OS: macOS (Darwin arm64)

## Related Repository

The affected code lives in [`forcedotcom/sfdx-core`](https://github.com/forcedotcom/sfdx-core) (`src/fs/fs.ts`).

WillieRuemmele:
## Summary
- `isWeb()` in `src/fs/fs.ts` and `Global.isWeb` in `src/global.ts` used `'self' in globalThis` which returns `true` in Bun (since Bun exposes `self` for Web API compatibility), incorrectly routing all fs operations through memfs
- Replaced with `'document' in globalThis && 'window' in globalThis` — both must be present, which only happens in actual browsers
- Added unit tests covering Node.js, Bun-like (`self` only), browser-like (`window` + `document`), and partial (`window` only) scenarios

Fixes https://github.com/forcedotcom/cli/issues/3535

## Work Item
@W-21890292@: isWeb() in fs.js incorrectly detects Bun as web environment, breaking keychain access

## Proof of Work
- Unit tests: 13 passing (4 new isWeb tests + 9 existing)
- NUT tests: 57 passing
- Lint: clean (0 errors, 7 pre-existing warnings)
- Pre-push hooks: all passed

## Validation
- Per-task review: 0 blockers, 2 warnings (Web Workers false negative — irrelevant for CLI library; FORCE_MEMFS inconsistency — pre-existing), 3 nits
- Integration: all checks pass

## Test plan
- [ ] Run under Bun: `bun -e "console.log('self' in globalThis)"` → true, but `isWeb()` → false
- [ ] Run under Node.js: `isWeb()` → false
- [ ] `FORCE_MEMFS=true` still forces memfs in `fs.ts`
- [ ] Existing browser bundle consumers still get memfs (window + document present)

### DIFF
--- a/src/fs/fs.ts
+++ b/src/fs/fs.ts
@@ -13,7 +13,10 @@ import type { VirtualFs } from './types';
 
 export let fs: VirtualFs;
 
-const isWeb = (): boolean => process.env.FORCE_MEMFS === 'true' || ('document' in globalThis && 'window' in globalThis);
+const isWeb = (): boolean => {
+  if (typeof process.versions?.bun !== 'undefined') return false;
+  return process.env.FORCE_MEMFS === 'true' || 'window' in globalThis || 'self' in globalThis;
+};
 
 export const getVirtualFs = (memfsVolume?: memfs.Volume): VirtualFs => {
   if (isWeb()) {
--- a/src/fs/fs.ts
+++ b/src/fs/fs.ts
@@ -9,14 +9,10 @@ import * as nodeFs from 'node:fs';
 // yes, we're going to import it even though it might not be used.
 // the alternatives were all worse without top-level await (iife, runtime errors from something trying to use it before it's initialized)
 import * as memfs from 'memfs';
+import { isWeb } from '../util/isWeb';
 import type { VirtualFs } from './types';
 
 export let fs: VirtualFs;
-
-export const isWeb = (): boolean => {
-  if (process.versions.bun) return false;
-  return process.env.FORCE_MEMFS === 'true' || 'window' in globalThis || 'self' in globalThis;
-};
 
 export const getVirtualFs = (memfsVolume?: memfs.Volume): VirtualFs => {
   if (isWeb()) {
--- a/src/fs/fs.ts
+++ b/src/fs/fs.ts
@@ -13,7 +13,7 @@ import type { VirtualFs } from './types';
 
 export let fs: VirtualFs;
 
-const isWeb = (): boolean => process.env.FORCE_MEMFS === 'true' || 'window' in globalThis || 'self' in globalThis;
+const isWeb = (): boolean => process.env.FORCE_MEMFS === 'true' || ('document' in globalThis && 'window' in globalThis);
 
 export const getVirtualFs = (memfsVolume?: memfs.Volume): VirtualFs => {
   if (isWeb()) {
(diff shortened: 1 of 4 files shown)


src/fs/fs.ts
```diff
@@ -13,7 +13,7 @@ import type { VirtualFs } from './types';
 
 export let fs: VirtualFs;
 
-const isWeb = (): boolean => process.env.FORCE_MEMFS === 'true' || 'window' in globalThis || 'self' in globalThis;
+const isWeb = (): boolean => process.env.FORCE_MEMFS === 'true' || ('document' in globalThis && 'window' in globalThis);
```
mshanemc: can you instead follow https://bun.com/docs/guides/util/detect-bun for the most conservative, bun-specific fix (ie, if the env exists, return `false` 

there's weird weird stuff in CodeBuilder and CodeBuilderWeb where this can get complicated (ex: you're in browser but vscode blocks access to the `window` object, etc).

we don't yet have the kind of "external-nut" for for CB/CBW/VScode ext to konw that this is safe without manually verifying all of them, and those builds can be a pain to set up.

mshanemc (dismissed):

mshanemc (approved):

PR #1309, opened 2026-07-01
feat: add projectPath and skipCache options to AuthRemover @W-23271398@

mshanemc:
### What does this PR do?

`AuthRemover.init` calls `ConfigAggregator.create()` with no arguments — which (a) resolves the **local** project config relative to `process.cwd()`, and (b) trusts the **process-cached** aggregator instance. That breaks two consumer scenarios:

- **cwd can't be trusted** (e.g. `salesforcedx-vscode`, whose extension host can run with `cwd` = `/`) → `removeAuth` can't find/unset local config keys like `target-org`.
- **long-lived cached aggregator** — a consumer that already created a `ConfigAggregator` for the project gets that same cached instance back (it's memoized by resolved projectPath). Its in-memory config may be stale (changed out of process, or set after the aggregator was built), so `unsetConfigValues` reads stale values via `getKeysByValue`, matches nothing, and silently no-ops.

This adds two optional, **backward-compatible** options to `AuthRemover`:

```ts
const remover = await AuthRemover.create({ projectPath: '/path/to/project', skipCache: true });
await remover.removeAuth('example@mycompany.com');
```

- **`projectPath`** — forwarded to `ConfigAggregator.create({ projectPath })` (matching the option name/shape `ConfigAggregator` already accepts) so local config resolves from the given project instead of `process.cwd()`.
- **`skipCache`** — when `true`, `reload()` the aggregator so config is re-read from disk before `unsetConfigValues` runs. Defaults to `false`.

**Existing behavior is unchanged**: omit both and `init` calls `ConfigAggregator.create()` with no options and does not reload — exactly as before. No existing caller is affected.

Tests: spy `ConfigAggregator.create` to assert `projectPath` is forwarded (and the no-arg case); spy `ConfigAggregator.prototype.reload` to assert it runs only when `skipCache` is `true` and not by default.

@W-23271398@ — enables `salesforcedx-vscode` to unset the local `target-org` on logout dir-explicitly (no `process.chdir`) and without being fooled by a stale cached aggregator.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/org/authRemover.ts
+++ b/src/org/authRemover.ts
@@ -39,11 +39,27 @@ const messages = Messages.loadMessages('@salesforce/core', 'auth');
  * );
  * await remover.removeAuth(auth.username);
  * ```
+ *
+ * Pass a `projectPath` to resolve local config from a specific project instead of `process.cwd()`,
+ * and `skipCache` to re-read config from disk when the process-cached aggregator can't be trusted:
+ *
+ * ```
+ * const remover = await AuthRemover.create({ projectPath: '/path/to/project', skipCache: true });
+ * await remover.removeAuth('example@mycompany.com');
+ * ```
  */
-export class AuthRemover extends AsyncOptionalCreatable {
+export class AuthRemover extends AsyncOptionalCreatable<AuthRemoverOptions> {
   private config!: ConfigAggregator;
   private stateAggregator!: StateAggregator;
   private logger!: Logger;
+  private readonly projectPath?: string;
+  private readonly skipCache?: boolean;
+
+  public constructor(options?: AuthRemoverOptions) {
+    super(options);
+    this.projectPath = options?.projectPath;
+    this.skipCache = options?.skipCache;
+  }
 
   /**
    * Removes the authentication and any configs or aliases associated with it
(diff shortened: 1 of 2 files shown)


src/org/authRemover.ts
```diff
@@ -190,3 +203,8 @@ export class AuthRemover extends AsyncOptionalCreatable {
     }
   }
 }
+
+export type AuthRemoverOptions = {
```
WillieRuemmele, line 216: should we re-export this at the top level? so you can `import {AuthRemoverOptions} from @salesforce/core`?

mshanemc: I considered it, but didn't have a use case for it.  (like, you'd only need that if you're passing around the type, like writing your own function that accepts/returns that type or something).

Your call

WillieRuemmele (dismissed):

WillieRuemmele (approved):

mshanemc:
QA was done with the consuming PR from vscode. https://github.com/forcedotcom/salesforcedx-vscode/pull/7624

the plugin-user NUT failure is related to 1p=>hyperforce not working well with the secondary user replication (known issue)

PR #1310, opened 2026-07-06
fix: remove double slash in determineIfDevHub request URL @W-23299190@

jonnyhork:
## Summary
@W-23299190@

- `determineIfDevHub` built its ScratchOrgInfo query URL as `` `${baseUrl.toString()}/services/data/...` ``. `SfdcUrl` extends the WHATWG `URL` class, and `toString()` on a bare origin always includes a trailing slash, so the extra leading slash produced a double slash (e.g. `https://instance.salesforce.com//services/data/...`).
- Removed the leading slash so the URL matches the pattern already used correctly by the sibling `retrieveUserInfo` method a few lines up in the same file.

## Test plan
- [x] Added a URL-shape assertion to the existing `determineIfDevHub` "should return true" unit test verifying the constructed request URL has a single slash between the instance URL and `services/data`
- [x] `yarn test:only`, `yarn lint`, `yarn compile`, `yarn bundle:check`, `yarn test:nuts` all pass

### DIFF
--- a/src/org/authInfo.ts
+++ b/src/org/authInfo.ts
@@ -1289,7 +1289,7 @@ export class AuthInfo extends AsyncOptionalCreatable<AuthInfo.Options> {
     const apiVersion = 'v51.0'; // hardcoding to v51.0 just for this call is okay.
     const instance = ensure(instanceUrl);
     const baseUrl = new SfdcUrl(instance);
-    const scratchOrgInfoUrl = `${baseUrl.toString()}/services/data/${apiVersion}/query?q=SELECT%20Id%20FROM%20ScratchOrgInfo%20limit%201`;
+    const scratchOrgInfoUrl = `${baseUrl.toString()}services/data/${apiVersion}/query?q=SELECT%20Id%20FROM%20ScratchOrgInfo%20limit%201`;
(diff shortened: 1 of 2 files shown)


src/org/authInfo.ts
```diff
@@ -1289,7 +1289,7 @@ export class AuthInfo extends AsyncOptionalCreatable<AuthInfo.Options> {
     const apiVersion = 'v51.0'; // hardcoding to v51.0 just for this call is okay.
     const instance = ensure(instanceUrl);
     const baseUrl = new SfdcUrl(instance);
-    const scratchOrgInfoUrl = `${baseUrl.toString()}/services/data/${apiVersion}/query?q=SELECT%20Id%20FROM%20ScratchOrgInfo%20limit%201`;
+    const scratchOrgInfoUrl = `${baseUrl.toString()}services/data/${apiVersion}/query?q=SELECT%20Id%20FROM%20ScratchOrgInfo%20limit%201`;
```
mshanemc, line 1292: because the instanceUrl is passed into this method a a param, this would break users who have it stored **without** the trailing slash.  These are often coming from the auth file.

if the problem is the double slash, it would be safer to take the instanceUrl, normalize it (either way, with or without a slash) and then build the url you want.

mshanemc (approved):

PR #1315, opened 2026-07-22
feat: env vars to override scratch org signup Connected App @W-23543471@

WillieRuemmele:
## Summary

- Adds `SF_SCRATCH_SIGNUP_CONNECTED_APP` and `SF_SCRATCH_SIGNUP_CALLBACK_URL` environment variables that override the Connected App used during scratch org signup
- Unblocks CI workflows using External Client Apps (ECA) + JWT for Dev Hub auth, which otherwise fail with C-1016 because Proxy Signup cannot replicate ECAs
- Env vars slot into the existing fallback chain without changing default behavior

## Context

- GitHub issue: https://github.com/forcedotcom/cli/issues/3515
- @W-23543471@

When a Dev Hub is authenticated via an ECA + JWT bearer flow, `sf org create scratch` fails with C-1016. The CLI derives the signup Connected App from the hub session's `clientId`, but ECAs cannot be replicated by Proxy Signup into the new scratch org.

**Usage:**
```bash
export SF_SCRATCH_SIGNUP_CONNECTED_APP=PlatformCLI
sf org create scratch ...
```

**Fallback priority:**
1. Explicit `--client-id` flag (existing behavior)
2. `SF_SCRATCH_SIGNUP_CONNECTED_APP` env var (new)
3. Hub org's clientId (existing behavior)
4. `PlatformCLI` default (existing behavior)

These env vars are intended as a temporary workaround until Proxy Signup supports ECA replication natively, at which point they become no-ops and can be deprecated.

## Test plan

- [x] Unit tests for env var override of consumer key
- [x] Unit tests for env var override of callback URL
- [x] Unit test verifying explicit flag still takes priority over env var
- [x] Unit test for both env vars together (full ECA workaround)
- [x] Existing tests unchanged (no regression)
- [ ] Integration test against ECA + JWT Dev Hub (community volunteer available)

### DIFF
--- a/messages/scratchOrgErrorCodes.md
+++ b/messages/scratchOrgErrorCodes.md
@@ -113,3 +113,10 @@ The template specified in the Scratch Definition isn’t supported. Specify a ge
 # C-9999
 
 A fatal signup error occurred. Please try again. If you still see this error, contact Salesforce Support for assistance
+
+# action.C-1016
+
+If your Dev Hub is authenticated via an External Client App (ECA), the signup service cannot replicate it into the new scratch org. As a workaround, set these environment variables to use Salesforce's built-in Connected App for signup instead:
+
+SF_SCRATCH_SIGNUP_CONNECTED_APP=PlatformCLI
+SF_SCRATCH_SIGNUP_CALLBACK_URL=http://localhost:1717/OauthRedirect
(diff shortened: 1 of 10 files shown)


WillieRuemmele:
confirmed fixed the[ issue](https://github.com/forcedotcom/cli/issues/3515#issuecomment-5051285103)

soridalac (approved):

PR #1319, opened 2026-07-29
feat!: require Node >=22.0.0, drop EOL Node versions @W-23480655@

WillieRuemmele:
## Summary
- Raise `engines.node` to `>=22.0.0`
- The dropped Node majors are past (or nearing) End-of-Life

## Work Item
@W-23480655@: Test: drop Node 18/20

BREAKING CHANGE: raises the minimum supported Node version to >=22.0.0.
Consumers on older Node must upgrade. This ships as a new major version.

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,12 +1,206 @@
-Copyright (c) 2026, Salesforce.com, Inc.
+Apache License Version 2.0
+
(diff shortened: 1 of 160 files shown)


iowillhoit (approved):
Test already passed, removed unused cache files

PR #1324, opened 2026-08-07
fix: resolve TS2694 pino.TransportSingleOptions type error in emitted .d.ts @W-23751214@

### LINKED ISSUE #3618: @salesforce/core type declaration references pino.pino.TransportSingleOptions, breaks builds with skipLibCheck: false (TS2694)
### Summary

`@salesforce/core`'s published type declarations reference `pino.TransportSingleOptions` in a way that fails `tsc` when library checking is enabled (`skipLibCheck: false`). Consumers that lib-check are forced to enable `skipLibCheck` to build.

### Error

```
node_modules/@salesforce/core/lib/logger/logger.d.ts(262,63): error TS2694: Namespace 'pino.pino' has no exported member 'TransportSingleOptions'.
```

### Root cause

`lib/logger/logger.d.ts` does:

```ts
import { type Logger as PinoLogger, pino } from 'pino';
// ...
export declare const getWriteStream: (level?: string) => pino.TransportSingleOptions;
```

pino's `pino.d.ts` uses `export = pino` with an **outer** `declare namespace pino` (where `TransportSingleOptions` is declared) and *also* a **nested** `namespace pino` inside it. The named import binding `{ pino }` resolves to the inner `pino.pino` namespace, which does not contain `TransportSingleOptions` — hence `TS2694: Namespace 'pino.pino' has no exported member 'TransportSingleOptions'`.

The fix is likely to reference the type off the default/`export =` binding rather than the named `pino` import, e.g. use the outer namespace type (`import type pino from 'pino'` / `pino.TransportSingleOptions`) or import `TransportSingleOptions` directly.

### Version bisect

| `@salesforce/core` | `logger.d.ts` line 1 import | References `pino.TransportSingleOptions`? | lib-check |
| --- | --- | --- | --- |
| 8.31.3 | `import { type Logger as PinoLogger } from 'pino';` | No | ✅ clean |
| 8.32.6 | `import { type Logger as PinoLogger, pino } from 'pino';` | Yes (line 262) | ❌ TS2694 |
| 9.1.0 | `import { type Logger as PinoLogger, pino } from 'pino';` | Yes (line 262) | ❌ TS2694 |

The regression was introduced in the 8.32.x line and persists in 9.1.0. pino resolved version at repro: `pino@9.14.0`.

### Reproduction

1. `npm i @salesforce/core@9.1.0`
2. In a `tsconfig.json` with `skipLibCheck` unset/false and `moduleResolution: NodeNext`, run `tsc --noEmit`.
3. Observe the TS2694 error above (the only error; it originates entirely inside `@salesforce/core`'s `.d.ts`).

### Impact

Downstream TypeScript consumers of `@salesforce/core` that keep `skipLibCheck: false` (the stricter, recommended setting) cannot compile without enabling `skipLibCheck`, which then disables `.d.ts` checking for *all* dependencies. A correct type declaration here would let consumers keep strict library checks.

### Environment

- `@salesforce/core`: 9.1.0 (and 8.32.6)
- `pino`: 9.14.0
- `typescript`: 6.0.3 (also reproduces on 7.0.2)
- `moduleResolution`: NodeNext

WillieRuemmele:
## Summary
- Fix TS2694 error where emitted `logger.d.ts` referenced `pino.TransportSingleOptions` through the inner `pino.pino` namespace, which doesn't export that type
- Import `TransportSingleOptions` directly from pino's module-level exports so the emitted `.d.ts` references the correct type path
- Also import `levels` directly (`pinoLevels`) to avoid namespace access through the same problematic inner binding

Fixes https://github.com/forcedotcom/cli/issues/3618

## Work Item
@W-23751214@: @salesforce/core type declaration references pino.pino.TransportSingleOptions, breaks builds with skipLibCheck: false (TS2694)

## Root Cause
`import { pino } from 'pino'` resolves to the inner namespace binding. When TypeScript emits `.d.ts`, it writes `pino.TransportSingleOptions` which consumers resolve as `pino.pino.TransportSingleOptions` — but `TransportSingleOptions` only exists in the outer `declare namespace pino`.

## Proof of Work
- `tsc --noEmit`: clean
- Full test suite: 7 scripts, all passing (31.8s)
- Emitted `lib/logger/logger.d.ts` now correctly shows:
  - `import { type TransportSingleOptions } from 'pino';`
  - `export declare const getWriteStream: (level?: string) => TransportSingleOptions;`

## Validation
- Per-task adversarial review: no blockers, no warnings
- Integration: all checks pass
- Type check: clean
- Build: clean

## Test plan
- [ ] Verify `tsc --noEmit` passes with `skipLibCheck: false` in a downstream consumer using `@salesforce/core`
- [ ] Verify existing logger tests pass
- [ ] Verify emitted `.d.ts` no longer references `pino.TransportSingleOptions`

### DIFF
--- a/src/logger/logger.ts
+++ b/src/logger/logger.ts
@@ -16,7 +16,7 @@
 import * as os from 'node:os';
 import * as path from 'node:path';
 
-import { type Logger as PinoLogger, pino } from 'pino';
+import { type Logger as PinoLogger, type TransportSingleOptions, pino, levels as pinoLevels } from 'pino';
 import { Env } from '@salesforce/kit';
 import { isKeyOf, isString } from '@salesforce/ts-types';
 import { Global, Mode } from '../global';
@@ -460,7 +460,7 @@ export class Logger {
 }
 
 /** return various streams that the logger could send data to, depending on the options and env  */
-export const getWriteStream = (level = 'warn'): pino.TransportSingleOptions => {
+export const getWriteStream = (level = 'warn'): TransportSingleOptions => {
   const env = new Env();
   // used when debug mode, writes to stdout (colorized)
   if (process.env.DEBUG) {
@@ -519,13 +519,13 @@ const levelFromOption = (value?: string | number): string => {
     case 'string':
       return value;
     default:
-      return pino.levels.labels[Logger.DEFAULT_LEVEL];
+      return pinoLevels.labels[Logger.DEFAULT_LEVEL];
   }
 };
 // /** match a number to a pino level, or if a match isn't found, the next highest level */
 const numberToLevel = (level: number): string =>
-  pino.levels.labels[level] ??
-  Object.entries(pino.levels.labels).find(([value]) => Number(value) > level)?.[1] ??
+  pinoLevels.labels[level] ??
+  Object.entries(pinoLevels.labels).find(([value]) => Number(value) > level)?.[1] ??
   'warn';
 
 const getDefaultLevel = (): LoggerLevel => {


soridalac (approved):

soridalac:
**QA notes**:

✅ Emitted `lib/logger/logger.d.ts` imports T`ransportSingleOptions` directly — no `pino.pino `namespace path
✅ `tsc --noEmit ` with s`kipLibCheck: false `— no TS2694 error
✅ Full test suite passes (37s, all green)
✅ `yarn build `compiles clean
✅ No remaining `pino.` namespace access in source

soridalac:
Failing tests/pdrNuts are not related to the PR.

